Issue #35, opened 2015-09-01
Put all public classes into a namespace to avoid pollution


whoshuu:
All of the classes being in the global namespace is probably a no-go for a lot of applications. The right thing to do is to introduce at least one namespace out of respect for those applications.

The most obvious namespace to use is just `cpr`. If everything is behind that namespace, that should be enough for the most part. Tests can use the normally bad `using namespace cpr` to avoid onerous changes in that directory.

whoshuu:
Labelling this issue as easy since anyone should be able to tackle this. If you're looking to contribute to this library, this would be a super easy, and impactful, place to start.

PR #44, opened 2015-09-06
Put everything into namespaces

smiley:
Diffs are a little wonky, but all I did was add:

```cpp
namespace cpr {

// ...
// Original content goes here
// ...

}
```

Around the files' content, and added `using namespace cpr;` to test files. (#35)


whoshuu:
Thanks for the PR!

PR #48, opened 2015-09-22
Modernize code (more idiomatic C++11 and faster container initialization)

SuperV1234:
Great talk and great library. 
Here's a small and simple pull request to modernize some code.

* Most containers can be constructed using `std::initializer_list`. 

* When dealing with initializer lists of key-value pairs, the key should be `const`. 

* Use for-each loops instead of iterators-based loops.

### DIFF
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -4,11 +4,8 @@
 
 namespace cpr {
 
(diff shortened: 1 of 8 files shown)


nabijaczleweli:
:+1:

whoshuu:
Looks good. Thanks @SuperV1234!

PR #49, opened 2015-09-22
Minor changes to modernize code (`using` instead of `typedef`, `struct` instead of `typedef struct`)

SuperV1234:


whoshuu:
Thanks for the PR @SuperV1234!

PR #53, opened 2015-09-29
Special case v1 cookies in GetEncoded()

skystrife:
Some application servers will put cookies in quotes if they contain certain characters. This happens with e.g. the base64 encoded session cookies for some webapps. In this case, url encoding the cookie actually breaks things on the recipient side of the request because the leading and trailing quotation marks were gobbled into %22s.

My fix was to simply special-case the scenario where we have a cookie value that begins and ends with a quotation mark, and to bypass the `urlEncode()` step for those. Maybe there's a better way of doing this, but this was the most obvious and fixed things for me.

### DIFF
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -10,14 +10,15 @@ Cookies::Cookies(const std::initializer_list<std::pair<const std::string, std::s
 std::string Cookies::GetEncoded() const {
     std::stringstream stream;
     for (const auto& item : map_) {
+        stream << cpr::util::urlEncode(item.first) << "=";
         // special case version 1 cookies, which can be distinguished by
         // beginning and trailing quotes
         if (!item.second.empty() && item.second.front() == '"' && item.second.back() == '"') {
-            stream << cpr::util::urlEncode(item.first) << "=" << item.second << "; ";
+            stream << item.second;
         } else {
-            stream << cpr::util::urlEncode(item.first) << "=" << cpr::util::urlEncode(item.second)
-                   << "; ";
+            stream << cpr::util::urlEncode(item.second);
         }
+        stream << "; ";
     }
     return stream.str();
 }
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -10,8 +10,14 @@ Cookies::Cookies(const std::initializer_list<std::pair<const std::string, std::s
 std::string Cookies::GetEncoded() const {
     std::stringstream stream;
     for (const auto& item : map_) {
-        stream << cpr::util::urlEncode(item.first) << "=" << cpr::util::urlEncode(item.second)
-               << "; ";
+        // special case version 1 cookies, which can be distinguished by
+        // beginning and trailing quotes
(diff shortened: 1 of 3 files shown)


whoshuu:
Interesting special case. Do those %22s get decoded back into quotations when you try to access the cookies on the client side? Or do they show up as %22s?

cpr/cookies.cpp
```diff
@@ -10,8 +10,14 @@ Cookies::Cookies(const std::initializer_list<std::pair<const std::string, std::s
...
     std::stringstream stream;
     for (const auto& item : map_) {
-        stream << cpr::util::urlEncode(item.first) << "=" << cpr::util::urlEncode(item.second)
-               << "; ";
+        // special case version 1 cookies, which can be distinguished by
+        // beginning and trailing quotes
+        if (!item.second.empty() && item.second.front() == '"' && item.second.back() == '"') {
+            stream << cpr::util::urlEncode(item.first) << "=" << item.second << "; ";
+        } else {
+            stream << cpr::util::urlEncode(item.first) << "=" << cpr::util::urlEncode(item.second)
```
nabijaczleweli: Looks like `stream << cpr::util::urlEncode(item.first) << "=" << ` and ` << "; "` is needlessly repeated, why not put those statements before/after the `if`s?

whoshuu: Good observation. Since we're using a `stringstream`, we might as well exploit it and avoid duplicate work when possible

skystrife:
If I use `curl -i` I get:

```
Set-Cookie: redacted_session="-snip-";Path=/
```

If I use CPR and print out the cookie (`response.cookies["redacted_session"]`) I get the following:

```
"-snip-"
```

where `-snip-` is a bunch of encoded stuff. The quotation marks, critically, *aren't* encoded in either case.

I can use `curl` with a cookie jar (`-c`) and the requests go through just fine (session is recognized) but with `cpr` it breaks because the re-encoding going on in `GetEncoded()` converts the expected `"` into `%22`.

This patch prevents this by checking whether the cookie's value begins and ends with quotation marks. Perhaps there's a better way, though. The issue, I think, is that the cookie that I'm being given as the result of making the first request is then being re-encoded by `cpr` when I use something like

```cpp
session.SetCookies(response.cookies);
```

which results in the server not seeing the exact cookie it asked me to set in the first request upon subsequent requests.

whoshuu:
I'd like to merge this in soon. Can you do one last thing and write up a test case to cement the behavior here? That way if the implementation you're providing changes, we'll know if it breaks in this special case. Per the coverage report, [this](https://coveralls.io/builds/3872702/source?filename=cpr%2Fcookies.cpp#L17) is the line I'm specifically interested in covering. I think a test case after [this one](https://github.com/whoshuu/cpr/blob/ed56e96ebeffec4ee264c707ddacc3ac3b3bb2fd/test/get_tests.cpp#L67-L80) would be appropriate.

Thanks for your work @skystrife. Just a few more changes and I'll merge this in.

whoshuu:
Thanks for the PR @skystrife!

PR #57, opened 2015-10-25
allow construction of cpr::Payload objects with std::vector

andihit:
if there is a better way to share the same code for these two constructors, please tell me and I'll update this pull request :)

### DIFF
--- a/cpr/payload.cpp
+++ b/cpr/payload.cpp
@@ -7,14 +7,14 @@
 
 namespace cpr {
 
-Payload::Payload(const std::initializer_list<Pair>& pairs) {
-    for (const auto& pair : pairs) {
-        if (!content.empty()) {
-            content += "&";
-        }
-        auto escaped = cpr::util::urlEncode(pair.value);
-        content += pair.key + "=" + escaped;
+Payload::Payload(const std::initializer_list<Pair>& pairs) : Payload(begin(pairs), end(pairs)) {}
+
+void Payload::AddPair(const Pair& pair) {
+    if (!content.empty()) {
+        content += "&";
     }
+    auto escaped = cpr::util::urlEncode(pair.value);
+    content += pair.key + "=" + escaped;
 }
 
 } // namespace cpr
--- a/include/payload.h
+++ b/include/payload.h
@@ -4,8 +4,10 @@
 #include <memory>
 #include <string>
 #include <initializer_list>
+#include <vector>
 
 #include "defines.h"
+#include "util.h"
 
 namespace cpr {
 
@@ -24,7 +26,15 @@ struct Pair {
 
 class Payload {
   public:
+    template <class It>
+    Payload(const It begin, const It end) {
+        for (It pair = begin; pair != end; ++pair) {
+            AddPair(*pair);
+        }
+    }
     Payload(const std::initializer_list<Pair>& pairs);
+
+    void AddPair(const Pair& pair);
 
     std::string content;
 };
--- a/test/post_tests.cpp
+++ b/test/post_tests.cpp
@@ -27,6 +27,39 @@ TEST(UrlEncodedPostTests, UrlPostSingleTest) {
     EXPECT_EQ(201, response.status_code);
 }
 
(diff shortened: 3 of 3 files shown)


whoshuu:
Hey @andihit, this looks good. It looks like there isn't a test for the interface you're writing, can you write a test that makes sure `Payload` constructed from `std::vector<Pair>` behaves as you would expect? A simple POST test in `post_tests.cpp` should suffice. Thanks!

andihit:
sure thing! nice lib btw.

one question though: (unrelated to this PR) how can I access curl errors? there is an `error` field in `CurlHolder`, but I couldn't find a public accessor for it
background: I got a response with status_code = 0 and couldn't figure out until I modified the `cpr` code to print out errors
(I was accessing a https resource but the local curl wasn't compiled with SSL support - current workaround is to use system curl (`USE_SYSTEM_CURL` option))

nabijaczleweli:
AFAICT The Standard Template Library [takes](http://en.cppreference.com/w/cpp/container/vector/vector) [`std::initializer_list`](http://en.cppreference.com/w/cpp/container/list/list) [by](http://en.cppreference.com/w/cpp/container/set/set) [value](http://en.cppreference.com/w/cpp/container/multimap/multimap), why not do it here?

andihit:
thx, this is even better

whoshuu:
@andihit, keep an eye on #60 and #61. There's work being done to surface errors from curl up to a client. Feel free to comment on either of those if you have strong opinions on how this should be implemented.

whoshuu:
Additionally, as a result of #58, there's an additional API for adding individual parameters to `cpr::Parameters`. Do you think your use case can be accomplished with this:

```c++
std::vector<cpr::Parameter> vector_of_parameters;
cpr::Parameters parameters;
for (auto& parameter : vector_of_parameters) {
    parameters.AddParameter(parameter);
}
```

Let me know if this works for you! I'd much prefer if we could keep container-specific assumptions out of the API's as much as possible. It would let me avoid interface explosion that isn't necessary: that is, a separate API for `std::list`, `std::vector`, `std::array`, etc.

Ideally, this could just be a templated function that takes in any container type, but it's got be written carefully enough to play nicely with the `std::initializer_list` constructor, since that's the primary way of constructing this object.

nabijaczleweli:
You *could* take an iterator pair, which is container-agnostic

andihit:
@whoshuu: yes, but I need it for `cpr::Payload` - should I change my PR to use the same API as in #58?
@nabijaczleweli: did that in my first commit  d791b36

whoshuu:
I think the point I'm trying to make is that the iterator pair constructor should be public, like the `std::initializer_list` one. There's no need to cement a constructor in that takes a vector because an iterator pair one is more generic, and works for user-defined containers to boot. So assuming there's a constructor like this:

```c++
class Payload {
  public:
    template <class It>
    Payload(const It begin, const It end) {
        for (auto& it = being; it != end; ++it) {
            AddPair(*it);
        }
    }
};
```

would that be suitable for your use case, ie:

```c++
std::vector<cpr::Pair> pairs;
cpr::Payload payload(pairs.begin(), pairs.end());
```

Let me know if this works for you. If it does we can start moving this PR in that direction.

andihit:
yes, that'd work
I implemented the changes and rebased the commits into a single one.

nabijaczleweli:
Looks like the build fails only on GCC5 with [double free or corruption](https://travis-ci.org/whoshuu/cpr/jobs/89975021#L1267). Interesting.

whoshuu:
@nabijaczleweli, I've actually seen that happen before. I'm re-running the build and seeing if that resolves it. In the meantime it might be a good idea to mark it down as an issue so it can be addressed at some point.

whoshuu:
@andihit, this looks good to me. I'm going to merge it in now. Thanks for the PR! Appreciate that you respected the suggestions in the comments.

nabijaczleweli:
@whoshuu Interesting, it shouldn't do that. Maybe some well-hidden UB?

PR #58, opened 2015-11-04
"Parameters" class lacks a method to add a parameter

mexus:
While writing an application which uses the `cpr` library I found impossible to add parameters to a request in a dynamic fashion.
For example, I have some parameters backed up in a std container, and I want to add all of them to the `cpr::Parameters` object, but I don't know the size of the container in advance.
Maybe I'm misleading something, maybe this funcionality already exists in the project, nevertheless I wasn't able to find it. Please correct me if I am wrong, but I think this simple patch can be useful.

### DIFF
--- a/cpr/parameters.cpp
+++ b/cpr/parameters.cpp
@@ -9,12 +9,16 @@ namespace cpr {
 
 Parameters::Parameters(const std::initializer_list<Parameter>& parameters) {
     for (const auto& parameter : parameters) {
-        if (!content.empty()) {
-            content += "&";
-        }
-        auto escaped = cpr::util::urlEncode(parameter.value);
-        content += parameter.key + "=" + escaped;
+        AddParameter(parameter);
     }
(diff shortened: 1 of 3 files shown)


whoshuu:
Hey @mexus, this is a pretty good find. You're right, there's no way to change `Parameters` after its been instantiated, except through direct modification of the `content` member. If possible, can you add a test after [this one](https://github.com/whoshuu/cpr/blob/0ac8394ee71db72cb11873ae584feb06aa003efb/test/get_tests.cpp#L129-L139) that demonstrates the expected behavior?

mexus:
@whoshuu , sure thing!

whoshuu:
Thanks @mexus, merged!

mexus:
@whoshuu, wow, that's fast =) BTW, thank you for your nice lib! Finally we have some sane and true-c++ curl wrapper!

Issue #31, opened 2015-08-10
Openssl support


whoshuu:

webmaster128:
Is this OpenSSL specific or are you talking about [generic TLS support](http://curl.haxx.se/docs/ssl-compared.html)? I guess as user of curl, you should be able to abstract away from the TLS implementation used.

whoshuu:
Yeah I should clarify that I want generic TLS support. The library client should probably not care what's used behind the scenes to do TLS, this is purely a build step detail.

That said, if I were to embed a particular TLS library, OpenSSL appears to be the way to go.

tiwoc:
On Windows and OS X, you can configure cURL to simply use the OS-provided TLS stacks (Secure Channel, Secure Transport), which has multiple upsides to shipping OpenSSL or one of its forks:
* smaller binary size, because the library is already there
* no need to maintain/update the list of trusted root certificates, which is not included in OpenSSL
* no need to issue emergency patches of your app when the next security issue in the TLS library is found, because the OS vendor will do it

On Linux, you can feed the OS-provided trusted root certificates to cURL if you can find them. Here's where we look for them in [Kullo](https://www.kullo.net/):
https://github.com/kullo/client-httpclient-curl/blob/master/httpclient/cabundle.cpp

The CA bundle can then be fed to cURL using `CURLOPT_CAINFO`:
https://github.com/kullo/client-httpclient-curl/blob/master/httpclient/httpclientimpl.cpp#L137

tyoungjr:
FYI for people just looking to get a quick and dirty requests over SSL. Linking against lib CURL (compiled with OpenSSL) I was able to make requests over https. 
Just add it to your cmake lists and executable:
find_package(CURL)
include_directories(${CURL_INCLUDE_DIRS})
target_link_libraries(exe ${CURL_LIBRARY})

rwebber:
is it currently possible to make connections via Oauth?

dontsovcmc:
There are some options for SSL:

Not check SSL:
```
curl_easy_setopt(curl_, CURLOPT_SSL_VERIFYPEER, false); 
curl_easy_setopt(curl_, CURLOPT_SSL_VERIFYHOST, false); 
```
Self-signed serfiticate:
```
curl_easy_setopt(curl_,CURLOPT_SSLCERTTYPE,"DER");
curl_easy_setopt(curl_,CURLOPT_CAINFO, path_to_cert_.c_str());
[...]

PR #61, opened 2015-11-07
Adding error objects to request responses

DEGoodmanWilson:
Sometimes CURL just gives up, and it would be nice to have some understanding of why. It would also be nice if we could do it in a non-CURL-centric way. This PR introduces a new Error object to encapsulate specifically _internal_ and _HTTP_ errors that can occur when performing a request. It also introduces a new enum ErrorCode to provide a machine-readable way of assessing errors that is also independent of the CURL codes.

Not all CURL errors map to ErrorCodes. This is because many of the errors are for protocols beyond the current scope of cpr (_e.g._ FTP). Many are due to errors on the server which are already suitably handled by the HTTP status_code feature. And a lot of them are CURL specific, of course, and so get rolled up into the generic INTERNAL_ERROR. (Also, many of the SSL errors are very specific, and I wasn't sure if the way CURL granulated them was both universally applicable and user friendly, so I just lumped them together under a small handful of SSL errors). The CURL-generated human readable text is placed in the ERROR object in case the end developer requires more context.

Test coverage is not 100%, because in many cases I am not certain how to induce certain of the CURL error conditions.

I've tried to stick to the house style, please let me know if there are any formatting mistakes as well! I look forward to reading your comments.


nabijaczleweli:
Looks like [here](https://travis-ci.org/whoshuu/cpr/jobs/89761082#L591) you're referring to `CURLE_HTTP2` which doesn't exist, causing the build to fail

DEGoodmanWilson:
Ah, I was using the system CURL, for which that value does exist. I suspect it can be safely elided, which I will do.

DEGoodmanWilson:
Hrm. Currently the tests are failing because the test for `SSL_CONNECTION_FAILURE` is triggering `UNSUPPORTED_PROTOCOL` instead. Is there an obvious compile-time check I can do to see if SSL is enabled, preferably without querying any defined CURL macros?

whoshuu:
When you say compile-time, do you mean within CMake? I think you can check `SSL_ENABLED` from [here](https://github.com/bagder/curl/blob/729b92afbe4a350ee725a5001f041170dd12789b/CMakeLists.txt#L574). Pending #31, it might even make sense to set `CMAKE_USE_OPENSSL` (found [here](https://github.com/bagder/curl/blob/729b92afbe4a350ee725a5001f041170dd12789b/CMakeLists.txt#L298)) to `OFF`.

DEGoodmanWilson:
Oh, I just meant conditional compilation by examining some macro definition. When I build using system curl on OSX, I get SSL for free. (And building using the included CURL fails, but that's a different issue). But when I build on Windows with the included CURL, there is no SSL (as you well know). So it doesn't make sense to test SSL functionality where it isn't available (but it seems like it should when it is).

Not my favorite thing to do ever (because it requires us to talk to an implementation detail that oughtn't be exposed), but I'm going to just do runtime check against the feature flags in `curl_version_info()`.

whoshuu:
Just did a thorough read over your PR here. This is a great start! I'm going to build out some support inside the embedded test server to surface the other errors you're having trouble reproducing. Once I get a good hold of that, and get as close to 100% coverage again as possible, I'll merge this in and throw my changes on top of it.

DEGoodmanWilson:
Great! Let me know if there is anything I can do from my end to ease that process.

DEGoodmanWilson:
Oh hello, just wanted to check in and see if there was anything I could do to pitch in on this?

whoshuu:
Hey @DEGoodmanWilson, just been busy with work! I'm gonna cut 1.2.0 and just get it out there, then get this into the master branch and we'll take it from there. Expect this within the next few days. Thanks for your patience!

PR #62, opened 2015-11-16
Move include/* to include/cpr/*, fix parameters

sean-:
The move is pretty mechanical and straight forward to understand.  Feel free to reject, but it's useful for being able to 'cp *' into an existing project.

The second item contains two fixes.  First: parameters with no value shouldn't include an = in the generated URL.  Second: both the key and the value need to be escaped.

Thanks again!


whoshuu:
Hey @sean-!! This is awesome! Can I suggest one thing before I take a more thorough look at things here? If possible, can you split this into two pull requests? You can do that however you like (close this and open two more, or rebase this and open another), it just makes it easier to review each set of changes independently, and prevent one from blocking the other.

The goal is to get as much of your contributions in, as soon as possible, and this'll certainly help us get there!

sean-:
Yeah, absolutely.  I intended to do that, but pushed one of the {{#include "cpr/"}} moves accidentally and said eff it, pushed it all.  Not a biggie, give me a few minutes.  I'll push the parameters fix first, then the small code reorg.

PR #79, opened 2016-02-29
Use fixed-size numerical types

nabijaczleweli:
On x86-64 `int` and `long` are 32 bits wide, so `std::int32_t` was used

Closes https://github.com/whoshuu/cpr/issues/78

### DIFF
--- a/cpr/error.cpp
+++ b/cpr/error.cpp
@@ -4,7 +4,7 @@
 
 namespace cpr {
 
-ErrorCode Error::getErrorCodeForCurlError(int curl_code) {
+ErrorCode Error::getErrorCodeForCurlError(std::int32_t curl_code) {
     switch (curl_code) {
         case CURLE_OK:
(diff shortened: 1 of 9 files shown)


whoshuu:
That was quick :smile:. Thanks for the PR @nabijaczleweli, I'll take a look tonight to make things check out and get this in if we're all good.

I think officially libcurl wants `long` type options to be at least 32 bits, which is more aptly satisfied with `int_least32_t`, but `int32_t` is technically at least 32 bits already, so we can use that instead. I prefer as little implementation defined behavior as possible, and fixing the width to 32 is perfectly agreeable.

nabijaczleweli:
Any update on this?

whoshuu:
Merged in https://github.com/whoshuu/cpr/commit/272f6353910457325f1dc9ce0615592cb271fef0. Thanks for the PR!

PR #84, opened 2016-03-07
Preliminary support for conan.io

DEGoodmanWilson:
I've been looking more and more at conan.io as a C++ dependency manager. I'm starting to like it. Indeed, I'm starting to need something like this for my projects, as I have totally outgrown git submodules as a system for managing dependencies (I now am having to untangle collisions when the same project—namely CPR :D—is required by two different dependencies!)

So, in the spirit of getting the ball rolling, here is a basic conanfile.txt that I whipped up. I don't have access to many systems for testing, so I would be interested to know if it works on, _e.g._ Linux and Windows—but so far it is working great on OS X. I need to put together a sample project that depends on CPR for demonstration and testing purposes, watch for that. Anyway, consider this a first draft, and the starting place for a conversation about whether this is wanted, and if so what can be improved.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -3,6 +3,9 @@ from conans import ConanFile, CMake
 class CPRConan(ConanFile):
     name = "cpr"
     version = "1.2.0"
+    url = "https://github.com/whoshuu/cpr.git"
+    license = "MIT"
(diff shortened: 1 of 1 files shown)


DEGoodmanWilson:
Actually, this isn't ready yet. :(

PR #85, opened 2016-03-08
Preliminary Conan support

DEGoodmanWilson:
OK, so I'm back, and this time it's more or less working. But it's a bit ugly, and I'd like to discuss ways to make this less ugly, if indeed Conan is the kind of thing you'd like to support.

More about Conan: https://www.conan.io

You can check out how this work by having a look at this project, and following the instructions.
https://github.com/DEGoodmanWilson/cpr-conan-demo

Note that there is one line that would have to change in this PR—to point to this repo, instead of mine—before it could be merged.

Looking forward to hearing your thoughts on this proposal.


whoshuu:
Hey @DEGoodmanWilson, sorry for the late response. I appreciate you taking the time to do the integration yourself and not just open a ticket saying, "Integrate with conan.io" (though that would have been appreciated as well).

A few questions before we move forward with this:

1. If CMake's maturity is at a 10 (in terms of adoption, documentation, interface stability, etc.), where would you put conan.io's maturity?
2. Would having CPR in this package manager increase adoption?
3. What maintenance burden do you foresee with keeping this integration up to date? Is it a simple matter of changing versions and checksums in the conanfile.py?

Questions aside, this is definitely headed in the right direction. C++ dependency management has always been a huge pain point when it comes to library integration and adoption, and if we could ease it for this library and for the community, considering the costs, then it's probably worth it.

test/error_tests.cpp
```diff
@@ -64,7 +64,7 @@ TEST(ErrorTests, BoolFalseTest) {
 TEST(ErrorTests, BoolTrueTest) {
     Error error;
     error.code = ErrorCode::UNSUPPORTED_PROTOCOL;
-    EXPECT_TRUE(error);
```
DEGoodmanWilson: I find the change at this line bothersome, BTW. When I compile _without_ Conan support, the original compiles just fine. _With_ Conan support, the original doesn't. And indeed, I find myself surprised that the original ever compiles at all! Life is full of mysteries.

DEGoodmanWilson:
TBH, I've been on the fence about this PR ever since I issued it. I am not sure that, Conan's maturity aside, that this PR is fully baked yet. I opened it to start a discussion in code, as it were. And in English ;)

1.  I'm not sure what the lower bound is, so I can't quite place it, but I will say that Conan is still in its early days, and is not yet ready for prime time. Part of this is that the Conan codebase itself remains immature, and part of it is that the ecosystem is still quite small and in large part curated by the Conan team themselves. Mostly just @lasote. I have already run into issues with CPR + Conan-provided libcurl+openssl that I am working with the author to figure out, but that reveal some potentially deeper issues with Conan. So I would rank it a…5, let us say.

2.  In the short-term, no. Conan was just started back in December, and I am an early adopter for sure. I am already seeing some benefit to using it in my projects, but adoption on my part remains experimental. I suspect the same is true of many other adopters. It rose out of the Biicode ashes, so I have high hopes for it getting some traction over the coming year, and so I do think that there will be an uptick in CPR adoption, but it will be hard to see for the next few months, and relatively minor about a year out. But as you yourself said, there is a real itch that Conan is scratching, and at the moment, I think Conan is the best tool to scratch this itch that has yet arrived on the scene. So it has a degree of inevitibaility to it, unless something measurably better comes along soon enough to supplant it---again, I would say within the next year.

3.  There are a handful of small maintainence burdens adopting this invokes, which I volunteer to take on myself: Version updating, keeping relevant options up to date, further dependency management, and keeping the Conan.io servers up-to-date with each release. For the most part, these are burdens are slight, since Conan aims to _describe_ a package, rather than provide any mechanism for building a package---it defers entirely (in this case) to CMake for all building.

Finally, I am not super happy with the way that Conan ended up integrated into the `CMakeLists.txt`s in this project, and I think I would like to rework that portion of this PR one last time before---if---it is approved. Your opinions here are greatly welcome.

Moreover, given the troubles with Conan-provided libcurl+OpenSSL, indeed, I think merging this is a bad idea _at the moment_. I am working with the Conan project to get these issues resolved, however, and I'm curious how that will shake out. (You can follow along, if you like, or have an opinion: https://github.com/lasote/conan-libcurl/issues/2)

DEGoodmanWilson:
OK, at this point, I think I have everything streamlined, and as unobtrusive on the main project as possible. There should be no interference with the way things currently work—only the addition of compatibility with Conan. I think this is really finally ready for review.

DEGoodmanWilson:
And now the tests are passing! I'd like to bump this PR for consideration again.

jgsogo:
+1 for going ahead with this PR. 

Maybe conan's maturity is not 10/10, but it does not conflict with the current CMake workflow. We are all early adopters, mainstream adoption will depend on having libraries available on conan server, let's overcome the egg/chicken problem :seedling:

DEGoodmanWilson:
Fixed in github.com/DEGoodmanWilson/conan-cpr

PR #90, opened 2016-03-20
Add cpr::Timeout constructor with std::chrono, and store timeout dura…

FelixVanorder:
…tion as std::chrono::milliseconds.

Since C++11 has standard tools to manage timepoints and durations, it's looks much more natural to setup timeout in `std::chrono::milliseconds`, or seconds, or minutes (they cast down transparently) or as difference between two timepoints or durations. And makes it totally clear which time measurement is used.

Modification is trivial and very straightforward. Only one remark: `curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, timeout)` expect last argument to be `long`, but `std::chrono::milliseconds::count()` return `int64_t`, so there could be some problems, if somebody will decide set timeout longer than 24 days and 20 hours (=2'147'483 seconds).

### DIFF
--- a/cpr/CMakeLists.txt
+++ b/cpr/CMakeLists.txt
@@ -16,6 +16,7 @@ add_library(${CPR_LIBRARIES}
     payload.cpp
     proxies.cpp
     session.cpp
+    timeout.cpp
     util.cpp
 
     # Header files (useful in IDEs)
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <functional>
 #include <string>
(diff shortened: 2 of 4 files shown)


whoshuu:
Re-running your build on travis to see if that resolves the bad host issues. If it doesn't, I'll take a closer look at things. Otherwise, really awesome stuff @FelixVanorder! This will definitely move the library forward!

FelixVanorder:
Can't figure out what's wrong with windows build now, and how to re-run it. Same as for failed linux builds. :(

whoshuu:
Thanks for the PR @FelixVanorder!

Issue #91, opened 2016-04-20
Every method call after PUT is a PUT


tungox6:
I store a singleton HttpSession to make requests to a web server. After making a PUT request, I make a GET request (httpSession.Get()) but HttpSession does a PUT request to server.

PR #92, opened 2016-04-23
Fix issue #91, by setting CUSTOMREQUEST to NULL for each GET

bobjansen:
The issue seems to be that for sessions the setting of CURLOPT_CUSTOMREQUEST is retained by the CURL handle. This causes every next GET request to be a PUT request, see http://manpages.ubuntu.com/manpages/precise/man3/curl_easy_setopt.3.html

This commit fixes that by always setting CURLOPT_CUSTOMREQUEST to NULL

This doesn't include tests, what would be a good way to perform a test on the HTTP method used in 
the current framework?


bobjansen:
I suspect that HEAD and POST suffer from the same problem.

bobjansen:
I've written some tests and for some reason HEAD and POST do not seem to suffer from the same problem.

whoshuu:
This looks great! Thanks @bobjansen!

Issue #78, opened 2016-02-29
Use fix-sized numerical types


whoshuu:
Safety first!

Currently we're using implementation-defined numerical types in these locations:

- [x] [timeout.h](https://github.com/whoshuu/cpr/blob/3de75778fc225593281cba75b1a77a614d82618b/include/cpr/timeout.h#L10)
- [x] [response.h](https://github.com/whoshuu/cpr/blob/3de75778fc225593281cba75b1a77a614d82618b/include/cpr/response.h#L25)
- [x] [session.h](https://github.com/whoshuu/cpr/blob/4167141783769246c2a463ece14bd3ef36322c7e/include/cpr/session.h#L39)

PR #94, opened 2016-04-24
Make response_code long as suggested by CURL documentation

### LINKED ISSUE #93: BadHosts test fail on g++
Currently the BadHosts fail, no error is detected

bobjansen:
Fixes #93, response_code should be a long according to
https://curl.haxx.se/libcurl/c/curl_easy_getinfo.html. Before this
commit it was a std::int32_t. In g++ this caused the call to
curl_easy_getinfo to overwrite curl_error and set it to 0 which made the
BadHost tests fail.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -352,7 +352,7 @@ Response Session::Impl::makeRequest(CURL* curl) {
     auto curl_error = curl_easy_perform(curl);
 
     char* raw_url;
-    std::int32_t response_code;
+    std::int64_t response_code;
     double elapsed;
     curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response_code);
     curl_easy_getinfo(curl, CURLINFO_TOTAL_TIME, &elapsed);
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -352,7 +352,7 @@ Response Session::Impl::makeRequest(CURL* curl) {
     auto curl_error = curl_easy_perform(curl);
 
     char* raw_url;
-    std::int64_t response_code;
+    long response_code;
     double elapsed;
     curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &response_code);
     curl_easy_getinfo(curl, CURLINFO_TOTAL_TIME, &elapsed);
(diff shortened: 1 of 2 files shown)


bobjansen:
Travis looks a lot better now. I hope that [together with this](https://github.com/whoshuu/cpr/commit/8082d89d51bbd362d0eed8f5b94d2e2b94f8a665#commitcomment-17223576) all can be fixed.

bobjansen:
But now AppVeyor fails horribily. Should response_code just be a long, that would correspond exactly with the CURL documentation?

nabijaczleweli:
No impl-def width numeric types

bobjansen:
Why? In the current set-up curl is compiled when building this lib. The [curl source uses a long](https://github.com/whoshuu/curl/blob/master/lib/getinfo.c#L122), the best way to ensure no mismatch in size to conform with the type used there. I don't see a better way to prevent this from happening.

nabijaczleweli:
#78

bobjansen:
That's all well and good but `curl_easy_getinfo` is expecting a pointer to a long, this pointer will have an implementation defined size and passing in an integer type of another size yields undefined behavior. We don't want that, right?

nabijaczleweli:
Ask @whoshuu :G

whoshuu:
I'm inclined to side with @bobjansen here. It's unfortunate that an internal interface (to curl) corresponds to an external interface on cpr, but since this is just supposed to be a simple wrapper, it makes sense to match up those integer types.

For the travis failure, I think my recent commit [here](https://github.com/whoshuu/cpr/commit/08ebe6473474d189b13559c29dd744d889c46f30) might help with that. I'll take a look into the appveyor stuff shortly.

whoshuu:
@bobjansen, I think just making `status_code` a long will fix up the appveyor builds. When you do that and all the builds are good, excepting the clang-3.8 one, I can merge this in.

bobjansen:
Fixed except for 3.8 @whoshuu \o/

bobjansen:
With the fix for travis all are passing: https://github.com/bobjansen/cpr

whoshuu:
Thanks @bobjansen!

PR #97, opened 2016-05-12
Only set property on curl target if it exists

klaussilveira:


whoshuu:
This looks good, I think this should go in after exposing `BUILD_CURL_EXE` to the project CMakeLists.txt with a default of `ON` to mirror the option in the actual curl project: https://github.com/whoshuu/curl/blob/06bf874bbca0a5c600b210b5db920eff9f95f0d0/CMakeLists.txt#L74.

I can go ahead and do this and afterward merge your PR in. Thanks @klaussilveira!

whoshuu:
Merged in https://github.com/whoshuu/cpr/commit/dae957ec3f3b480561f1eceada46d84e60739bca, thanks @klaussilveira!

PR #99, opened 2016-05-17
Allow sending a body in a DELETE requests

fuchsch1234:
Currently payload set in the body is silently dropped, when a DELETE requests is sent.
For example
`cpr::Delete(cpr::Url{URL},cpr::Body{"'foo': 'bar'"}, cpr::Header{{"Content-Type", "application/json"}});`
will send a DELETE requests without the body.
With this change the body will be transferred.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -67,6 +67,7 @@ Session::Impl::Impl() {
         curl_easy_setopt(curl, CURLOPT_MAXREDIRS, 50L);
         curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, curl_->error);
         curl_easy_setopt(curl, CURLOPT_COOKIEFILE, "");
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, 0L);
 #ifdef INSECURE_CURL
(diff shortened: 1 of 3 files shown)


bobjansen:
LGTM but could you add a test?

fuchsch1234:
Added a test for delete requests with a simple json body.
Additionally I had to set make sure POSTFIELDSIZE is initialized to 0, else delete without a body does not work anymore.

whoshuu:
It's very strange that the build is failing for clang-3.8. You're not making any changes there, so I suspect it's an issue with travis. Rerunning the build doesn't help, so I think there's some caching being done specifically for your pull request that's making it impossible to break through the hurdle. I'm going to push a branch up to travis to build this again and see if that helps.

nabijaczleweli:
Travis police! :ZOOP: :ZOOP:

Despite lack of access to an actual PC (and therefore lacking the ability to review the build logs (Travis' wubsite doesn't work on mobile)) there's possibility you don't add the `clang-3.8` sources because unsuffixed ones are now for Clang 3.9.

whoshuu:
Merged in https://github.com/whoshuu/cpr/commit/ee29db0d9c5f4b9e9af05caf2fa873514e906008. Thanks for the PR @fuchsch1234!

Issue #74, opened 2016-02-09
Multi-threaded DNS lookup


wngr:
Hi,

I'm developing an application, which is heavily multi-threaded. From time to time the application segfaults. After some investigation I came about this thread: http://stackoverflow.com/questions/21887264/why-libcurl-needs-curlopt-nosignal-option-and-what-are-side-effects-when-it-is

After adding the CURLOPT_NOSIGNAL option and using c-ares for async dns lookup (needs recompilation), the application now runs stable.

Maybe this is just worth an entry in the documentation. Or maybe it can be added to the cpr API, but this may need some further investigations.

What are you thoughts on that?

Oliver

whoshuu:
Hey @wngr, thanks for bringing this issue to my attention. You said that the application became stable only after recompiling with c-ares support *and* setting `CURLOPT_NOSIGNAL` to `1`. The stackoverflow answer you linked seems to say you only need to do one or the other, but not both. Can you clarify why you had to do both to achieve stability?

I'm open to adding a note in the documentation about this. If you are feeling charitable, feel free to fork this project and put something small together for the [Advanced Usage](https://github.com/whoshuu/cpr/blob/4236d2a6be6536b557cc3735782d123ca50bb0cd/advanced-usage.md) section, preferably just adding to the very bottom of the page, and we can go from there!

As for adding that to the actual API, I'll have to ponder that one for a bit. My initial inclination is to have it be a compile time option, since the other solution is also compile time (integrating with c-ares). That way, a user won't have to constantly set `CURLOPT_NOSIGNAL` every time they make a network request in a multithreaded context.

This removes some flexibility from the user, but I think the cost is low compared to the benefit of being able to set something and forget it, without having to worry about instability at some undefined point in their application lifetime.

My thoughts might change as I get more time to think on this, but those are my initial thoughts for now. Feel free to chime in with yours!

puppalasatish:
Hi Whoshuu,

I am facing the same issue. As I am new to cmake and make, can you please help me with the steps to build cpr with c-ares. Also, if time permits, please tell me on how to set CURLOPT_NOSIGNAL. I want to try both the options. Thanks in advance..

Note:  I am not placing cpr as subproject. Instead, i am using .a file as library which is generated by building example project.

puppalasatish:
I am able to compile the curl with c-ares and issue got fixed, thanks :)

PR #105, opened 2016-06-17
Build option to disable use of signals in curl

fuchsch1234:
Curl sometimes segfaults on linux if used in a multithreaded program due to the use of signals. This is already discussed in issue #74.

This commit adds a build option to disable the use of signals by Curl.


whoshuu:
Looks good to me. Thanks for the PR @fuchsch1234! Merging in now.

PR #110, opened 2016-07-16
Add LowSpeed option

vladimirgamalian:
Sometimes it could be convinient to use CURL options CURLOPT_LOW_SPEED_LIMIT/CURLOPT_LOW_SPEED_TIME instead CURLOPT_TIMEOUT_MS. For example, when you don't know how many bytes can be transferred.
This commit adds a LowSpeed option to define these two CURL options.
Also, an appropriate test has been added to error_tests.cpp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ C++ Requests currently supports:
 * Basic authentication
 * Digest authentication
 * Timeout specification
+* Timeout for low speed connection
 * Asynchronous requests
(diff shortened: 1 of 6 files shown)


whoshuu:
Merged in https://github.com/whoshuu/cpr/commit/697b750b0c4c904673aef1a6ef7b63b8b012cf7e. Thanks for the PR @vladimirgamalian!

PR #115, opened 2016-07-28
Enable ccache if found for faster builds

thedrow:
When hacking on this library/running CI there's no need to recompile the entire project every time.
This should make compilation faster.


thedrow:
Yup from ~75s to ~5s. Please merge.

adam-nielsen:
Don't you normally configure `ccache` to be enabled system-wide, so that all projects use it by default?  Then you don't need to add support for it to every project you work with.  For example [here is how to do it on Arch Linux](https://wiki.archlinux.org/index.php/ccache#Enable_for_command_line), other distros are similar I'm sure.

thedrow:
CMake insists on not using ccache unless configured explicitly from my experience.
See http://stackoverflow.com/questions/1815688/how-to-use-ccache-with-cmake
It works and I see no reason not to enable it if it's available.

adam-nielsen:
Are you sure cmake ignores ccache?  The accepted answer on that Stack Overflow post is about how to enable it system wide so it works across all builds, cmake or otherwise.  I'm not saying there's anything wrong with the patch, just that it seems redundant to specifically add ccache support to one project, just to cater for those machines where ccache isn't configured properly.

I've just done some tests on my own system and `cpr` is definitely using `ccache` already:
```
$ make clean
$ time make -j1
real    0m22.782s
$ make clean
$ ccache -C
$ time make -j1
real    1m43.042s
```
Out of curiosity, how does your patch work with `colorgcc`?  For example I have `g++` run `colorgcc` which then runs `ccache` which then runs the real compiler, so `cpr` currently benefits from both cached compilation and more readable coloured error messages.  Does this patch still respect my configuration or does it override my choice and call `ccache` directly, bypassing `colorgcc` and causing me to lose my coloured compile errors?

whoshuu:
@thedrow, love the suggestion. Hadn't thought about using `ccache` on the CI server before. I think @adam-nielsen is right about leaving that out of the `CMakeLists.txt` file though. I just ran some local tests and the cache works just fine when setting the path to pick up `ccache` before the compilers, with no modifications to the base `CMakeLists.txt` file at all.

Can we give that a try? According to https://docs.travis-ci.com/user/caching/#ccache-cache, all you need to do is set `cache: ccache` and it'll handle setting the `PATH` for you, which is pretty nifty. Since that's already done, let's remove the changes to `CMakeLists.txt` and let travis give it a go. We can then check the build logs to make sure `ccache` is used in the compiler path in the `cmake` output.

thedrow:
Done...

whoshuu:
Checks out! Thanks @thedrow for the PR! Merging in now.

PR #117, opened 2016-07-29
Run compilation and tests with 2 jobs

thedrow:
Travis has ~2 cores and it'd be a shame not to use them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,7 +117,7 @@ before_script:
   - cmake -LAH ..
 
 script:
-  - make VERBOSE=1 -- -j2
+  - make -j2 VERBOSE=1
   - ctest -j2 -V
     
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,7 +118,7 @@ before_script:
 
 script:
   - make -j2 VERBOSE=1
-  - ctest -j2 -V
+  - ctest -V
     
 after_success:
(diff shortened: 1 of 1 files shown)


thedrow:
Looks like there's a problem with running tests in parallel. Any hints on how to fix this?

nabijaczleweli:
Better option: `-j$(nproc)`

thedrow:
Not on travis. It doesn't report the number of cpus correctly.

nabijaczleweli:
How so?

thedrow:
Because their virtualization technology doesn't override the /proc and /etc/cpuinfo filesystem correctly.

whoshuu:
Hey @thedrow, this looks good. I think running the build in parallel definitely makes sense. Should we just run it with `-j`, where no second argument is specified? `make` will run as many jobs as the system would be able to handle.

As for the tests, it looks like there's no real consistency in which tests are failing, which tells me the issue is in how the test driver (`ctest`) loads and runs the test executables. The embedded mongoose server does use some statics if I recall, so that might be where the issue is? This would require some more investigation.

For now, I think the best thing to do is to remove parallelism for `ctest` and bump it up to the max for `make`. We get definite improvements now and can figure out the other issue later.

What are your thoughts on this?

thedrow:
-j will spawn 32 jobs in Travis as there is no correct CPU information on that environment.
I can remove the parallelism from ctest.

whoshuu:
Looks good, thanks @thedrow for the PR! Merging in now.

Issue #123, opened 2016-08-13
Cookies cannot be traversed


eXeC64:
Currently, to access a cookie you need to know its key. As such, you cannot iterate through all the cookies, listing them unless you know which cookies are already set.

As far as I can tell, the simplest solution is just to implement `begin()` and `end()` methods that wrap the `map_`'s `begin()` and `end()`,  and provide `using iterator = std::map<std::string, std::string>::iterator;` in the class' definition.

COM8:
Fixed by #365.

Issue #107, opened 2016-06-19
Adding cookies and getting current cookies of sessions


lapinozz:
I would like to be able to retrieve cookies from a session and merge them to others or simply add cookies to a session without overwritting currents one.

COM8:
#365 should fix this.

Issue #102, opened 2016-05-26
how does merge cookie


charlessoft:
have two cpr::Session , so have two different cookies, how does merge the cookies?

charlessoft:
how does traversal the cpr::Cookie

whoshuu:
Currently there isn't a way to merge or traverse the cookies without modifying source code. If you want, you can change the private `map_` [here](https://github.com/whoshuu/cpr/blob/3fc28f0ea332a0423474be9077e91f4146d83661/include/cpr/cookies.h#L21) to be public and iterate through it through the usual means since it's just a `.std::map`.

COM8:
I'm closing this, since it seems to be fixed by #365.
Feel free to reopen this, when needed.

PR #124, opened 2016-08-13
Added iterator interface to Cookies class.

nshcat:
Proposed fix for issue #123 .

### DIFF
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -27,4 +27,60 @@ std::string& Cookies::operator[](const std::string& key) {
     return map_[key];
 }
 
+std::string& Cookies::at(const std::string& key) {
+    return map_.at(key);
+}
+
+const std::string& Cookies::at(const std::string& key) const {
+    return map_.at(key);
+}
+
+auto Cookies::begin() -> iterator {
+    return map_.begin();
+}
+
+auto Cookies::end() -> iterator {
+    return map_.end();
+}
+
+auto Cookies::begin() const -> const_iterator {
+    return map_.begin();
+}
+
+auto Cookies::end() const -> const_iterator {
+    return map_.end();
+}
+
+auto Cookies::cbegin() const -> const_iterator {
+    return map_.cbegin();
+}
+
+auto Cookies::cend() const -> const_iterator {
+    return map_.cend();
+}
+
+auto Cookies::rbegin() -> reverse_iterator {
+    return map_.rbegin();
+}
+
+auto Cookies::rend() -> reverse_iterator {
+    return map_.rend();
+}
+
+auto Cookies::rbegin() const -> const_reverse_iterator {
+    return map_.rbegin();
+}
+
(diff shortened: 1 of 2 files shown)


whoshuu:
Hey @nshcat, thanks for the pull request. I've been swamped recently with work stuff, but I've gotten some spare cycles to review pull requests.

One question about what you've added here. I'm not a language expert, not a compilers wizard, so if you could tell me the portability implications of your `auto method_name() -> return_type` syntax, that would really help move things along. Is this a c++14 construction? What compilers currently support this? What's the complexity/cost of degrading this back to an older syntax?

Any help would be appreciated. Thanks once again!

nshcat:
Well, it is a C++14 feature. I didn't realize that when I first wrote this, so my apologies for that.

The syntax is called "trailing return type". One example in my pull request:

nshcat:
I just realized that the "trailing return type" syntax I used is indeed part of C++14. Welp, I guess that happens when one starts C++ with C++14.

That syntax is basically used as syntactic sugar in a few egde cases when using a class-local type alias in a method's implementation, like so:

```c++

**Comment by nshcat** [comment]
I just realized that the "trailing return type" syntax I used is indeed part of C++14. Welp, I guess that happens when one starts C++ with C++14.

That syntax is basically used as syntactic sugar in a few egde cases when using a class-local type alias in a method's implementation, like so:

```c++
// Header
class X
{
      using my_type = int;

      my_type test();
};

// Source file
X::my_type X::test() { /* ... */  };
```

This is the way this would be done in C++11. Note that we need to specify the scope wherein `my_type` resides, because at the point where `X::my_type` is parsed, `X::test` was not yet encountered and thus the compiler does not know yet that we are in fact inside class `X` . (C++ is parsed line by line, from left to right).

Trailing return type can fix this, as we are declaring the return type _after_ mentioning the containg class, thus making the implementation look less cluttered:

But yea, since this is C++14 I am going to replace it with the C++11 implementation.

erengy:
I wonder if it'd be a better idea to ditch the private `map_` at this point and use `class Cookies : public std::map<std::string, std::string>` instead. It'd introduce modifier functions too, which may help with issues such as #102 and #107.

whoshuu:
@erengy, I think you're probably right in this case. I looked over the [RFC for cookies](https://tools.ietf.org/html/rfc6265) and it looks like some cookies are case insensitive. That's unfortunate but the current implementation inside cpr doesn't use the case insensitive map in the first place. There's no added disadvantage to just exposing a map as a public field in the `Cookies` class.

One question I'd like to generate more discussion on is if we should just make `Cookies` a typedef of a `std::map<std::string, std::string>`. That would reduce some complexity at the cost of losing some utility (such as having to make `GetEncoded` a non-member function). What do you all think of that?

dima74:
@whoshuu, I think we should make `Cookies` a typedef of a `std::map<std::string, std::string>`. It would be very elegant.

COM8:
Fixed by #365 .

PR #127, opened 2016-08-16
fix: when http head is "Auth: \n"

msly:
fix one bug
urlEncode add some unEncode word


whoshuu:
I think this PR has a few too many things in it that make it hard to get this merged in. Let's focus on one feature/bugfix at a time.

Once #143 is merged in, the library will handle `Auth: \n` just fine. As for supporting specifying cookies in a file, that's an excellent feature and should be done separately to the bugfix.

The effort is much appreciated though, but for now I think this should be split up a bit more.

PR #129, opened 2016-08-22
Add support for POST file uploads from in-memory buffers

adam-nielsen:
This patch adds support for supplying a file's contents in a memory buffer, so that if you already have the file in memory, you don't need to write it out to a temporary file before you can upload it.

The buffer is specified as iterator ranges, so it works with both STL containers and C-style arrays, without making an extra copy of the data.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -180,17 +180,40 @@ void Session::Impl::SetMultipart(Multipart&& multipart) {
         struct curl_httppost* lastptr = NULL;
 
         for (auto& part : multipart.parts) {
-            auto content_option = CURLFORM_COPYCONTENTS;
-            if (part.is_file) {
-                content_option = CURLFORM_FILE;
-            }
-            if (part.content_type.empty()) {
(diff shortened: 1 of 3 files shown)


skystrife:
I'm pretty sure this is dangerous if the `Iterator` is anything other than a `char*`. The implicit assumption being made here is that the data is contiguous in memory, which won't necessarily be true for lots of STL containers (think things like `std::list`).

I think the only safe STL containers here are likely `std::vector`, `std::array`, and `std::string`.

See e.g. http://think-async.com/Asio/asio-1.10.6/doc/asio/reference/buffer.html

adam-nielsen:
Hmm good point.  Any suggestions how to restrict it to a contiguous type, while allowing both STL containers and raw pointers to be used?  Preferably without a hundred overloaded functions!

niklas88:
@adam-nielsen I think there will be a concept for this in C++17, I don't know whether there is an enable_if based way

whoshuu:
Merged, thanks @adam-nielsen!!

PR #130, opened 2016-08-22
Documentation update for PR #129 - POST file from memory buffer

adam-nielsen:
Add examples for POSTing a memory buffer as a file upload, to explain how to use the changes in PR #129.


whoshuu:
This is just what we needed, thanks @adam-nielsen for the PR!

Issue #128, opened 2016-08-22
Memory leak - curl_formfree() not used after multipart POST


adam-nielsen:
`curlform_add()` is used in `session.cpp` to build forms, however [according to the curl documentation](https://curl.haxx.se/libcurl/c/curl_formfree.html) the `formpost` variable must be passed to `curlfree()` after the form has been submitted.

cpr does not do this, so it leaks a little memory every time a multipart form is POSTed.

dlee-libo:
I think store formpost in CurlHolder would be better.  And unique_ptr with  custom deleter is much cleaner.
[I just did it in my fork](https://github.com/dlee-libo/cpr/commit/8f3979eb53290cc85ff10d5e648e966606b1ca0e)

adam-nielsen:
That would work too - either fix is fine with me.

whoshuu:
Fixed in https://github.com/whoshuu/cpr/commit/075fd0dce606c584d6d5eb5581e09437c908c778.

PR #131, opened 2016-08-23
Issue 128 - fix memory leak in multipart form

adam-nielsen:
This is a fix for #128 which calls `curl_formfree()` on the value created by `curl_formadd()` to prevent memory from leaking after a multipart POST upload.

Only commit d39993c is needed, the other commits are because I based this request on top of PR #129.  If PR #129 is merged, then this will also merge cleanly on top of that with only a single commit.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -180,17 +180,40 @@ void Session::Impl::SetMultipart(Multipart&& multipart) {
         struct curl_httppost* lastptr = NULL;
 
         for (auto& part : multipart.parts) {
-            auto content_option = CURLFORM_COPYCONTENTS;
-            if (part.is_file) {
-                content_option = CURLFORM_FILE;
-            }
-            if (part.content_type.empty()) {
-                curl_formadd(&formpost, &lastptr, CURLFORM_COPYNAME, part.name.data(),
-                             content_option, part.value.data(), CURLFORM_END);
+            if (part.is_buffer) {
+                if (part.content_type.empty()) {
+                    curl_formadd(&formpost, &lastptr,
+                                 CURLFORM_COPYNAME, part.name.data(),
+                                 CURLFORM_BUFFER, part.value.data(),
+                                 CURLFORM_BUFFERPTR, part.data,
+                                 CURLFORM_BUFFERLENGTH, part.datalen,
+                                 CURLFORM_END);
+                } else {
+                    curl_formadd(&formpost, &lastptr,
+                                 CURLFORM_COPYNAME, part.name.data(),
+                                 CURLFORM_BUFFER, part.value.data(),
+                                 CURLFORM_BUFFERPTR, part.data,
+                                 CURLFORM_BUFFERLENGTH, part.datalen,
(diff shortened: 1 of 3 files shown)


cpr/session.cpp
```diff
@@ -47,13 +48,15 @@ class Session::Impl {
     Url url_;
     Parameters parameters_;
     Proxies proxies_;
+    struct curl_httppost* formpost_;
```
wilhelmtell: I don't think this is a good idea. You're placing here a raw pointer that is owning.

This demands you add a destructor, which in turn demands you add custom copy and move semantics. I think it's more code than you have to. Place the pointer in a unique_ptr, initialize it with the `curl_formfree()` function, and then you don't need any of the destructor, copy and move functions. You also don't need the check for null in the cleanup, because that's already done in `curl_formfree()`.

I haven't tried coding this, but I imagine that when you do, you'll see that `Session` loses its copy semantics, because `std::unique_ptr<>` doesn't allow copying. This in turn won't compile if `Session` needs copy semantics. Does it? Either way, this is a great way to find a bug at compile time. Because this PR doesn't add the (otherwise) necessary copy semantics, so if you'd copy a `Session` now, you'll have a double-free, and that's UB.

whoshuu: I don't disagree with your assessment here, just a minor point of clarification. `Session` already doesn't have copy semantics because it has an owning `std::unique_ptr` to `CurlHolder`. This can actually be leveraged to achieve memory safety for forms in an elegant way, and I'll explain it more in a comment below.

whoshuu:
Hey @adam-nielsen, I appreciate your work on this memory leak. I think the approach is mostly right (getting `Session` to be RAII compliant at the interface level), but the implementation can be simplified considerably.

There's already a pattern for freeing curl-ish pointers on destruction of the `Session` object that you can leverage to achieve memory safety here for forms. See the `CurlHolder` definition [here](https://github.com/whoshuu/cpr/blob/04222915ece1850e54c7381eb55316a32016d93c/include/cpr/curlholder.h#L12), specifically for the pointer `chunk`.

That same `chunk` is set whenever a header is added [here](https://github.com/whoshuu/cpr/blob/04222915ece1850e54c7381eb55316a32016d93c/cpr/session.cpp#L124), and it gets freed automatically in the static call to `freeHolder` [here](https://github.com/whoshuu/cpr/blob/04222915ece1850e54c7381eb55316a32016d93c/cpr/session.cpp#L89).

Using that pattern to fix the memory leak would also remove the dependency between this specific pull request and the one for the buffer form POST, which makes it a lot easier to get this into the master branch.

Let me know if you disagree with the approach I've outlined, if you have a simpler solution than that I'm all ears.

cpr/session.cpp
whoshuu: You are correct though that adding the destructor for the sake of deleting that raw pointer would remove move semantics, which `Session` currently has.

whoshuu:
The leak is now fixed in https://github.com/whoshuu/cpr/commit/075fd0dce606c584d6d5eb5581e09437c908c778, thanks @adam-nielsen for the PR!

PR #132, opened 2016-08-27
Adding possibility to install library and find_package it

daminetreg:
Dear @whoshuu,

I'm proposing you this small addition to your cmake files in order to allow seamless installation, find_packageing and package generation of CPR.

This is really useful when building systems depending on the library, for example we build IoT devices based on linux, and as we have multiple apps depending on CPR, we would like to compile it and install it once in our development sysroot and depend on it from our different application.

This also ease creating a package with CPack, that you can then distribute to developer as precompiled development package. Or even integrating your library in the [CMake Package Manager](https://github.com/ruslo/hunter)

For example, thanks to this I can now install and use cpr this way in my project : 

``` cmake
cmake_minimum_required(VERSION 3.0.0)

project(testcpr)

set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

find_package(cpr 1.2.1 REQUIRED)
add_executable(testcpr main.cpp)
target_link_libraries(testcpr cpr::cpr)
```

Thank your for developing this wrapper, I have 2 colleagues who really appreciate the simplicity of the API. 

I would be happy to know what do you think from this patch. :smile:


daminetreg:
Just noticed PR #56 was bringing a similar feature, I think this implementation is cleaner, but that's up to you. The package generation is based on the template of ruslo : https://github.com/forexample/package-example which is in my experience the cleanest way to make packages in modern cmake.

daminetreg:
I'll fix the build error, when building CURL from within the submodules. 

Some of your CI job requires however CMake older than 3.0.0. It is something you want so ? Meaning I should fix it, or don't you want to upgrade to CMake 3.0.0. I ask this, because package config generation works really in 3.0.0 and not that good before.

tastytea:
Please use [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). It is portable and the variables can be overridden.

COM8:
I’m closing this.

PR #141, opened 2016-09-27
Remove response processing

### LINKED ISSUE #140: text files are modified if downloading via range requests
line 379 of session.cpp has:
```c++
    response_string = cpr::util::parseResponse(response_string);
```
the method's definition:
```c++
std::string parseResponse(const std::string& response) {
    if (!response.empty()) {
        if (response.back() == '\n') {
            return response.substr(0, response.length() - 1);
        }
    }

    return response;
}
```
It strips newlines at the end of a response, this caused a carriage return to be stripped from a text file downloaded in sections.
I commented this line out and now the downloaded file matches the source file.
I'm not sure why this is done? Could i be causing more harm than good?

nabijaczleweli:
Closes #140

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -381,7 +381,6 @@ Response Session::Impl::makeRequest(CURL* curl) {
     curl_slist_free_all(raw_cookies);
 
     auto header = cpr::util::parseHeader(header_string);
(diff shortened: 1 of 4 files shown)


whoshuu:
Thanks @nabijaczleweli! This looks great, merging it in!

PR #142, opened 2016-10-05
Move constructor from std::string for cpr::Body

pravic:
Currently it uses only copy-ctors.


whoshuu:
Thanks for the PR @pravic. Can you add a test of this specific constructor? Somewhere in [raw_body_tests.cpp](https://github.com/whoshuu/cpr/blob/master/test/raw_body_tests.cpp) is probably fine.

pravic:
Cannot now, because I deleted cpr's repository. I'll look what can I do.

PR #147, opened 2016-10-16
Updated submodule URL to https instead of SSH

himanshub16:
git@github.com:whoshuu/cpr.git used SSH which gave authentication errors on **repository rights**.
However, https://github.com/whoshuu/cpr.git worked. Please check the issue from some account other than your own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ For just getting this library up and running, I highly recommend forking the [ex
 If you already have a project you need to integrate C++ Requests with, the primary way is to use git submodules. Add this repository as a submodule of your root repository:
 
 ```shell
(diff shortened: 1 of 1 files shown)


whoshuu:
I appreciate the patch, but I think the more appropriate thing to do here is to note in the README that submodules might be better added against their HTTPS urls as opposed to their SSH urls. 

In either case, it's not the responsibility of the project to teach proper `git submodule` usage for any particular user's environment. There are plenty of resources for that.

My (possibly unreasonable / caffeine induced) opinion is that SSH is a better default because I hate passwords. But to each their own!

himanshub16:
AFAIK, Git clone / addition doesn't require password, unless the repository is private on GitHub. However, I've added both the methods.

PR #153, opened 2016-11-11
Disable build of curl executable.

xpol:


opt/CMakeLists.txt
```diff
@@ -52,7 +53,6 @@ if(NOT USE_SYSTEM_CURL OR NOT CURL_FOUND)
         ${CURL_BINARY_DIR}/include/curl)
         
     # Group under the "external" project folder in IDEs such as Visual Studio.
-    set_property(TARGET curl PROPERTY FOLDER "external")
```
whoshuu: I think this folder should be created if `BUILD_CURL_EXE` is set to `ON`. What do you think?

whoshuu: It looks like #97 already does this. I'll merge in both at some point today :+1:.

PR #156, opened 2016-11-27
Update git submodules path in .gitsubmodule

DanySpin97:
Change submodules path to make Github pages build correctly as explained by Github documentation.
Check here for more info (https://help.github.com/articles/page-build-failed-invalid-submodule/).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "opt/curl"]
 	path = opt/curl
-	url = git://github.com/whoshuu/curl.git
+	url = https://github.com/whoshuu/curl.git
 [submodule "opt/googletest"]
 	path = opt/googletest
-	url = git://github.com/whoshuu/googletest.git
+	url = https://github.com/whoshuu/googletest.git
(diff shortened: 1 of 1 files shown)


whoshuu:
Maybe we can just delete the submodules from the github pages branch? None of the submodules are actually used in the document pages. I'd rather not change the submodule url's from ssh, it's a convenience that's hard to give up, and I prefer the security features of ssh over having to type the password every now and then for https based git urls.

DanySpin97:
There is no need to use an ssh key neither a password to clone open sourced repository like the gitsubmodules cpr has. Changing the url of submodules would just change url of submodule leaving the repo as it. You could still use ssh to clone cpr as you're doing now.
Documentation has not to be in gh-pages branch but it can also be in /docs path of master branch, leading to this building error.

whoshuu:
Force of habit I guess. I normally work on private repos and prefer the convenience of ssh over https. But I can understand why having https would work better for most people in this particular situation.

Additionally, I shouldn't work on the submodules from inside of cpr anyway, so https is a better default in this case.

PR #160, opened 2016-12-21
WIP:add multiple build targets

ethanhs:
This adds multiple build targets (static and shared) and builds on Windows with WinSSL if you have CURL installed in the system directory C:\Program Files (x86)\CURL. I will need to look at how the build machine deals with building, as I don't know why that last test would fail...

Also, this adds static and shared targets.

This PR is a WIP and should __NOT__ be merged.


apollo13:
I don't think building static and shared is what is needed in the general case. I'd very much prefer if the existing `STATIC` just got dropped and the user is put into the choice via: https://cmake.org/cmake/help/v3.8/variable/BUILD_SHARED_LIBS.html

ethanhs:
Yeah, I was using cpr for a project, but ended up not using it, so I dont have time to work on this anyway, if someone wants to take over, be my guest.

apollo13:
@ethanhs What did you end up with instead if you don't mind me asking? I am currently also toying with cpr and open to alternatives.

ethanhs:
The project I was using ended up not happening, so nothing. But using plain curl is decent. The cpprestsdk is pretty nice, but I like cpr.

ethanhs:
I have lost time and interest in working on this, so I am closing.

PR #165, opened 2016-12-28
Fix compilation warning. Silence some -Wextra warnings. Fix travis.

coin-au-carre:
I just hided an unused parameter warning : `is_random_access_iterator(Iterator /* begin */, Iterator /* end */ ) `

I have added some compile flags that are used in the projects I work on and hide them in CPR so that they do not show anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,11 @@ else()
     endif()
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic") # add warning flags
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic") # if compiler is GCC or Clang add warning flags
+endif()
+
 set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR}/bin)
 set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR}/lib)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,11 +35,6 @@ else()
     endif()
 endif()
 
-
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic") # if compiler is GCC or Clang add warning flags
-endif()
-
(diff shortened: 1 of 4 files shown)


coin-au-carre:
Ok I will remove the extra flag on Windows which does not recognize it...

test/server.cpp
```diff
@@ -369,7 +368,7 @@ static int deleteRequest(struct mg_connection* conn) {
     auto has_json_header = false;
     for (int i = 0; i < num_headers; ++i) {
         auto name = headers[i].name;
-        if (std::string{"Content-Type"} == headers[i].name &&
+        if (std::string{"Content-Type"} == name &&
```
whoshuu: Hmm, it looks like `name` isn't used anywhere but here. Let's just replace it with `headers[i].name` and remove the above variable definition.

CMakeLists.txt
```diff
@@ -35,6 +35,11 @@ else()
     endif()
 endif()
 
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic") # if compiler is GCC or Clang add warning flags
+endif()
+
```
whoshuu: I think setting warning flags is a faux pas in a library project like CPR. Most people use this library inside of another cmake project where they are setting their own warning flags.

Perhaps a good compromise is to set these warning flags in the travis build config so that we can see the warnings in those builds? Let me know what you think.

In either case I'm unlikely to merge in this PR if it's setting warning flags like this.

test/server.cpp
```diff
@@ -164,12 +164,11 @@ static int basicAuth(struct mg_connection* conn) {
     auto response = std::string{"Hello world!"};
     const char* requested_auth;
     auto auth = std::string{"Basic"};
-    std::string auth_string;
     if ((requested_auth = mg_get_header(conn, "Authorization")) == NULL ||
         mg_strncasecmp(requested_auth, auth.data(), auth.length()) != 0) {
         return MG_FALSE;
     }
-    auth_string = {requested_auth};
+    auto auth_string = std::string{requested_auth};
```
whoshuu: :+1:

include/cpr/multipart.h
```diff
@@ -33,7 +33,7 @@ struct Buffer {
     template <typename Iterator>
     typename std::enable_if<std::is_same<typename std::iterator_traits<Iterator>::iterator_category,
                                          std::random_access_iterator_tag>::value>::type
-    is_random_access_iterator(Iterator begin, Iterator end) {}
+    is_random_access_iterator(Iterator /* begin */, Iterator /* end */ ) {}
```
whoshuu: :+1:

CMakeLists.txt
coin-au-carre: Sure I understand. Thanks for your answer. I will remove theses changes and solve the warning issues in my project since I have these compilation flags. I do not think they are necessary either.

whoshuu:
Looks good, thanks @coin-au-carre!

PR #167, opened 2017-01-02
Update styles.css

pierriko:
fix code style, add "mono" to code/pre (on Fedora).

before:

screenshot from 2017-01-02 17-46-52

after:

screenshot from 2017-01-02 17-47-14


whoshuu:
Thanks for catching this @pierriko!

PR #176, opened 2017-01-24
Improve cmake installation

nyorain:
Currently, cmake will not generate any build commands to install cpr header files or libraries.
With these small changes, the install target generated by cmake will install the header files to include/cpr and the libraries to lib (static) or bin (shared).
This is very useful when pulling in cpr as a dependency in another project.


cpr/CMakeLists.txt
```diff
@@ -45,3 +45,8 @@ add_library(${CPR_LIBRARIES}
 message(STATUS "Using CURL_LIBRARIES: ${CURL_LIBRARIES}.")
 target_link_libraries(${CPR_LIBRARIES}
     ${CURL_LIBRARIES})
+    
+install(TARGETS ${CPR_LIBRARIES}
```
whoshuu: I just tried this. This installs all of this:

>-- Installing: /usr/local/include/cpr
-- Installing: /usr/local/include/cpr/api.h
-- Installing: /usr/local/include/cpr/auth.h
-- Installing: /usr/local/include/cpr/body.h
-- Installing: /usr/local/include/cpr/cookies.h
-- Installing: /usr/local/include/cpr/cpr.h
-- Installing: /usr/local/include/cpr/cprtypes.h
-- Installing: /usr/local/include/cpr/curlholder.h
-- Installing: /usr/local/include/cpr/defines.h
-- Installing: /usr/local/include/cpr/digest.h
-- Installing: /usr/local/include/cpr/error.h
-- Installing: /usr/local/include/cpr/low_speed.h
-- Installing: /usr/local/include/cpr/max_redirects.h
-- Installing: /usr/local/include/cpr/multipart.h
-- Installing: /usr/local/include/cpr/parameters.h
-- Installing: /usr/local/include/cpr/payload.h
-- Installing: /usr/local/include/cpr/proxies.h
-- Installing: /usr/local/include/cpr/response.h
-- Installing: /usr/local/include/cpr/session.h
-- Installing: /usr/local/include/cpr/ssl_options.h
-- Installing: /usr/local/include/cpr/timeout.h
-- Installing: /usr/local/include/cpr/util.h
-- **Installing: /usr/local/bin/curl-config**
-- **Installing: /usr/local/lib/pkgconfig/libcurl.pc**
-- **Installing: /usr/local/include/curl/curlbuild.h**
-- **Up-to-date: /usr/local/include/curl**
-- **Installing: /usr/local/include/curl/curl.h**
-- **Installing: /usr/local/include/curl/curlrules.h**
-- **Installing: /usr/local/include/curl/curlver.h**
-- **Installing: /usr/local/include/curl/easy.h**
-- **Installing: /usr/local/include/curl/mprintf.h**
-- **Installing: /usr/local/include/curl/multi.h**
-- **Installing: /usr/local/include/curl/stdcheaders.h**
-- **Installing: /usr/local/include/curl/typecheck-gcc.h**
-- **Installing: /usr/local/lib/libcurl.dylib**
-- Installing: /usr/local/lib/libcpr.a

The bolded lines might be problematic. Is there a way to restrict the install to only copy the `libcpr.a` library? Or to add an additional option (which would default to `OFF`) that exports curl with cpr during install?

nyorain: oh, overlooked that. That would indeed be problematic.
The curl headers and libraries are installed even without the patch (tried with master) since curl is
simply added as subdir, so it is not possible at the moment to not install it.

privateOmega: @nyorain Have you found a solution to it?

nyorain: nope, not using cmake anymore

PR #178, opened 2017-01-25
Add pthread linking in tests

### LINKED ISSUE #177: Build fails on Arch Linux
Fails during linking:
```
[ 31%] Linking CXX executable ../bin/alternating_tests
/usr/bin/ld: CMakeFiles/alternating_tests.dir/server.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/lib/libpthread.so.0: error adding symbols: DSO missing from command line
```
Fixed it by adding pthread to target_link_libraries in test/CMakeLists.txt but of course it's not very portable.

noh4hss:
Fixes #177 problem with pthread linking in a portable way. 
[https://stackoverflow.com/questions/1620918/cmake-and-libpthread](https://stackoverflow.com/questions/1620918/cmake-and-libpthread)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,10 +1,12 @@
 macro(add_cpr_test _TEST_NAME)
+    find_package(Threads)
     add_executable(${_TEST_NAME}_tests
         ${_TEST_NAME}_tests.cpp server.cpp server.h)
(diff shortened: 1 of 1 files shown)


whoshuu:
Perfect!

PR #184, opened 2017-02-07
Add support for extending a multipart object

noh4hss:
Currently the only way to build a Multipart object is to use a constructor that receives a initializer_list.
This makes doing a multipart POST request practically impossible when we want to decide at runtime whether to include some paremeter in the request or not.

This patch fixes it by adding to the Multipart class:
-a iterator constructor
-a method that allows to add a Part struct
 
It mirrors the interface of the Payload class.

### DIFF
--- a/cpr/multipart.cpp
+++ b/cpr/multipart.cpp
@@ -4,4 +4,8 @@ namespace cpr {
 
 Multipart::Multipart(const std::initializer_list<Part>& parts) : parts{parts} {}
 
+void Multipart::AddPart(const Part& part) {
+    parts.push_back(part);
+}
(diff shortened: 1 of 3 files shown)


tstack:
Not sure if you're still interested in this... But, I'm wondering if this change is really necessary since the `parts` field is currently public.  I realize you've modeled the changes after Payload, but the methods in Payload are doing some extra string formatting work.  Whereas these changes are just passing through to the vector methods.

COM8:
I'm closing this since there has not been a response from @noh4hss.
Feel free to reopen/create a new PR.

PR #187, opened 2017-02-28
Fix Set{Body,Payload,Mulitpart} implementation

noh4hss:
I'm gonna be talking about problems with the pair of SetBody functions but the same thing applies to SetPayload and SetMultipart.
The following code demonstrates a bug in SetBody(const Body&):
```
    cpr::Session session;
    session.SetUrl(cpr::Url{"http://httpbin.org/post"});
    
    auto body = cpr::Body{"a"};
    session.SetBody(body);
    body = cpr::Body{"b"};
    
    auto r = session.Post(); // sends request with body equal to "b"
    std::cout << r.text << std::endl;
    
        
    if(true) {
        auto new_body = cpr::Body{"c"};
        session.SetBody(new_body);
    }
    
    r = session.Post(); // during this call libcurl uses pointer to freed memory
```
The current implementation doesn't instruct libcurl to copy data. As the above code suggests, we have to copy data in this SetBody function. 
However we can omit copying in the SetBody(Body&&) function but the current implementation of this function copies data. We can't do it by simply swapping implementations of SetBody(const Body&) and SetBody(Body&&). We have to take ownership of data that a Body object represents to be sure we don't have a pointer to freed memory when a request is actually done. We do this by adding a std::string variable to the Session::Impl class and using the move assignment operator of std::string. SetPayload(Payload&&) also uses this new string variable. For SetMultipart(Multipart&&) we add a Multipart object to the Session::Impl.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -49,6 +49,7 @@ class Session::Impl {
     Url url_;
     Parameters parameters_;
     Proxies proxies_;
+    std::string body_;
 
     Response makeRequest(CURL* curl);
     static void freeHolder(CurlHolder* holder);
@@ -151,8 +152,9 @@ void Session::Impl::SetDigest(const Digest& auth) {
 void Session::Impl::SetPayload(Payload&& payload) {
     auto curl = curl_->handle;
     if (curl) {
-        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, payload.content.length());
-        curl_easy_setopt(curl, CURLOPT_COPYPOSTFIELDS, payload.content.data());
+        body_ = std::move(payload.content);
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, body_.length());
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDS, body_.data());
     }
 }
 
@@ -263,8 +265,9 @@ void Session::Impl::SetCookies(const Cookies& cookies) {
 void Session::Impl::SetBody(Body&& body) {
     auto curl = curl_->handle;
     if (curl) {
-        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, body.length());
-        curl_easy_setopt(curl, CURLOPT_COPYPOSTFIELDS, body.data());
+        body_ = std::move(body);
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, body_.length());
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDS, body_.data());
     }
 }
 
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -50,6 +50,7 @@ class Session::Impl {
     Parameters parameters_;
     Proxies proxies_;
     std::string body_;
+    Multipart multipart_;
 
     Response makeRequest(CURL* curl);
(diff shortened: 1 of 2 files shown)


whoshuu:
I'm not sure I'm following along right here. Calls to `SetBody(const Body&)` aren't supposed to take ownership of the data. It's a reference to existing data, and if that data changes, then at the time of request the body will change as well.

Using `SetBody(const Body&)` is efficient in that the data is not copied, but the cost of that is that the programmer has to be careful about keeping that data around until after the request completes. This is the same advice that's given in the libcurl docs:

>The data pointed to is NOT copied by the library: as a consequence, it must be preserved by the calling application until the associated transfer finishes. This behaviour can be changed (so libcurl does copy the data) by setting the CURLOPT_COPYPOSTFIELDS option.

The only time the library can be sure that the data being passed in is not kept around by the client is if it's passed by rvalue reference. Only in that case should the data be copied by the library.

noh4hss:
I am saying that ```SetBody(const Body&)``` should copy and ```SetBody(Body&&)``` can avoid copying by taking ownership of the data.

> the programmer has to be careful about keeping that data around until after the request completes

I think it is a really bad idea for a function in c++ library that takes const reference of an object. Also the passed object cannot be changed. For example setting body to something longer after call to the SetBody() may result in segfault during the request.
When a function takes rvalue reference we can actually avoid copying. In my solution if a programmer does not want to copy then he can call SetBody(std::move(body)). 

Again I think it is reasonable to expect that the code I gave above should work. That mean ```SetBody(const Body&)``` must copy. I think our primary focus should be simplicity of use, not avoiding copying by introducing completely non-intuitive behavior. We can add one more function that takes const reference and does not copy which will allow a programmer to avoid copying while not destroying object by moving it.

whoshuu:
I appreciate the feedback, but I respectfully disagree.

You're saying that this should work:

```c++
if (true) {
    auto new_body = cpr::Body{"c"};
    session.SetBody(new_body);
}
```

Would this block of code not be strictly improved by changing it to this:

```c++
if (true) {
    session.SetBody(cpr::Body{"c"});
}
```

The danger in the first call was that you were passing in a reference that might expire by the time the request is made.

The fact that it was expiring is exactly why there's an rvalue reference overload for that interface. If data is going to expire, then it should be moved directly into the session. Right now that "move" is actually a copy, because curl doesn't know about strings, it only knows about char*. There's some room for optimization there if session has it's own string body field.

But I still think const& should be a non-copied operation. A priori assumptions about what constitutes a good interface or a bad one are less important than what the signature conveys.

A signature that takes const& as the argument explicitly conveys that the object passed in is a reference (it must outlive the caller), and that it cannot be modified.

Basically, I agree with you that correctness and usability is more important than efficiency. I disagree that the intuitive response to a const& parameter is to pass it a temporary, especially when an explicit interface that takes temporaries exists as well.

tstack:
I would kinda lean toward the original premise of this PR.  People probably don't expect a method that takes an object by const reference to retain a pointer to that object.  They probably just expect the object to be used in whatever processing the method was doing and not used after that.

COM8:
@noh4hss I'm planing to merge this PR since the current behaviour it's introducing interesting behaviour.

COM8:
Could you please resolve the conflicts?

COM8:
Feel free to open a new issue.

PR #188, opened 2017-02-28
fix SetHeader() implementation

### LINKED ISSUE #144: Memory leak in session - header parsing
Hi,

There's a potential memory leak in session.cpp when parsing headers:
Here's a patch, but I don't think this is the best way to do it.

```
74d73
< #ifdef INSECURE_CURL
77c76
< #endif

---
>
115c114,115
<         struct curl_slist* chunk = NULL;

---
>         struct curl_slist* chunk = NULL, *temp = NULL;
>
116a117
>             temp = NULL;
123c124,130
<             chunk = curl_slist_append(chunk, header_string.data());

---
>
>             temp = curl_slist_append(chunk, header_string.data());
>             while(temp == NULL){
>                 temp = curl_slist_append(chunk, header_string.data());
>             }
>             chunk = temp;
>
```

For more information, see:
https://github.com/collectd/collectd/pull/1639 and https://github.com/collectd/collectd/pull/1671

Essentially
curl_slist_append() overwrote previous pointer which would leak if
any curl_slist_append() failed since it returns NULL in case of errors.

### LINKED ISSUE #185: Session::SetHeader() sets CURLOPT_HTTPHEADER multiple times
Didn't you mean this?
```diff
 void Session::Impl::SetHeader(const Header& header) {
     auto curl = curl_->handle;
     if (curl) {
         struct curl_slist* chunk = NULL;
         for (auto item = header.cbegin(); item != header.cend(); ++item) {
             auto header_string = std::string{item->first};
             if (item->second.empty()) {
                 header_string += ";";
             } else {
                 header_string += ": " + item->second;
             }
             chunk = curl_slist_append(chunk, header_string.data());
-            curl_easy_setopt(curl, CURLOPT_HTTPHEADER, chunk);
-            curl_->chunk = chunk;
         }
+        curl_easy_setopt(curl, CURLOPT_HTTPHEADER, chunk);
+        curl_->chunk = chunk;
     }
 }
```

noh4hss:
fixes #185, fixes #144

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -118,7 +118,11 @@ void Session::Impl::SetHeader(const Header& header) {
             } else {
                 header_string += ": " + item->second;
             }
-            chunk = curl_slist_append(chunk, header_string.data());
+
+            auto temp = curl_slist_append(chunk, header_string.data());
+            if (temp) {
+                chunk = temp;
+            }
         }
         curl_easy_setopt(curl, CURLOPT_HTTPHEADER, chunk);
 
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -119,9 +119,11 @@ void Session::Impl::SetHeader(const Header& header) {
                 header_string += ": " + item->second;
             }
             chunk = curl_slist_append(chunk, header_string.data());
-            curl_easy_setopt(curl, CURLOPT_HTTPHEADER, chunk);
-            curl_->chunk = chunk;
         }
+        curl_easy_setopt(curl, CURLOPT_HTTPHEADER, chunk);
+
+        curl_slist_free_all(curl_->chunk);
+        curl_->chunk = chunk;
     }
 }
 


whoshuu:
Hey @noh4hss, I really appreciate all of the time and work you've put in to making these pull requests and reviewing the codebase. I'll take out some time this weekend to review your things and get them into shape for merging.

cpr/session.cpp
```diff
@@ -119,9 +119,11 @@ void Session::Impl::SetHeader(const Header& header) {
                 header_string += ": " + item->second;
             }
             chunk = curl_slist_append(chunk, header_string.data());
```
whoshuu: If `curl_slist_append` fails, then the return value is `NULL`. Let's store the value in a temporary and only assign `chunk` to that temporary when it's not `NULL`. That way we can be sure that `chunk` always points to the latest valid tail in the slist, and subsequent frees will not leak memory.

PR #202, opened 2017-06-05
Use target_include_directories() in CMake

aldanor:
When using `cpr` in a CMake project, simply doing `add_subdirectory(path/to/cpr)` doesn't add curl and cpr include directories to target's public include path, as one might expect, so it has to be done manually.

This should fix that by attaching these include directories to the target as public.

### DIFF
--- a/cpr/CMakeLists.txt
+++ b/cpr/CMakeLists.txt
@@ -1,7 +1,4 @@
 message(STATUS "Using CURL_INCLUDE_DIRS: ${CURL_INCLUDE_DIRS}.")
-include_directories(
-    ${CPR_INCLUDE_DIRS}
-    ${CURL_INCLUDE_DIRS})
 
 add_library(${CPR_LIBRARIES}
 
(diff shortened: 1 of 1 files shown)


aldanor:
(This is a very trivial fix, would appreciate if it gets merged soon).

cpr/CMakeLists.txt
```diff
@@ -45,3 +42,9 @@ add_library(${CPR_LIBRARIES}
 message(STATUS "Using CURL_LIBRARIES: ${CURL_LIBRARIES}.")
 target_link_libraries(${CPR_LIBRARIES}
     ${CURL_LIBRARIES})
+target_include_directories(
+        ${CPR_LIBRARIES}
+        PUBLIC
+        ${CPR_INCLUDE_DIRS}
+        ${CURL_INCLUDE_DIRS})
+
```
whoshuu: Hate to be a stickler but let's get rid of this extraneous newline

aldanor:
@whoshuu All makes sense -- removed the evil extra newline, added "is cmake >= 3.0" check, reverts to old behavior on pre-3.0 versions.

whoshuu:
@aldanor excellent stuff, thanks for making those changes!

PR #207, opened 2017-06-23
Fix padding of 'Error' and 'Response' classes.

Ristovski:
Gets triggered by '-Wpadded'.
Tested with clang 5.


Ristovski:
Checks have passed but it causes an internal segfault, seems like I missed something.

PR #210, opened 2017-07-19
Add HTTP/2 support

### LINKED ISSUE #116: HTTP2 support
Curl already supports HTTP2.
It'd be useful to have a client that automatically upgrades the connection to HTTP2 if the server supports it.

p0fi:
This adds support for the HTTP/2 protocol to cpr. To use it set the option `cpr::ProtocolVersion{cpr::HTTP::v2}`. If the underlying CURL doesnt support HTTP/2 it throws a runtime error informing you about this issue.

It would be cool if this could be added as a CMake option which also checks if the underlying CURL supports HTTP/2, but I don't know enough about CMake to make this happen. Maybe someone else could add this here too.

This will resolve #116

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -294,20 +294,22 @@ void Session::Impl::SetVerifySsl(const VerifySsl& verify) {
 }
 
 void Session::Impl::SetProtocolVersion(const ProtocolVersion& protocolversion) {
+#if LIBCURL_VERSION_MAJOR >= 7
+#if LIBCURL_VERSION_MINOR >= 33
+#if LIBCURL_VERSION_PATCH >= 0
     auto curl = curl_->handle;
     if (curl) {
-        if ( protocolversion == cpr::HTTP::v2 )
-        {
-            if ( curl_version_info(CURLVERSION_NOW)->features & CURL_VERSION_HTTP2 )
-            {
+        if (protocolversion == cpr::HTTP::v2) {
+            if (curl_version_info(CURLVERSION_NOW)->features & CURL_VERSION_HTTP2) {
                 curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
-            }
-            else
-            {
+            } else {
                 throw std::runtime_error("no HTTP/2 support");
             }
         }
     }
+#endif
+#endif
+#endif
 }
 
 Response Session::Impl::Delete() {
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -35,6 +35,8 @@ class Session::Impl {
     void SetBody(const Body& body);
     void SetLowSpeed(const LowSpeed& low_speed);
     void SetVerifySsl(const VerifySsl& verify);
+    void SetProtocolVersion(const ProtocolVersion& protocolversion);
+
 
     Response Delete();
     Response Get();
@@ -291,6 +293,23 @@ void Session::Impl::SetVerifySsl(const VerifySsl& verify) {
     }
 }
 
+void Session::Impl::SetProtocolVersion(const ProtocolVersion& protocolversion) {
+    auto curl = curl_->handle;
+    if (curl) {
+        if ( protocolversion == cpr::HTTP::v2 )
+        {
+            if ( curl_version_info(CURLVERSION_NOW)->features & CURL_VERSION_HTTP2 )
+            {
+                curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
+            }
+            else
+            {
+                throw std::runtime_error("no HTTP/2 support");
+            }
+        }
+    }
+}
+
 Response Session::Impl::Delete() {
     auto curl = curl_->handle;
     if (curl) {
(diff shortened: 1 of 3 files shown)


pravic:
May be it is better to use the `class enum`? Because currently it lacks of visibility:
```cpp
cpr::ProtocolVersion{cpr::v2} // v2 of what ???
```

p0fi:
I'm afraid I don't fully understand what you mean. Your code would not work because you are missing the `HTTP::`. Could you explain you you mean in more detail?

pravic:
It would work, of course. This code just desn't look good.

p0fi:
You are totally right, I missed that one. I added the class keyword to force the use of `cpr::HTTP::v2` 🙏

cpr/session.cpp
```diff
@@ -291,6 +293,25 @@ void Session::Impl::SetVerifySsl(const VerifySsl& verify) {
     }
 }
 
+void Session::Impl::SetProtocolVersion(const ProtocolVersion& protocolversion) {
+#if LIBCURL_VERSION_MAJOR >= 7
```
mway: noop'ing (or ignoring) this call silently based on the libcurl version is maybe a little surprising/unintuitive; thoughts on changing this to throw an exception if the protocol version isn't supported?

cpr/session.cpp
```diff
@@ -291,6 +293,25 @@ void Session::Impl::SetVerifySsl(const VerifySsl& verify) {
...
+#if LIBCURL_VERSION_MINOR >= 33
+#if LIBCURL_VERSION_PATCH >= 0
+    auto curl = curl_->handle;
+    if (curl) {
+        if (protocolversion == cpr::HTTP::v2) {
```
mway: currently only setting the protocol version if it's http/2; is it important to set http/1.0 or http/1.1? what if you want to reuse a `Session` across both http/1 and http/2 requests?

include/cpr/cprtypes.h
```diff
@@ -12,6 +12,7 @@ struct CaseInsensitiveCompare {
 
 using Header = std::map<std::string, std::string, CaseInsensitiveCompare>;
 using Url = std::string;
+using ProtocolVersion = enum class HTTP { v1x, v2 };
```
mway: do you think it's worthwhile to separate 1.0 and 1.1?

cpr/session.cpp
```diff
@@ -35,6 +35,8 @@ class Session::Impl {
     void SetBody(const Body& body);
     void SetLowSpeed(const LowSpeed& low_speed);
     void SetVerifySsl(const VerifySsl& verify);
+    void SetProtocolVersion(const ProtocolVersion& protocolversion);
+
```
whoshuu: Unnecessary extra line

cpr/session.cpp
```diff
@@ -291,6 +293,25 @@ void Session::Impl::SetVerifySsl(const VerifySsl& verify) {
     }
 }
 
+void Session::Impl::SetProtocolVersion(const ProtocolVersion& protocolversion) {
```
whoshuu: Prefer lowerCamelCase for all variable names, so `protocolVersion` is more consistent here

cpr/session.cpp
```diff
@@ -445,6 +466,7 @@ void Session::SetBody(const Body& body) { pimpl_->SetBody(body); }
 void Session::SetBody(Body&& body) { pimpl_->SetBody(std::move(body)); }
 void Session::SetLowSpeed(const LowSpeed& low_speed) { pimpl_->SetLowSpeed(low_speed); }
 void Session::SetVerifySsl(const VerifySsl& verify) { pimpl_->SetVerifySsl(verify); }
+void Session::SetProtocolVersion(const ProtocolVersion& protocolversion) { pimpl_->SetProtocolVersion(protocolversion); }
```
whoshuu: Same comments as before about using lowerCamelCase applies here too

cpr/session.cpp
whoshuu: Agreed here, let's let users set any version, not just go from `default` to `v2`

include/cpr/cprtypes.h
whoshuu: Since this library is a wrapper on curl, anything a user can set should be allowed here. That means we should include both 1.0 and 1.1, and additionally have a `default` version that sets `CURLOPT_HTTP_VERISON` to `CURL_HTTP_VERSION_NONE`.

whoshuu:
I added a `Tests Needed` label but I doubt mongoose will support HTTP/2 any time soon. I'll remove the label when I get a chance to inspect the outgoing HTTP request manually.

cpr/session.cpp
whoshuu: This is a tricky subject, but I think the right thing is to throw an exception here if the curl version doesn't support HTTP/2. Having clients send HTTP/1.x requests when they think they're sending HTTP/2 would be pretty bad

COM8:
What is the state of this PR?

COM8:
Feel free to open a new issue.

bugproof:
@COM8 what is more to be done here?

COM8:
> @COM8 what is more to be done here?

- [ ] Rebase
- [ ] Tests
- [ ] Documentation

PR #219, opened 2017-10-05
Fix -Wold-style-cast warnings

antonlindstrom:
This fixes warnings outputted by `-Wold-style-cast` which were used in a
few places.

Thank you!

### DIFF
--- a/cpr/timeout.cpp
+++ b/cpr/timeout.cpp
@@ -20,7 +20,7 @@ long Timeout::Milliseconds() const {
                                    std::to_string(ms.count()) + " ms.");
     }
 
-    return static_cast<long>(ms.count());
+    return ms.count();
 }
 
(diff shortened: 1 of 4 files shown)


antonlindstrom:
I'm not 100% sure but it looks like the failures are unrelated.

antonlindstrom:
Rebased and should be good to go :)

whoshuu:
Thanks for the PR @antonlindstrom!

pravic:
> Fix some casts that is not needed (3b76835)

Thank you, from all MSVC users :-\

antonlindstrom:
@pravic So sorry about this! I wasn't aware that it was needed for MSVC. Thank you for fixing my mistake!

pravic:
;)

PR #223, opened 2017-10-11
Turn Response's constructor arguments into rvalues to trigger move-co…

bandzaw:
Turn Response's constructor arguments into rvalues to trigger move-construction. (I just fixed this particular case and have not scanned the code looking for similar situations but maybe that could be worthwhile)


whoshuu:
Great catch @bandzaw!

Issue #166, opened 2017-01-02
Setting "Body" doesn't work for cpr::Get


mranzinger:
When I call cpr::Get with the body of the message I want (json), my server receives a request with an empty body. However, if I hack my server to support the same logic for POST as well, and use cpr::Post, it works as expected.

```
string b = "{ \"Message\": \"Hello World\" }";

// Fails because it sends an empty body
auto x = cpr::Get(cpr::Url{ "http://" + serverUrl },
                             cpr::Body{ b },
                             cpr::Header{{ "Content-Type", "application/json" }}
);

// Works, but wrong method
auto x = cpr::Post(cpr::Url{ "http://" + serverUrl },
                             cpr::Body{ b },
                             cpr::Header{{ "Content-Type", "application/json" }}
);
```

Is there a different `Body` like object I should be using with ::Get calls?

mranzinger:
It appears that setting the body of a GET is technically allowed in the HTTP spec, but doing so is dubious. I updated my server to use POST, but it does appear that there are real world useful systems that use GET with bodies, such as elastic search.

whoshuu:
This is a bug then. Being *allowed* to send a body with a GET request is distinct and separate from servers which *should not* (see [this](http://stackoverflow.com/a/983458/2297365)) respect such bodies in forming their responses. If the elastic search API parses the body for a GET request, that's their problem, but this library should at least allow you to send those requests in the first place.

PR #228, opened 2017-10-16
Fix Get Request with Body

joshleeb:
Setting "Body" doesn't work for cpr::Get #166.

Fix sending a `GET` request with a body.


whoshuu:
Thanks for the PR @joshleeb!

PR #234, opened 2017-10-30
Revert static cast for MSVC.

pravic:
Revert one commit of #219. That cast is needed for MSVC compiler.


whoshuu:
Thanks for catching this @pravic!

PR #235, opened 2017-10-30
Fix warnings of multipart implementation

pravic:
However this commit is based on the current state of master, it does not rely on #187.

I can rebase this if #187 will be merged earlier.


whoshuu:
Looks like Clang is complaining about this:

```
non-constant-expression cannot be narrowed from type 'typename iterator_traits<unsigned char *>::difference_type' (aka 'long') to 'size_t' (aka 'unsigned long') in initializer list [-Wc++11-narrowing]
```

whoshuu:
Same with the appveyor build:

```
warning C4244: 'return': conversion from '__int64' to 'long', possible loss of data
```

pravic:
I'll check it too, thanks.

PR #239, opened 2017-11-21
Add option to set Position-Independent Code

whoan:
It can be useful if you need to link `libcpr.a` to a dynamic library (as I needed to do).


COM8:
Ping @kingkili.

KingKili:
Related to #321

PR #245, opened 2018-01-12
Expose user agent in Session API

Uroc327:
This adds an API call for the Session to override the default user agent with a custom one.


mtangoo:
This would be nice to have!

PR #251, opened 2018-01-30
make the response object return more info about headers

paulohefagundes:
I've added code to make a few Header related info  to be available through the Response object

What's now available:
- Response.raw_header - contains the status_line and all headers that were received as a string
- Response.status_line - example: "HTTP/1.1 407 Proxy Authentication Required"
- Response.reason - example: "Proxy Authentication Required"

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -413,13 +413,20 @@ Response Session::Impl::makeRequest(CURL* curl) {
     }
     curl_slist_free_all(raw_cookies);
 
+    std::string status_line;
+    std::string reason;
+    Header header = cpr::util::parseHeader(header_string, &status_line, &reason);
+
     return Response{static_cast<std::int32_t>(response_code),
                     std::move(response_string),
-                    cpr::util::parseHeader(header_string),
+                    std::move(header),
                     std::move(raw_url),
                     elapsed,
                     std::move(cookies),
-                    Error(curl_error, curl_->error)};
+                    Error(curl_error, curl_->error),
+                    std::move(header_string),
+                    std::move(status_line),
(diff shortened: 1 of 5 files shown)


cpr/session.cpp
```diff
@@ -413,13 +413,13 @@ Response Session::Impl::makeRequest(CURL* curl) {
...
-                    std::move(response_string),
-                    cpr::util::parseHeader(header_string),
-                    std::move(raw_url),
-                    elapsed,
-                    std::move(cookies),
-                    Error(curl_error, curl_->error)};
+    std::string status_line;
+    Header header = cpr::util::parseHeader(header_string, &status_line);
+
+    return Response{std::move(status_line),     static_cast<std::int32_t>(response_code),
```
paulohefagundes: clang-format is what made it into 2 columns of params

test/util_tests.cpp
```diff
@@ -9,15 +9,15 @@ using namespace cpr;
...
-        "HTTP/1.1 200 OK\r\n"
-        "Server: nginx\r\n"
-        "Date: Sun, 05 Mar 2017 00:34:54 GMT\r\n"
-        "Content-Type: application/json\r\n"
-        "Content-Length: 351\r\n"
-        "Connection: keep-alive\r\n"
-        "Access-Control-Allow-Origin: *\r\n"
-        "Access-Control-Allow-Credentials: true\r\n"
-        "\r\n"};
+            "HTTP/1.1 200 OK\r\n"
```
paulohefagundes: clang-format is what indented most of this file

COM8:
Nice. Thanks!

PR #267, opened 2018-04-04
Simple try at conan package file + add install commands for cmake.

deblauwetom:
I saw the other attempt, but I wanted to show that it could be simpler, afaik about conan. So if nothing else, it can be inspirational :)

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,0 +1,57 @@
+from conans import ConanFile, tools, CMake
+import os
+
+class cprConan(ConanFile):
+    name = "cpr"
+    version = "1.3.0"
+    license = "MIT"
+    url = "https://github.com/whoshuu/cpr"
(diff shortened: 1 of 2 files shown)


tastytea:
Please use [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html). It is portable and the variables can be overridden.

jgsogo:
Hi! We are adding `cpr` to  `conan-center-index` here https://github.com/conan-io/conan-center-index/pull/450. As you can see there is a lot of people from the community contributing with new recipes for different libraries and maintaining them. We think that having this kind of central repository is the way to go for _community packages_, it makes it easier to keep recipes up to date.

jgsogo:
It's been merged, now you can install the library using

```
conan install cpr/1.3.0@
```

COM8:
Ping @kingkili.

PR #277, opened 2018-06-15
Added the ability to do "make install" command in the CMakeList.txt; …

huiguorou101:
Added the ability to do "make install" command in the CMakeList.txt by copying headers into "include" folder and copying libcpr.so into "lib" folder in the specified installation path;

Added the ability to configure generating static object(".a") or shared object(".so")


sum01:
There's at least 3 PR's doing this in various ways
* https://github.com/whoshuu/cpr/pull/176
* https://github.com/whoshuu/cpr/pull/132
* https://github.com/whoshuu/cpr/pull/267

Not saying there's anything wrong with this PR but it's probably a duplicate.

PR #282, opened 2018-07-12
Add unix socket support

edmBernard:
I need to add test but it seem to work

```
  auto r = cpr::Get(cpr::Url{"http://localhost/images/json"},
                    cpr::UnixSocket{"/var/run/docker.sock"});
  r.status_code;                  // 200
  r.header["content-type"];       // application/json; charset=utf-8
  r.text;                         // JSON text string
  std::cout << "r.status_code :" << r.status_code << std::endl;
  std::cout << "r.header :" << r.header["content-type"] << std::endl;
  std::cout << "r.text :" << r.text << std::endl;
```


edmBernard:
It's strange I was not able to reproduce issues on travis. 
```
error: ‘CURLOPT_UNIX_SOCKET_PATH’ was not declared in this scope
```

edmBernard:
It seem build test use `-DUSE_SYSTEM_CURL` which is not the curl we install with CPR repository. 
Is it possible this Curl installation don't support unix_socket ?

COM8:
Could you please fix the merge conflicts so we can test and merge it?

COM8:
Btw. What is the current state?

edmBernard:
I will check. If I remember the things that bloc me was a CI issu where the ci use the os curl instead of the one installed

tstack:
Is the `#ifdef USE_UNIX_SOCKET` necessary?  It's a little noisy.

edmBernard:
the USE_UNIX_SOCKET is defined in CURL because curl can be configure without unix socket even on linux. but if we consider our curl lib will always suport unix socket we can remove it.

tstack:
> the USE_UNIX_SOCKET is defined in CURL because curl can be configure without unix socket even on linux. but if we consider our curl lib will always suport unix socket we can remove it.

Sorry, maybe I'm not paying close enough attention, but I don't see it in a curl header.  Can you point it out to me?

Looking at the curl.h, I don't see the `CURLOPT_UNIX_SOCKET_PATH` option being guarded by an `#ifdef`:

https://github.com/curl/curl/blob/df28ed6116862921994b1abc5147cb64360e4241/include/curl/curl.h#L1768

Should the `#ifdef` in cpr be for this specific CURLOPT instead?

edmBernard:
@tstack Sorry it's a bit old, I don't remember why I have to do this `#ifdef`. As soon I have time I'll retry without it to check

edmBernard:
I'll try to test on windows with vcpkg to be sure it did break somethings

edmBernard:
It work on windows with vcpkg. Is it you that maintain the vcpkg port ?

PR #298, opened 2018-11-22
Use CURLOPT_POST and CURLUOPT_HTTPGET instead of CURLOPT_CUSTOMREQUEST

finkandreas:
CURLOPT_POST and CURLOPT_HTTPGET change the behaviour of curl, e.g. an HTTP-Return with 302 would use a subsequent GET to follow the new location. When CURLOPT_CUSTOMREQUEST is used, then all redirections are also done with this method, which can break expected behaviour.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -314,7 +314,7 @@ Response Session::Impl::Get() {
     auto curl = curl_->handle;
     if (curl) {
         curl_easy_setopt(curl, CURLOPT_NOBODY, 0L);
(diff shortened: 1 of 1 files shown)


COM8:
Requesting a body fails with this.
Could you fix this?
https://travis-ci.org/github/whoshuu/cpr/jobs/458477456#L1414-L1420

COM8:
When replacing this requests seem to fail.
This is due to a missing last package not being send.

PR #301, opened 2018-11-30
Remove clearing of system proxy

EliSchleifer:
Setting CURLTOPS_PROXY to "" will ignore the system level proxy settings. This is not curls default behavior


cpr/session.cpp
```diff
@@ -381,8 +381,6 @@ Response Session::Impl::makeRequest(CURL* curl) {
     auto protocol = url_.substr(0, url_.find(':'));
     if (proxies_.has(protocol)) {
         curl_easy_setopt(curl, CURLOPT_PROXY, proxies_[protocol].data());
-    } else {
-        curl_easy_setopt(curl, CURLOPT_PROXY, "");
```
tstack: Seems like it might be better to set this to nullptr (which I think is different from the empty string) instead of not setting it at all.  I would fear the handle would be reused after the cpr proxies had been cleared and the change not reflected.

PR #309, opened 2019-04-14
Extend VerifySsl option params to allow set host and peer verification separately

Liastre:


cpr/ssl_options.cpp
```diff
@@ -1,11 +1,18 @@
 #include "cpr/ssl_options.h"
 
-namespace cpr {
+using namespace cpr;
```
COM8, line 3: Do not use `using` here.

Liastre: Why not, it is good practice to use `using namespace` within translation unit scope. Using the whole scope is unnecessary nesting.

COM8: Yes, but it's against the coding "style" of this library and in my eyes better to read.
Also it does not spam so much functions into the auto complete if you do not use `using namespace XYZ`.

COM8:
Feel free to open a new PR.

PR #322, opened 2019-08-28
Fix bug where cmake would detect the wrong version on Windows.

Kansattica:
Windows versions of Cmake, or at least the ones that come with Visual Studio, have version numbers like this: 3.14.19060802-MSVC_2, causing this part of the makefile to try and probe for the wrong compiler flags. This fixes that.


Kansattica:
Made a different PR with just the relevant commit.

PR #325, opened 2019-09-15
Added http status codes

diogo464:
I didn't find status codes in the library so I scrapped them from https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

### DIFF
--- a/include/cpr/status_codes.h
+++ b/include/cpr/status_codes.h
@@ -84,7 +84,6 @@ constexpr bool is_redirect(const std::int32_t code) {
 constexpr bool is_client_error(const std::int32_t code) {
     return (code >= 400 && code < 500);
 }
-
 constexpr bool is_server_error(const std::int32_t code) {
     return (code >= 500 && code < 600);
 }
--- a/include/cpr/status_codes.h
+++ b/include/cpr/status_codes.h
@@ -3,69 +3,92 @@
 #include <cstdint>
 namespace cpr {
 namespace status {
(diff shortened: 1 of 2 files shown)


include/cpr/status_codes.h
```diff
@@ -0,0 +1,71 @@
+#ifndef _CPR_STATUS_CODES
+#define _CPR_STATUS_CODES
+#include <cstdint>
+namespace cpr {
+namespace status {
+constexpr std::uint32_t HTTP_100_CONTINUE = 100;
```
tstack: The current Response class uses int32_t for the status code, so this should probably match up.

include/cpr/status_codes.h
```diff
@@ -0,0 +1,71 @@
...
+constexpr std::uint32_t HTTP_101_SWITCHING_PROTOCOL = 101;
```
tstack: I'm finding it a little unusual to include the code number in the constant name, are you okay with removing it (e.g. make this HTTP_SWITCHING_PROTOCOL)?

diogo464: I had them like that because I was used to Django but I've removed them now.

tstack:
thanks!

mtangoo:
Great addition. I have found that status reason important especially to the user. There is this code[1] here under CC0. Should we add it to library so that when we get status code we can request reason to display to user for example?

[1] https://github.com/j-ulrich/http-status-codes-cpp

PR #326, opened 2019-09-24
Allows setting network interface

ukseong:
Can set CURLOPT_INTERFACE value. 

https://curl.haxx.se/libcurl/c/CURLOPT_INTERFACE.html

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -38,6 +38,7 @@ class Session::Impl {
     void SetLowSpeed(const LowSpeed& low_speed);
     void SetVerbose(const Verbose& verbose);
     void SetVerifySsl(const VerifySsl& verify);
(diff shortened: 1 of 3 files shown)


cpr/session.cpp
```diff
@@ -38,6 +38,7 @@ class Session::Impl {
     void SetLowSpeed(const LowSpeed& low_speed);
     void SetVerbose(const Verbose& verbose);
     void SetVerifySsl(const VerifySsl& verify);
+    void SetInterface(const Interface& interface);
```
tstack: "Interface" is pretty generic, how about "NetworkInterface" ?  Maybe that will fix the windows build failure as well?

ukseong: That sounds good. I will create another pull request with the change.

Issue #313, opened 2019-05-02
unsafe for multiple threads to call cpr::Get?


LumaGhost:
If I am understanding correctly: `cpr::Get` ultimately calls `curl_easy_init`. Based on [the curl documentation](https://curl.haxx.se/libcurl/c/curl_easy_init.html)

"If you did not already call curl_global_init, curl_easy_init does it automatically. This may be lethal in multi-threaded cases, since curl_global_init is not thread-safe, and it may result in resource problems because there is no corresponding cleanup."

I searched this repo for `curl_global_init` but did not find anything. does the user have to call `curl_global_init` on their own in order to be thread safe? does your library handle the global init in any way? I do not have any experience with curl so I may be missing something. 

I did notice https://github.com/whoshuu/cpr/issues/101 but I think similar errors are still possible. Since `Post` can still end up calling `curl_global_init` from more than one thread?

I am trying to understand the behavior better. any help is appreciated (:

NightNightNight:
so, how to use cpr in multiple threads? any question?

brownfeng:
> so, how to use cpr in multiple threads? any question?

我们这边直接在全局启动时候.手动初始化了curl_global_init.

ITotalJustice:
I noticed that the library doesn't call `curl_gloab_init` and instead lets `curl_easy_init` call it instead (hence the need for mutex locks). I think a better solution would be to just require the user to call `curl_global_init` at least once, or make a wrapper for the function (better) because otherwise `curl_easy_init` is thread safe. However I understand that by forcing an init, it takes away some of the simplicity of the library, but then again, I find it very common in most libs to require at least one init before use.

COM8:
The problem with the user calling `curl_global_init` would be, we would have to provide a wrapper call to prevent them having to link against libcurl.
Also, like you mentioned this removes a bit of simplicity and convenience for users.
So I'd suggest we are keeping this mutex here.

ITotalJustice:
> 
> 
> The problem with the user calling `curl_global_init` would be, we would have to provide a wrapper call to prevent them having to link against libcurl.
> Also, like you mentioned this removes a bit of simplicity and convenience for users.
> So I'd suggest we are keeping this mutex here.

[...]

PR #330, opened 2019-10-06
Add connection pool option

cleaton:
This PR adds a simple connection pool option inspired by https://curl.haxx.se/libcurl/c/threaded-shared-conn.html.

std::shared_ptr is used to enable copy constructor in order to support the `*Async` operations that will copy parameters instead of using reference.

I wrote a simple benchmark to test the difference against nginx hello world endpoint:
```C++
#include <cpr/cpr.h>

#include <cassert>
#include <chrono>
#include <future>
#include <ratio>
#include <vector>

auto requests_per_worker = 5000;
auto worker_count = 10;
auto requests = requests_per_worker * worker_count;

void run_bench(std::function<cpr::Response()> req_fun) {
    auto start = std::chrono::steady_clock::now();
    std::vector<std::future<bool>> workers;
    for (int i = 0; i < worker_count; ++i) {
        std::async(std::launch::async, [&]() {
            for (int i = 0; i < requests_per_worker; ++i) {
                auto response = req_fun();
                assert(200 == response.status_code);
            }
            return true;
        });
    }
    for (auto&& worker : workers) {
        assert(worker.get() == true);
    }
    auto end = std::chrono::steady_clock::now();
    std::chrono::duration<double, std::milli> duration = end - start;
    printf("%f requests/s\n", ((double) requests) / (duration.count() / 1000.0));
}

int main(int argc, char** argv) {
    auto url = cpr::Url{"http://localhost/hello/world"};
    cpr::ConnectionPool pool;
    printf("Without pool: ");
    run_bench([&]() { return cpr::Get(url); });
    printf("With pool: ");
    run_bench([&]() { return cpr::Get(url, pool); });
}
```
```
Without pool: 6496.959742 requests/s
With pool: 13910.127405 requests/s
```


cleaton:
As mentioned in #313, the documentation is on how to use cpr from multiple threads with regards to   `curl_global_init`. In my tests I run the non-threaded version first so `curl_global_init` will be called from the main thread via `curl_easy_init`.

COM8:
Could you create an additional PR for the `gh-pages` branche and add documentation and an/multiple example(s) please?
https://github.com/whoshuu/cpr/blob/gh-pages/advanced-usage.md

COM8:
Feel free to open a new PR.

PR #359, opened 2020-03-30
use -D BUILD_SHARED_LIBS to create .so file instead of .a

mwochal:
I thought this switch may be useful. It certainly is for me.

### DIFF
--- a/cpr/CMakeLists.txt
+++ b/cpr/CMakeLists.txt
@@ -1,6 +1,10 @@
 message(STATUS "Using CURL_INCLUDE_DIRS: ${CURL_INCLUDE_DIRS}.")
 
-add_library(${CPR_LIBRARIES}
+if($(BUILD_SHARED_LIBS))
+    set(SHARED SHARED)
+endif()
+
+add_library(${CPR_LIBRARIES} ${SHARED}
 
     # Source files
     auth.cpp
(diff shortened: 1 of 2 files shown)


COM8:
@kingkili what do you think about this?

cpr/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 message(STATUS "Using CURL_INCLUDE_DIRS: ${CURL_INCLUDE_DIRS}.")
 
-add_library(${CPR_LIBRARIES}
+if($(BUILD_SHARED_LIBS))
+    set(SHARED SHARED)
+endif()
+
+add_library(${CPR_LIBRARIES} ${SHARED}
```
KingKili, line 7: This can be removed, since CMake does use BUILD_SHARED_LIBS if no build type is specified see [docs](https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html).

deblauwetom: I also think that this is a standard feature from cmake, so not really needed

mwochal:
Hi, sorry for late response.
It was quite some time ago, and I don't remember details but I think BUILD_SHARED_LIBS didn't work.
Now I checked it again and it is just fine, so it had to be my mistake.
Anyway - great library, I like it so much.

PR #343, opened 2019-12-31
Update curl to 7.67 

MacDue:
This addresses #342 -- uploading buffers larger than about 16,000 bytes results in corrupt data.

There's no changes to cpr just simply a curl version bump.

I've not rigorously tested this change yet.

Issue #342, opened 2019-12-30
Buffers corrupted during upload for files approximately larger than 16,000 bytes.


MacDue:
Hey I'm trying to use cpr to upload an image from memory (generated with CImg). It 'works' for most of the file however all data after 0x00003EDF becomes corrupt (note sending the same image using a file on disk first works -- though is not a usable solution).

I can't figure out what's going on.
I've been toying in a debugger for ages & I've yet to locate where the bug occurs.
I've tried doing stuff like removing all frees to see if some kind of use after free is happening.

I expect this is probably on my side but I can't for the life of me find it.

Some code:

```c++
    char *buffer = NULL;
    size_t size = 0;
    FILE *fp = open_memstream(&buffer, &size);
    image.save_png(fp);
    // image.save_png("./test2.png"); -- this works
    fflush(fp);
    // data also seems correct in a debugger.
    get_client()->uploadFile(
      channel, reinterpret_cast<uint8_t*>(buffer), size, message, name); 
    fclose(fp);
    free(buffer);
```

Skipping some miss-direction that end up here where it's added to the session.
```c++
	void CPRSession::setMultipart(const std::initializer_list<Part>& parts) {
		std::vector<cpr::Part> cprParts;
		for (Part m : parts) {
			if (m.isFile) 
				cprParts.push_back(cpr::Part(m.name, cpr::File(m.value)));
			else if (m.buffer_len > 0) {
				auto buffer = cpr::Buffer(m.buffer, m.buffer + m.buffer_len, m.name);
				cprParts.push_back(cpr::Part(m.name, buffer));
			}
			else
				cprParts.push_back(cpr::Part(m.name, m.value));
		}

		muiltpart.parts = cprParts;
		session.SetMultipart(muiltpart); // add to cpr session
	}
```
that will after a while call `` session.Post  ()``.
At all stages in the code a above the data seems fine. Note the cpr calls are going through a library first hence the abridged code.

I've attached two sample images (correct.png & courrupt.png).
![Corrupt](https://user-images.githubusercontent.com/11597044/71605257-7602bd80-2b5f-11ea-9dca-f8b676f1a27e.png)
![Correct](https://user-images.githubusercontent.com/11597044/71605270-8024bc00-2b5f-11ea-9c5f-daff075aab16.png)

MacDue:
Also to note the exact code above works when posting smaller images from memory. 
[...]

PR #364, opened 2020-04-16
[WIP] Updated curl from 7.56.0 to 7.69.1

COM8:
Updated curl from `7.56.0` to `7.69.1`.

### Related Issues
* #342 
* #387 

### Related PRs
* #343

### Changes
* Updated curl from `7.56.0` to `7.69.1`
* Switched from a self maintained fork of `mongoose` to the official `mongoose` repo
* Switched from a self maintained fork of `googletest` to the official `googletest` repo
* Updated all unit tests to the latest version of the `mongoose` server
* Refactored the `mongoose` http web server

### TODO
- [x] ~~Fix Windows linker errors~~
- [x] ~~Fix MacOS not executing tests~~
- [x] Fix the segfault caused by `ssl-test`


tstack:
Hmm, what's going on with the appveyor build here?  Why would the mongoose build be affected?

COM8:
I don't know for sure.
But I guess it's caused by compiling under Windows with an rather old compiler (VS 2015).
I'm currently investigating it.

COM8:
Validated everything again with VS2019 and it looks like it's working like it should.
Only one problem: the tests fail, since it somehow does not build `gtest`.
**But** `mongoose` and `cpr` are building just fine.

Perhaps we should switch to an more up to date compiler (for Windows and Linux) and build system (perhaps GitHub actions).

COM8:
@tstack 
Do not include this PR if you want to create a new tag.
I can confirm everything works like it should on older VS compilers but on newer ones other stuff is breaking.
I had to update `googletest` since the [fork](https://github.com/whoshuu/googletest), we are using is so outdated it is not compatible with VS <= 2017 due to deprecated macros.

I also updated `mongoose` to a release **1080** commits more into the future.
**But** this is breaking the test server since `mongoose` changed so much.
In theory it complies, but the tests do not.

I'm updating the server and then we can create a new tag again when this PR is ready to be merged.

Any concerns?

tomay3000:
I am just curious here,
Starting with C++11 std::string::data() is guaranteed to have a null ('\0') terminating character which means it is the same as std::string::c_str().
If this is just for consistency and readability then this should be OK.

COM8:
@tomay3000 do you have any references for this?
Based on the official documentation over at [cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string/data) it is not guaranteed.
Yes, there are implementations of `std::string` that handle it like that, but I can not find any general reference for it.

> The pointer is such that the range [data(); data() + size()) is valid and the values in it correspond to the values stored in the string. 

Source: [cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string/data)

tomay3000:
> @tomay3000 do you have any references for this?

Based on `std::string::data()` official documentation over at [cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string/data):

Description | C++11
-- | --
The returned array **is not required** to be **null-terminated**.If empty() returns true, the pointer is a non-null pointer that should not be dereferenced. | **UNTIL C++11**
The returned array **is null-terminated**, that is, `data()` and `c_str()` perform the same function.If empty() returns true, the pointer points to a single null character. | **SINCE C++11**
---
And on `std::string::c_str()` official documentation over at [cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string/c_str):

Description | C++11
-- | --
`c_str()` and `data()` perform **THE SAME** function. | **SINCE C++11**
---
Most of C++11 code uses both functions only for consistency, though they are the same.

Meaning:
- `std::string::data()` is meant to be used when retrieving data to be used as data (as binary)
- `std::string::c_str()` is meant to be used when retrieving data to be used as C-String.

It is the C++11 stand that messed it up when it was ok prior to C++11.

TIA.

COM8:
You are right.
So, this change was not necessary.
I still think it is clearer to read in my eyes.

tomay3000:
> You are right.
> So, this change was not necessary.
> I still think it is clearer to read in my eyes.

Ok, and BTW **curl v7.71.0**, has been released today.

PR #366, opened 2020-04-16
Have libcurl request all compression encodings it can handle.

Kansattica:
Following the instructions [here](https://curl.haxx.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html), this makes libcurl use all the flavors of compression it was built with, usually including zlib at the least.


COM8:
Nice work.
Thanks!

PR #374, opened 2020-05-08
build: add cmake config to use find_package

gocarlos:
# what
* add cmake config
* install config file
* add conan to readme
# why
* allow usage of cmake's `find_package`


COM8:
@kingkili Ping.

KingKili:
Overall this looks good, but I am running into problems when using the submodule curl. Can please have look into it.

gocarlos:
> Overall this looks good, but I am running into problems when using the submodule curl. Can please have look into it.

which problems do you have?
usging `find_package` is only intended when you choose to use the system CURL library
using the vendored CURL library should probably only done for development, as linking two different CURL libs results in undefined behaviour

KingKili:
Yes but it should be still possible to build the library without system curl, since it fails while configuring it else.
Steps to reproduce:
```
git clone https://github.com/gocarlos/cpr --recursive
mkdir cpr/build
cd cpr/build
cmake .. -DUSE_SYSTEM_CURL=OFF -DUSE_SYSTEM_GTEST=OFF
´´´
Which will result in the [log](https://github.com/whoshuu/cpr/files/4626856/log.txt)

File: CMakeLists.txt [hunk 1]
```diff
@@ -92,4 +92,31 @@ if(BUILD_CPR_TESTS)
     add_subdirectory(test)
 endif()
 
-install(DIRECTORY include/cpr DESTINATION include)
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+set(CPR_CONFIG_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
+
+install(
+  TARGETS ${PROJECT_NAME}
```
KingKili, line 100: Please do not rely on ${PROJECT_NAME} because CMake 2.8 😒
please use ${CPR_LIBRARIES} instead

COM8:
I’m closing this.

PR #375, opened 2020-05-12
Remove self/duplicate includes from session.h

kiroma:


COM8:
The failing Jenkins should not be a problem. 
It's failing because `apt` failed.

PR #380, opened 2020-05-19
Add move constructor and assignment to Session

a-pushkin:
Add move constructor and assignment to Session to enable returning Session object from functions like this:

```cpp
cpr::Session session_factory()
{
  cpr::Session session{};
  session.SetUrl(url_);
  if (auth_.has_value())
  {
    session.SetAuth(*auth_);
  }
  return session;
}
```

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -549,6 +549,8 @@ Response Session::Impl::makeRequest(CURL* curl) {
 
 // clang-format off
 Session::Session() : pimpl_{ new Impl{} } {}
+Session::Session(Session&& other) : pimpl_{ std::move(other.pimpl_) } {}
+Session& Session::operator=(Session&& other) { pimpl_ = std::move(other.pimpl_); return *this; }
 Session::~Session() {}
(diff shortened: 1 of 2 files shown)


COM8:
Thanks for your PR.

a-pushkin:
> What about all the other members of `Session`?
> https://github.com/whoshuu/cpr/blob/571fbe17c9201571f3930d982f73a6458d2b88cb/cpr/session.cpp#L62-L65

These are members of `Session::Impl`. The only member `Session` has is:
https://github.com/whoshuu/cpr/blob/571fbe17c9201571f3930d982f73a6458d2b88cb/include/cpr/session.h#L98

COM8:
> > What about all the other members of `Session`?
> > https://github.com/whoshuu/cpr/blob/571fbe17c9201571f3930d982f73a6458d2b88cb/cpr/session.cpp#L62-L65
> 
> These are members of `Session::Impl`. The only member `Session` has is:
> 
> https://github.com/whoshuu/cpr/blob/571fbe17c9201571f3930d982f73a6458d2b88cb/include/cpr/session.h#L98

Ah, right.
Then, good job.

PR #386, opened 2020-05-22
export upload size and download size in response

chrisliu2345:
export upload size and download size in response


COM8:
Nice idea.
Thanks.

PR #389, opened 2020-06-03
Update VERSION

KingKili:
Push version file to current version.
I think this should be also back ported to the last release.


COM8:
It's not "possible" to add a commit to an existing tag. This would invalidate the rule of never change an existing git tag.

KingKili:
Still it would make sense to change this file since it will create wrongly named versions of cpr

PR #390, opened 2020-06-04
Fix build after 7e4a803

luk1337:


COM8:
Thanks!
What compiler are you using?
Since it looks like the CI was fine with that broken transitive include.

luk1337:
MSVC (win) and clang-10 (linux) were affected here.

PR #392, opened 2020-06-04
Fix tests build when CMAKE_USE_OPENSSL=OFF

luk1337:
* We are defining MG_ENABLE_SSL only when CMAKE_USE_OPENSSL
  is enabled thus [ssl_cert, ssl_key] fields would be missing
  otherwise.


COM8:
I'm aware of this issue and it will be fixed by #383 .
In #383 we go so far to not even build the `HttpsServer` and the associated tests if no OpenSSl was found or has been disabled.

COM8:
But thanks for the fix!
I don't think I will merge this PR since #383 is almost done.

Issue #395, opened 2020-06-09
Build broke in aa6132 when using system CURL


audaki:
Heya cpr maintainers,

our build broke in commit https://github.com/whoshuu/cpr/commit/aa6132278e623739d3c0520564edfc77b35e7bbd

Since that commit CMake is trying to use CURL::libcurl but without doing a FindCURL beforehand. Our build broke in openSUSE Tumbleweed and Leap. It should be fixable by using https://cmake.org/cmake/help/latest/module/FindCURL.html beforehand.

kind regards, Kira

COM8:
Ping @kingkili.

KingKili:
You are right, but I think it would be smarter to just do the dependency logic in the root `CMakeLists.txt`. But I will prepare a quick fix.

PR #396, opened 2020-06-09
find_package(CURL) where it is needed

audaki:
Fix issue #395


COM8:
Fixed by #397 .
Thanks for your help!

PR #397, opened 2020-06-09
Move Find_Package operations to root, Fixes #395

### LINKED ISSUE #395: Build broke in aa6132 when using system CURL

KingKili:
A alternative fix for #395


CMakeLists.txt
```diff
@@ -39,8 +39,100 @@ cpr_option(CMAKE_USE_OPENSSL "Use OpenSSL code. Experimental" ON)
 message(STATUS "=======================================================")
 
 include(GNUInstallDirs)
+include(FetchContent)
+include(cmake/code_coverage.cmake)
+
+# Curl configuration
+if(USE_SYSTEM_CURL)
+    find_package(CURL)
```
COM8, line 47: Thanks for the PR.
Could you please add some output, once `FindCURL` has finished depending on what the result of `CURL_FOUND` is.

Something like:
`True` -> System CURL found
`ELSE` -> System CURL not found and then fail

KingKili: FindPackage prints such things by default, if curl is not found cpr will try to use fetch content

PR #401, opened 2020-06-14
Fix docs for #383

KingKili:


index.md
```diff
@@ -61,30 +61,23 @@ Support for the following will be forthcoming (in rough order of implementation
...
-
-```
-include_directories(${CPR_INCLUDE_DIRS})
-target_link_libraries(your_target_name ${CPR_LIBRARIES})
+```cmake
+target_link_libraries(your_target_name PRIVATE cpr::cpr)
 ```
 
-and that should do it! Using the submodule method of integrating C++ Requests, there's no need to handle libcurl yourself, all of those dependencies are taken care of for you.
+That should do it!  Using the `fetch_content` method of integrating C++ Requests, there's no need to handle `libcurl` yourself, all of those dependencies are taken care of for you.
```
COM8, line 80: Double space between "it!" and "Using"

PR #402, opened 2020-06-19
Pass correct data type when setting CURLOPT_POSTFIELDSIZE_LARGE

luk1337:
* Somehow when compiling with MSVC CURLOPT_POSTFIELDSIZE_LARGE ends up
  being set to some really large value, casting length to curl_off_t
  fixes that.
* Example nginx log message: `100 client intended to send too large body: 200127437130956800 bytes`.


COM8:
You are right.
Thanks for the PR.

For later reference:
https://curl.haxx.se/libcurl/c/CURLOPT_POSTFIELDSIZE_LARGE.html

PR #403, opened 2020-06-19
Add ability to add proxy out of initialization

palikar:
In a project of mine, I want to construct the proxies option based on external input. This is currently not possible as `cpr::Proxies` has to be completely constructed in the initialization and there is no way to add new proxy for a certain protocol for the object.

With my slight change in the API, something like the following will be possible
``` c++
cpr::Proxies proxies{};
proxies["http"]="url_to_proxy"
```

I would love to be able to use that.

### DIFF
--- a/include/cpr/proxies.h
+++ b/include/cpr/proxies.h
@@ -13,6 +13,7 @@ class Proxies {
     Proxies() {}
     Proxies(const std::initializer_list<std::pair<const std::string, std::string>>& hosts);
 
+    void add(const std::string& protocol, std::string proxy_url);
     bool has(const std::string& protocol) const;
     const std::string& operator[](const std::string& protocol);
 
--- a/cpr/proxies.cpp
+++ b/cpr/proxies.cpp
@@ -10,6 +10,10 @@ namespace cpr {
 Proxies::Proxies(const std::initializer_list<std::pair<const std::string, std::string>>& hosts)
         : hosts_{hosts} {}
 
+void add(const std::string& protocol, std::string proxy_url) {
+    hosts_[protocol] = std::move(proxy_url);
+}
+
 bool Proxies::has(const std::string& protocol) const {
     return hosts_.count(protocol) > 0;
 }


COM8:
Thanks for contributing to CPR.
I will have a look at this PR in the next couple of days, since I think there was a CURL related reason, why we had those set to const.

palikar:
Oh, ok, I'll look into it myself. If you need the const, it's probably because the crul API need some sort of `const char *` and the lack of const may be a problem.

palikar:
Yes, there is line in `session.cpp`:
```c++
...
curl_easy_setopt(curl, CURLOPT_PROXY, proxies_[protocol].c_str());
...
```
`c_str()` is a const function so it cannot be called on a non-const object. This makes my original change unviable. Now I've just added and extra method to the Proxy class for adding a new proxy.

include/cpr/proxies.h
```diff
@@ -13,6 +13,7 @@ class Proxies {
     Proxies() {}
     Proxies(const std::initializer_list<std::pair<const std::string, std::string>>& hosts);
 
+    void add(const std::string& protocol, std::string proxy_url);
```
COM8, line 16: `std::string proxy_url` should be `std::string&& proxy_url`, so the user knows about the copy of `proxy_url`.

COM8:
Any updates for this PR?

include/cpr/proxies.h
xloem: > Yes, there is line in `session.cpp`:
> 
> ```c++
> ...
> curl_easy_setopt(curl, CURLOPT_PROXY, proxies_[protocol].c_str());
> ...
> ```
> 
> `c_str()` is a const function so it cannot be called on a non-const object. This makes my original change unviable. Now I've just added and extra method to the Proxy class for adding a new proxy.

It's totally fine to call a const function on a non-const object.  It looks fine to me to remove the constness and make it modifiable like Cookies is.  Proxies is read once by Session at the start of every request, and the Session object has its own storage, so user modification to the object after it is passed would have no effect anyway.

xloem:
> Yes, there is line in `session.cpp`:
> 
> ```c++
> ...
> curl_easy_setopt(curl, CURLOPT_PROXY, proxies_[protocol].c_str());
> ...
> ```
> 
> `c_str()` is a const function so it cannot be called on a non-const object. This makes my original change unviable. Now I've just added and extra method to the Proxy class for adding a new proxy.

It's totally fine to call a const function on a non-const object.  It looks fine to me to remove the constness and make it modifiable like Cookies is.  Proxies is read once by Session at the start of every request, and the Session object has its own storage, so user modification to the object after it is passed would have no effect anyway.  (sorry prvious reply became a line comment for some reason, probably hit wrong button somewhere)

include/cpr/proxies.h
xloem: i was looking for something similar to this pattern to understand it better and found https://github.com/whoshuu/cpr/blob/master/include/cpr/curl_container.h#L14 which is used by https://github.com/whoshuu/cpr/blob/master/include/cpr/curl_container.h#L42 for basically the same thing in the Parameters object.  `add` is capitalized to `Add` and both `const &` and `&&` forms are provided.

COM8:
Closing for inactivity.

PR #404, opened 2020-06-23
CMake add ability to use WINSSL

KingKili:


COM8:
When setting `USE_OPENSSL=OFF` you have to set also `CMAKE_USE_OPENSSL=OFF` to tell CURL to use alternative SSL implementations like winSSL.

CMakeLists.txt
```diff
@@ -52,10 +62,21 @@ if(NOT USE_SYSTEM_CURL OR NOT CURL_FOUND)
     set(BUILD_CURL_EXE OFF CACHE INTERNAL "" FORCE)
     set(BUILD_TESTING OFF CACHE INTERNAL "" FORCE)
     set(HTTP_ONLY ON CACHE INTERNAL "" FORCE)
-    if(NOT CMAKE_USE_OPENSSL)
+
+    if (USE_WINSSL OR USE_OPENSSL)
+        set(SSL_ENABLED ON CACHE INTERNAL "" FORCE)
+    else()
+        set(CURL_CA_PATH "none" CACHE INTERNAL "" FORCE)
```
COM8, line 69: This is not required. Since CURL should use it's default paths for the CA files.

CMakeLists.txt
```diff
@@ -52,10 +62,21 @@ if(NOT USE_SYSTEM_CURL OR NOT CURL_FOUND)
...
+    endif()
+
+    if(USE_WINSSL)
+        set(CMAKE_USE_WINSSL ON CACHE INTERNAL "" FORCE)
         set(CURL_CA_PATH "none" CACHE INTERNAL "" FORCE)
     endif()
-    set(SSL_ENABLED ${CMAKE_USE_OPENSSL} CACHE INTERNAL "" FORCE)
+
+    if(USE_OPENSSL) 
+        set(CMAKE_USE_OPENSSL ON CACHE INTERNAL "" FORCE)
```
COM8, line 78: Otherwise set to `OFF`.

COM8:
https://github.com/whoshuu/cpr/blob/c8d33915dbd88ad6c92b258869b03aba06587ff9/test/CMakeLists.txt#L54-L61

Shouldn't this be changed to:
```CMake
 if (USE_OPENSSL) 
```

or

```CMake
 if (SSL_ENABLED) 
```

CMakeLists.txt
KingKili: This is required, cause in that case ssl is deactivated, and curl will complain while building

PR #408, opened 2020-06-26
Fix broken api

dilshodm:
Before it was possible to use Payload:
1. with string variables, like:
```c++
std::string value = "hello";
Payload payload{{"key", value}};
```
With broken API it accepted only:
```c++
std::string value = "hello";
Payload payload{{"key", value.c_str()}};
// or
Payload payload2{{"key", string(value)}};
```

2. Before it was able to use AddPair with only pair argument:
```c++
Payload payload{{"key", "value"}};
payload.AddPair({"key2", "value2"});
```

3. Allow to use arbitrary size signed/unsigned integers (uint64_t, etc)

### DIFF
--- a/include/cpr/parameters.h
+++ b/include/cpr/parameters.h
@@ -22,7 +22,7 @@ class Parameters {
     Parameters() = default;
     Parameters(const std::initializer_list<Parameter>& parameters);
 
-    void AddParameter(const Parameter& parameter, const CurlHolder& holder);
+    void AddParameter(const Parameter& parameter, const CurlHolder& holder = CurlHolder());
 
     std::string content;
 };
--- a/include/cpr/parameters.h
+++ b/include/cpr/parameters.h
@@ -22,7 +22,7 @@ class Parameters {
     Parameters() = default;
     Parameters(const std::initializer_list<Parameter>& parameters);
 
-    void AddParameter(const Parameter& parameter, const CurlHolder& holder = CurlHolder());
+    void AddParameter(const Parameter& parameter, const CurlHolder& holder);
 
     std::string content;
 };
--- a/include/cpr/payload.h
+++ b/include/cpr/payload.h
@@ -16,7 +16,8 @@ struct Pair {
             : key(p_key), value(p_value) {}
     Pair(std::string&& p_key, std::string&& p_value)
             : key(std::move(p_key)), value(std::move(p_value)) {}
-    Pair(std::string&& p_key, const std::int32_t& p_value)
+    template<typename ValueType, typename std::enable_if<std::is_integral<ValueType>::value, bool>::type = true>
+    Pair(std::string&& p_key, ValueType p_value)
             : key(std::move(p_key)), value{std::to_string(p_value)} {}
 
     std::string key;
--- a/include/cpr/payload.h
+++ b/include/cpr/payload.h
@@ -35,7 +35,7 @@ class Payload {
     }
     Payload(const std::initializer_list<Pair>& pairs);
 
-    void AddPair(const Pair& pair, const CurlHolder& holder = CurlHolder());
+    void AddPair(const Pair& pair, const CurlHolder& holder);
 
     std::string content;
 };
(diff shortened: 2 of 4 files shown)


include/cpr/parameters.h
```diff
@@ -22,7 +22,7 @@ class Parameters {
     Parameters() = default;
     Parameters(const std::initializer_list<Parameter>& parameters);
 
-    void AddParameter(const Parameter& parameter, const CurlHolder& holder);
+    void AddParameter(const Parameter& parameter, const CurlHolder& holder = CurlHolder());
```
COM8, line 25: I'm not a fan of a default argument here for `holder` since creating a new `CurlHolder()` object can be rather expensive.
We should force users to reuse theirs.

include/cpr/payload.h
```diff
@@ -33,7 +36,7 @@ class Payload {
     }
     Payload(const std::initializer_list<Pair>& pairs);
 
-    void AddPair(const Pair& pair, const CurlHolder& holder);
+    void AddPair(const Pair& pair, const CurlHolder& holder = CurlHolder());
```
COM8, line 39: Same like above with the default value for `holder`.

include/cpr/payload.h
```diff
@@ -12,9 +12,12 @@
 namespace cpr {
 
 struct Pair {
+    Pair(const std::string& p_key, const std::string& p_value)
+            : key(p_key), value(p_value) {}
     Pair(std::string&& p_key, std::string&& p_value)
             : key(std::move(p_key)), value(std::move(p_value)) {}
-    Pair(std::string&& p_key, const std::int32_t& p_value)
+    template<typename ValueType, typename std::enable_if<std::is_integral<ValueType>::value, bool>::type = true>
```
COM8, line 19: Might just me, but why do we need to template this function?
I prefer it, if users have to convert their values to string by them selves to prevent unexpected behaviour.

dilshodm:
@COM8  It seems like (unrelated to this PR) cpr_callback_tests failed due to timeout. Looks like race condition test, where timeout 50ms maybe not always enough. Should I do some insignificant commit to restart CI tests again?

dilshodm:
Regarding unwanted surprises using int32 as a argument type:
```c++
    uint32_t aaa = 3'000'000'000;
    cpr::Payload payload {
        {"key", aaa},
    };
```
is compiling with narrowing warning, and if user skips that warning, then the output will be
```
key=-1294967296 
```

COM8:
> @COM8 It seems like (unrelated to this PR) cpr_callback_tests failed due to timeout. Looks like race condition test, where timeout 50ms maybe not always enough. Should I do some insignificant commit to restart CI tests again?

I don't get it why the CI is suddenly failing. This is not related to you. I'm working on a fix for it. It's the same for master.

COM8:
> Regarding unwanted surprises using int32 as a argument type:
> 
> ```c++
>     uint32_t aaa = 3'000'000'000;
>     cpr::Payload payload {
>         {"key", aaa},
>     };
> ```
> 
> is compiling with narrowing warning, and if user skips that warning, then the output will be
> 
> ```
> key=-1294967296 
> ```

Is this the current behaviour or with templates?

dilshodm:
> > Regarding unwanted surprises using int32 as a argument type:
> > ```c++
> >     uint32_t aaa = 3'000'000'000;
> >     cpr::Payload payload {
> >         {"key", aaa},
> >     };
> > ```
> > 
> > 
> > is compiling with narrowing warning, and if user skips that warning, then the output will be
> > ```
> > key=-1294967296 
> > ```
> 
> Is this the current behaviour or with templates?

This is the current behaviour, because it's used int32_t as a type. If we are using templates it will automatically guess the correct type. However, in current implementation and with template if we use `{ "char", 'a'}` we will get `char=97`

COM8:
> > Is this the current behaviour or with templates?
> 
> This is the current behaviour, because it's used int32_t as a type. If we are using templates it will automatically guess the correct type. However, in current implementation and with template if we use `{ "char", 'a'}` we will get `char=97`

I removed the constructor for `cpr::Pair` since it is inconsistent.
Probably only a remnant of an earlier test version.

dilshodm:
Closing this PR and will open new one with clean commit

PR #413, opened 2020-07-03
Fix for Parameters using string variable

dilshodm:
Similar to Payload fix this PR fixes to use Parameters with string variables:
```c++
std::string key = "hello";
std::string val = "world";
Parameters parameters{ {key, val} };
```


COM8:
Nice. Thanks 👍🏻.

PR #415, opened 2020-07-04
Provide hashes for git content to replace the security of submodules

xloem:
I noticed the project is switching to using FetchContent instead of git submodules, but isn't using commit hashes to identify content the way submodules do.

This opens the system to a number of additional security concerns where an attacker could alter a dependency or the code of cpr itself for a developer using it, by using a signed certificate and dns spoofing, or compromising the github account of a developer and changing a tag, or working with github's servers themselves to offer differing data, all things that nations could do to gain advantage over each other at the expense of the citizens.  These are concerns that were severely mitigated with submodules, and could be comparably mitigated by using commit hashes instead of tags in FetchContsnt.

This PR adds a git commit hash to uses of FetchContent so that the content fetched can be verified correct.


COM8:
Looks good. Thanks.

PR #425, opened 2020-07-23
Simplify Authentication classes and make them copyable/movable

jagerman:
I ran into an issue where I want to be able to assign a member Authentication variable from a provided Authentication value, but that is impossible because of the `const` private members.  Removing the `const` fixes this (and since they are already private doesn't change anything), but while I was here I also noticed some unnecessary related code, so simplified that as well.  See individual commit messages for details.

### DIFF
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -18,7 +18,7 @@ class Authentication {
     }
     virtual ~Authentication() = default;
 
-    const char* GetAuthString() const noexcept;
+    virtual const char* GetAuthString() const noexcept;
 
   protected:
     std::string auth_string_;
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -15,7 +15,7 @@ class Authentication {
 
     const char* GetAuthString() const noexcept;
 
-  private:
+  protected:
     std::string auth_string_;
 };
 
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -9,17 +9,13 @@ namespace cpr {
 
 class Authentication {
   public:
-    Authentication(std::string&& username, std::string&& password)
-            : username_(std::move(username)),
-              password_(std::move(password)), auth_string_{this->username_ + ":" +
-                                                           this->password_} {}
+    Authentication(const std::string& username, const std::string& password)
+            : auth_string_{username + ":" + password} {}
     virtual ~Authentication() = default;
 
     const char* GetAuthString() const noexcept;
 
   private:
-    const std::string username_;
-    const std::string password_;
     const std::string auth_string_;
 };
 
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -12,10 +12,7 @@ class Authentication {
     Authentication(const std::string& username, const std::string& password)
             : auth_string_{username + ":" + password} {}
     Authentication(std::string&& username, std::string&& password)
-            : auth_string_{std::move(username)} {
-        auth_string_ += ':';
-        auth_string_ += std::move(password);
-    }
+            : auth_string_{std::move(username) + ":" + std::move(password)} {}
     virtual ~Authentication() = default;
 
     virtual const char* GetAuthString() const noexcept;
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -11,6 +11,11 @@ class Authentication {
   public:
     Authentication(const std::string& username, const std::string& password)
             : auth_string_{username + ":" + password} {}
+    Authentication(std::string&& username, std::string&& password)
+            : auth_string_{std::move(username)} {
+        auth_string_ += ':';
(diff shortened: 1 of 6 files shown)


include/cpr/digest.h
```diff
@@ -7,12 +7,7 @@ namespace cpr {
 
 class Digest : public Authentication {
   public:
-    Digest(std::string&& username, std::string&& password)
-            : Authentication{std::move(username), std::move(password)} {}
-
-    ~Digest() override = default;
-
-    const char* GetAuthString() const noexcept;
+    using Authentication::Authentication;
```
COM8, line 10: I'm not a fan of this notation.

Could you please replace it with the appropriate overrides for constructors.

include/cpr/auth.h
```diff
@@ -9,18 +9,14 @@ namespace cpr {
...
-            : username_(std::move(username)),
-              password_(std::move(password)), auth_string_{this->username_ + ":" +
-                                                           this->password_} {}
+    Authentication(const std::string& username, const std::string& password)
+            : auth_string_{username + ":" + password} {}
     virtual ~Authentication() = default;
 
     const char* GetAuthString() const noexcept;
 
   private:
```
COM8, line 18: Could you change the visibility to protected for properties.

include/cpr/ntlm.h
```diff
@@ -8,11 +8,7 @@ namespace cpr {
 
 class NTLM : public Authentication {
   public:
-    template <typename UserType, typename PassType>
-    NTLM(UserType&& username, PassType&& password)
-            : Authentication{std::move(username), std::move(password)} {}
-
-    const char* GetAuthString() const noexcept;
+    using Authentication::Authentication;
```
COM8, line 11: Same problem like above - not a fan of this.

include/cpr/auth.h
```diff
@@ -9,18 +9,14 @@ namespace cpr {
 
 class Authentication {
   public:
-    Authentication(std::string&& username, std::string&& password)
```
COM8, line 12: Please also add rvalue constructors for `Authentication` to keep compatibility with older implementations.

include/cpr/digest.h
jagerman: I don't understand "not a fan" -- this is a bog standard C++11 inherited constructor.

include/cpr/auth.h
jagerman: Rvalues can be passed to const lvalue reference arguments, so this does not break compatibility.

include/cpr/auth.h
jagerman: Changed.

include/cpr/digest.h
jagerman: I put explicit constructors back in.

jagerman:
@COM8 - one other possible change to be added here (checking first before I change it) is to make the `const char* GetAuthString() const noexcept;` virtual.  Without that, as it stands currently, there's no way for me to create a subclass that overrides it (because once it is cast to the base class the `auth.GetAuthString()` is not going to do a virtual method lookup).

include/cpr/auth.h
COM8: You are right.
Nevertheless, I'd like to keep the `rvalue` constructors for verbosity reasons.

COM8:
> @COM8 - one other possible change to be added here (checking first before I change it) is to make the `GetAuthString()` method virtual. Without that, as it stands currently, there's no way for me to create a subclass that overrides it (because once it is cast to the base class and given to a `cpr` method the `auth.GetAuthString()` is not going to do a virtual method lookup).

@jagerman Thats a good idea. I'd say go for it :+1:

include/cpr/digest.h
COM8: Without the explicit constructors I have to open two files to find out, which constructors are provided.
So this is more work for me as a developer using `CPR`.

include/cpr/auth.h
jagerman: Done.

include/cpr/auth.h
```diff
@@ -9,18 +9,19 @@ namespace cpr {
 
 class Authentication {
   public:
+    Authentication(const std::string& username, const std::string& password)
+            : auth_string_{username + ":" + password} {}
     Authentication(std::string&& username, std::string&& password)
-            : username_(std::move(username)),
-              password_(std::move(password)), auth_string_{this->username_ + ":" +
-                                                           this->password_} {}
+            : auth_string_{std::move(username)} {
```
COM8, line 15: Any reason against:
```C++
Authentication(std::string&& username, std::string&& password)
    : auth_string_(std::move(username) + ":" + std::move(password)) {
}
```

jagerman: No reason; updated.

COM8:
Great work.Thanks. 👍🏻

PR #426, opened 2020-07-26
Add missing Session methods

jagerman:
SetUnixSocket, SetSslOptions, and SetVerbose are declared in the Session
header but missing implementations (thus causing linking failures if
attempting to use any of them) -- only the SetOption(...) overloads had
implementations.  This adds the missing implementations.


COM8:
Nice. Thanks :+1: .

PR #428, opened 2020-07-27
fix: Don't search for invalid cURL components

mikedld:
`PROTOCOLS` and `FEATURES` aren't real cURL package components. AFAICT neither CMake-bundled FindCURL module nor cURL-provided config package (when building it with CMake) support those components, and both fail to find the library when they are passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,11 +55,11 @@ include(cmake/code_coverage.cmake)
 
 # Curl configuration
 if(USE_SYSTEM_CURL)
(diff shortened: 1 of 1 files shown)


mikedld:
Note that this change may still not result in proper cURL search since there's another defect in CMake find module which was there since the introduction of cURL components support: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/2677

KingKili:
Looks good to me. Thanks. @COM8 can you please merge

Issue #432, opened 2020-08-08
Add an option or callback to cancel a request


ITotalJustice:
Hi, thanks for this amazing library, it's been frustration free so far 😄 

In the app I'm making, i have an option for the user to search for shows. For this, I'm using the `GetCallback`. However I want the option for the user to search again _before_ the first async request has completed, or simple stop searching and do something else. I tried finding an option for this in this library, though i couldn't find anything.

libcurl has a decent option for this if `CURLOPT_XFERINFOFUNCTION` is set. Its a callback function for tracking the progress which can also be used for cancelling a request is you return a non zero value. [More info about it here.](https://curl.haxx.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html)

Maybe having a callback function as a third optional parameter to would be good idea? 

EDIT: this would also be useful to allow us to create a progress bar 😄

Issue #111, opened 2016-07-17
Feature Request: Progress callback


danielkop:
Normally when integrating libcurl, I could use the WRITE_FUNCTION to keep a download progress.

```
size_t write_callback(void *buffer, size_t size, size_t nitems, CurlState* state) {
        auto *cBuffer = static_cast<char *>(buffer);
        state->data.append(cBuffer, cBuffer + size * nitems);
        if(!state->callback(state->data)) //Allow callback to cancel the download
              return -1;
        return size * nitems;
}

```

I think this could be a great feature to add to CPR as well.

nabijaczleweli:
Wrap the code in code blocks :v

tolstenko:
+1

COM8:
For later reference: https://curl.haxx.se/libcurl/c/CURLOPT_PROGRESSFUNCTION.html

Issue #271, opened 2018-05-18
Streaming Download


RPGillespie6:
I read the documentation, but it's not apparent to me if it's possible to do streaming GET downloads.

The problem is that the download size is greater than RAM size, so how can I tell cpr to stream to an ofstream object for example that writes to flash?

And in the same vein, I would also like to be able to resume a partial download (i.e. [curl -C](https://www.cyberciti.biz/faq/curl-command-resume-broken-download/))

Issue #255, opened 2018-02-13
Request feature - Callbacks support


WHK102:
Howto download a large file? cpr store buffer in a variable to return content, please, add a callback as argument (like as std::function) for each group of bytes, by example, each 2048 bytes or each line.

PR #433, opened 2020-08-08
progress callback draft

ITotalJustice:
Hi! I added an issue here #432 a few minutes ago, but then i realised that there was an already existing issue that's been open for quite a while sadly  #111.

This is a quick draft / example of adding a progress callback option, and it works 😄 . This allows us to create progress bars and also cancel any request early by returning a non zero value from the callback!

If there's any chance of adding this and you want me to clean up anything, let me know! Thanks


ITotalJustice:
Made some changes to the initial pr. I simple class that the user passes a function callback and a data that they want to be passed into callback (this was missing from my initial pr). I think this looks much better that the initial pr.

ITotalJustice:
I was thinking, would it be a good idea to create 2 more classes for write and read callbacks? I could rename `progresscb.h` to `callback.h` and then have the 3 classes there. 

Iirc there's a few more libcurl callbacks also, such as headers. Maybe it might be a good option for me to add those also?

ITotalJustice:
I noticed that a lot of the time, people call `curl_easy_getinfo` in the progress callback. A way to still support this would be give curl a callback created cpr, which when called, calls the user callback _also_ passing in a const *curl.

Or I think a better way would to pass a const struct / class to the user callback. This would simplify the callback that the user has to create as it will only have one pram in the function, then access only what they want from the struct / class.

One immediate problem is that let's say the user returns `!= 0` from the user callback (meaning that they want to exit), this value would actually be returned back to the cpr function. I mean, the cpr class can just save that value and then return that value on the next progress call, but it wouldn't be immediate. However it should be fast enough, and i think it is 100% worth that trade-off of a max 1 second delay (usually only  like 0.3 second) to instead allow the user full access to `curl_easy_getinfo`.

include/cpr/progresscb.h
```diff
@@ -0,0 +1,28 @@
+#ifndef CPR_PROGRESSCB_H
+#define CPR_PROGRESSCB_H
+
+namespace cpr {
+
+class ProgressCallback {
+  #if LIBCURL_VERSION_NUM < 0x072000 // older than 7.32.0 uses the older callback.
+  typedef int(*progress_cb_t)(void*, double, double, double, double);
+  #else
+  typedef int(*progress_cb_t)(void*, int64_t, int64_t, int64_t, int64_t);
```
COM8, line 10: Based on the documentation, the callback requires `curl_off_t` as parameter.
https://curl.haxx.se/libcurl/c/progressfunc.html

Perhaps we should create a `typedef curl_off_t progress_t;` for `curl_off_t` so we do not expose this type.
So other other applications do not need to link against libcurl.

@KingKili, what do you think about this one?

include/cpr/progresscb.h
```diff
@@ -0,0 +1,28 @@
...
+  #endif
+
+  public:
+    ProgressCallback() = default;
+    ProgressCallback(const progress_cb_t cb) : cb{cb} {}
+    ProgressCallback(const progress_cb_t cb, void * userpt)
```
COM8, line 16: Placement of `*` should be on the left to match.
-> `void* userptr`

include/cpr/progresscb.h
```diff
@@ -0,0 +1,28 @@
...
+          : cb{cb}, userpt{userpt} {}
+
+    void AddCallback(const progress_cb_t callback) { cb = callback; }
+    void AddUserData(void *userData) { userpt = userData; }
```
COM8, line 20: `void* userData`

COM8:
Hi @ITotalJustice, thank you for taking the time to add support for this this feature to `CPR`!

COM8:
> I was thinking, would it be a good idea to create 2 more classes for write and read callbacks? I could rename `progresscb.h` to `callback.h` and then have the 3 classes there.
> 
> Iirc there's a few more libcurl callbacks also, such as headers. Maybe it might be a good option for me to add those also?
> 

I agree. This would be a great idea. I'd say go for it.

> EDIT: Not sure why one of the tests fails. It says it's the callback test. I checked if maybe i named something the same as a func / var in the test, but i haven't. I did change 2 const vars to constexpr, but surely it can't be that thinking

Don't worry about macOS. It fails for every second commit and we currently do not know why.

COM8:
> [...]
> EDIT: After reading the easy curl header, i found this in regards to the `curl_easy_getinfo` _**This function is intended to get used _AFTER_ a performed transfer, all results from this function are undefined until the transfer is completed.**_. I am not sure why a lot of liburls examples go completely against this statement...
> If the results are undefined then this makes this whole idea (in this message) pretty irrelevant.

I also do not think, we should use `curl_easy_getinfo` for progress callbacks.
To me it looks like a good idea to give the user access to the `curl_easy_getinfo` results, but those results should then be made available after the request in the `Response` class of the request.
I will create a PR for this.

COM8:
Thinks that have to be done before a merge:
- [ ] Please create a PR against the `gh-pages` branche to add documentation for this great feature.
- [ ] Update the [README.md](https://github.com/whoshuu/cpr#features) features list.
- [ ] Add unit tests

include/cpr/progresscb.h
KingKili: Thats true, but we need to make sure that we notice when libcurl changes the type.

COM8: @ITotalJustice You might want to rebase and have a look at `cprtypes.h` where I defined a wrapper for `curl_off_t` in 943194f16fbb68b406a00e7404321d886031b026.

ITotalJustice:
I've rebased against master, however it's now showing the master commit history in this PR. Is this correct? It's my first time rebasing haha, I hope that I didn't mess anything up.

I'll work on the documentation now anyway.

COM8:
> I've rebased against master, however it's now showing the master commit history in this PR. Is this correct? It's my first time rebasing haha, I hope that I didn't mess anything up.
> 
> I'll work on the documentation now anyway.

Interesting, but should be fine :)

ITotalJustice:
So i've changed how the progress callback works. no longer is the user function directly passed to libcurl, instead, we wrap the function and pass everything in a nice tidy struct like this:
```
int ProgressCallbackFunction(ProgressCallbackData* data, cpr_off_t dlnow, cpr_off_t dltotal, cpr_off_t ulnow, cpr_off_t ultotal) {
    data->data.downloadNow = dlnow;
    data->data.downloadTotal = dltotal;
    data->data.uploadNow = ulnow;
    data->data.uploadTotal = ultotal;
    return data->cb(&data->data);
  }
```
which then for the user, all they have to do is this:
```
int myCallback(cpr::ProgressCallbackUser *data) {
    MyData *myData = (MyData*)data->userData;
    std::cout << myData->message << std::endl;
    return 0;
}
```
I think this is ideal because it makes the callback struct for the user much simpler (mirrors how the response class works) and also because we can add more info to the struct if libcurl api ever changes _without_ the user having to change anything.

The names i've given everything are just temporary, so let me know what you would like to rename them as. Also let me know where I should put the structs / function, or are you happy with where they are?

ITotalJustice:
Sorry for making another big change 😄 

I've added read and write callback support. This will allow for users to write data to disk in chunks, rather than storing it all in memory. This closes a few more issues such as #255 #208 #111 maybe #25 #271

include/cpr/callback.h
```diff
@@ -0,0 +1,106 @@
...
+
+    void AddCallback(const progress_cb_t cb) { callback.cb = cb; }
+    void AddUserData(void* userData) { callback.data.userData = userData; }
+
+    ProgressCallbackData callback;
+};
+
+} // namespace cpr
+
+#endif
```
COM8, line 106: Please add one empty line at the end.

COM8:
I think something went wrong during the rebase earlier.
I'd suggest you try it again, since there are a lot of conflicts, that should not be conflicts.
For reference: https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa

include/cpr/callback.h
```diff
@@ -0,0 +1,106 @@
...
+  private:
+    typedef struct {
+    progress_cb_t cb;
+    ProgressCallbackUser data;
+  } ProgressCallbackData;
+
+  #if LIBCURL_VERSION_NUM < 0x072000
+  int ProgressCallbackFunction(ProgressCallbackData* data, double dlnow, double dltotal, double ulnow, double ultotal) {
+  #else
+  int ProgressCallbackFunction(ProgressCallbackData* data, cpr_off_t dlnow, cpr_off_t dltotal, cpr_off_t ulnow, cpr_off_t ultotal) {
```
COM8, line 81: Please rename the parameters to lower camel case so they match the ones above (`downloadNow`, `downloadTotal`, `uploadNow`, `uploadTotal`).

ITotalJustice: So I should change
```
int ProgressCallbackFunction(ProgressCallbackData* data, double dlnow, double dltotal, double ulnow, double ultotal)
```
to
```
int ProgressCallbackFunction(ProgressCallbackData* data, double downloadNow, double downloadTotal, double uploadNow, double uploadTotal)
```
?

COM8: Yes, please.

ITotalJustice:
I rebased following the guide that you sent (thank you!). Did this fix the issues?

xloem:
Hey, I found this PR and see it's been sitting for about a month.  Do you guys need help with the tests and documentation situation, or do you just need some time?

COM8:
Only the documentation for the return values is still pending.
Besides that, this PR is ready to be merged.

cpr/session.cpp
```diff
@@ -97,6 +104,38 @@ Session::Impl::Impl() {
     }
 }
 
+void Session::Impl::SetReadCallback(const ReadCallback& read) {
+    CURL* curl = curl_->handle;
+    if (curl) {
+        curl_easy_setopt(curl, CURLOPT_READFUNCTION , read.callback.cb);
```
xloem, line 110: I'm looking at this patch a little and noticed that CURLOPT_READFUNCTION is assigned a value of a type of `int(*read_cb_t)(ReadCallbackUser*);` (types in callback.h), whereas https://curl.haxx.se/libcurl/c/CURLOPT_READFUNCTION.html specifies a type of `size_t (*read_callback)(char *buffer, size_t size, size_t nitems, void *userdata);`.  It seems like that would cause a crash if used, due to the change: am I mistaken?

cpr/session.cpp
```diff
@@ -97,6 +104,38 @@ Session::Impl::Impl() {
...
+        curl_easy_setopt(curl, CURLOPT_READDATA, &read.callback.data);
+        this->userRead_ = true;
```
xloem, line 112: this member variable appears unused

xloem:
I'm not sure this PR is complete; it kind of looks like a proof of concept waiting for confirmation around design choices, maybe even for somebody else to finish when needed.  It looks like most of the work is in the progress callback.  I might try to contribute a little.

ITotalJustice:
Just got an email notification for this, very sorry again for late to replying (can't believe it's been a month already!)

@xloem yeah it was mainly a quick draft. i put it together in like an hours because i needed to early cancel api requests. There's a few(?) unused variables as i changed the design of the pr a few times over. It would be great if you can contribute to the pr (or a new pr). I don't really have much free time sadly

xloem:
I'm spending some time bringing this together locally.  I expect to finish the tests, likely the docs.

xloem:
Completed work for this PR at https://github.com/ITotalJustice/cpr/pull/1

COM8:
Closing this one in favour of #446.

Issue #439, opened 2020-08-27
Consider change curl download to http to save time


summershrimp:
https://github.com/whoshuu/cpr/blob/52bb78e4928ddfc32861c25c99fd29a420e22eb2/CMakeLists.txt#L92-L96
FetchContent with git repository would download the whole curl repository about 70MiB for now, but compress package for specific version is only ~3MiB. 
Such as:
```
   FetchContent_Declare(
        curl
        URL https://github.com/curl/curl/releases/download/curl-7_72_0/curl-7.72.0.tar.xz
        URL_HASH  SHA256=0ded0808c4d85f2ee0db86980ae610cc9d165e9ca9da466196cc73c346513713
    )
```

COM8:
Hey, good idea. Thanks.
Do you mind creating a PR for this?

PR #440, opened 2020-08-31
Change external projects download method to https

summershrimp:
Since CMake FetchContent with git would clone the whole repository,
using http to download only specific version would save a lot of
time and space.

Related issue #439 

Signed-off-by: Yibai Zhang <xm1994@gmail.com>


COM8 (approved):

COM8:
Nice. Thanks!

@KingKili, do you have any concerns about this?

KingKili:
No I can think of any reason why not. I think it's a pretty neat idea.

PR #441, opened 2020-08-31
Fix: typo in curlholder.cpp

summershrimp:


COM8:
Aha 😆.
Good catch. Thanks!

PR #443, opened 2020-09-02
Add Parameters::AddParameter overload without CurlHolder

alex-87:
To use the current `void Parameters::AddParameter(const Parameter& parameter, const CurlHolder& holder)` requires to create a `CurlHolder` object before to call the method.

This pull request consists in overloading the method by requiring the `const Parameter& parameter` argument only. The `CurlHolder` object is now created before to call the original method `void Parameters::AddParameter(const Parameter& parameter, const CurlHolder& holder)`.

### DIFF
--- a/cpr/CMakeLists.txt
+++ b/cpr/CMakeLists.txt
@@ -4,6 +4,7 @@ add_library(cpr
     auth.cpp
     cookies.cpp
     cprtypes.cpp
+    curl_container.cpp
     curlholder.cpp
     error.cpp
     multipart.cpp
--- a/cpr/curl_container.cpp
+++ b/cpr/curl_container.cpp
@@ -1,0 +1,62 @@
+#include "cpr/curl_container.h"
+
+
+namespace cpr {
+
+
+template<class T>
+CurlContainer<T>::CurlContainer(const std::initializer_list<T>& containerList) : containerList_(containerList) {
+
+}
+
+
+template<class T>
+void CurlContainer<T>::Add(const std::initializer_list<T>& containerList) {
+    for(const auto& element : containerList) {
+        containerList_.push_back(std::move(element));
+    }
+}
+
+template<class T>
+void CurlContainer<T>::Add(const T& element) {
+    containerList_.push_back(std::move(element));
+}
+
+template<>
+const std::string CurlContainer<Parameter>::GetContent(const CurlHolder& holder) const {
+    std::string content{};
+    for(const auto& parameter : containerList_) {
+        if (!content.empty()) {
+            content += "&";
+        }
+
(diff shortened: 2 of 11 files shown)


COM8:
Hey @alex-87 👋🏻.
Thanks for creating this PR.

I also thought about this, when implementing the `CurlHolder`.
I decided against providing an overload for `AddParameter` because of the following reason:

Creating a `CurlHolder` object is really expensive. We have to acquire a lock and then initialize a new curl session.
My intent was to prevent the following from happening:

```c++
Parameters paras;
for(size_t i = 0; i < 1000; i++) {
    paras.AddParameter(someParameter);
}
```

This behavior would be awfully inefficient.

So I decided to force the user to create a `CurlHolder` by their own.

What do you think about this?

alex-87:
Thank you for your answer,

I agree, the `CurlHolder`'s initialization is expensive :thinking: 

After a deeper look, I noticed the `Parameters::content` attribute is read 4 times, all in [session.cpp](https://github.com/whoshuu/cpr/blob/master/cpr/session.cpp) (2 for the emptiness string check, 2 for the content).

I also noticed that `Session::Impl` has its own [`std::unique_ptr<CurlHolder>`](https://github.com/whoshuu/cpr/blob/97d9b865dddc16aa677b7cb89fea9c02e1a35848/cpr/session.cpp#L63) instance.

In order for the users to be able do an `AddParameter` loop, would a solution to be :

- To store the parameters in a `std::vector<Parameter>` attribute, in `Parameters` (filled with constructor and possibly with  `AddParameter`)
- For the `Parameters` constructor to not create a `CurlHolder` object
- To benefit from the already instanciated `std::unique_ptr<CurlHolder> Session::Impl::curl_` instance for building the URL-encoded parameters using a new `std::string getParameters(const CurlHolder&) const` method

This would avoid creating a `CurlHolder` object, while using an already existing instance during the parameter reading phase only.

What do you think about this ?

COM8:
This sounds like a good idea.
Do you want to create a PR for this?

alex-87:
I have improved this current pull request with the following modifications : 

- Removing my previous modifications
- Introducing new `CurlContainer` generic object for support of `Parameters` and `Payload`
- Modifying `Parameters` and `Payload` to use `CurlContainer` features by a derivative specialization
- Introducing the `Add` method for `Parameters` and `Payload` (both provided by `CurlContainer`)
- Introducing the `GetContent(const CurlHolder&)` method for `Parameters` and `Payload` to retrieve the encoded content using the already existing `CurlHolder` instance in Session
- Modification of `Session` for the use of `GetContent(const CurlHolder&)` method for both `Parameters` and `Payload`
- Adapting corresponding tests

COM8 (approved):
Great work!
Thanks.

PR #459, opened 2020-10-20
Implement Url Decoding

luca-schlecker:
I found myself needing to decode text. This function does the opposite of `cpr::util::urlEncode(const std::string& s)` using `curl_easy_unescape`.

Signed-off-by: Luca Schlecker <luca.schlecker@hotmail.com>

### DIFF
--- a/cpr/util.cpp
+++ b/cpr/util.cpp
@@ -146,7 +146,7 @@ std::string urlEncode(const std::string& s) {
 
 /**
  * Creates a temporary CurlHolder object and uses it to unescape the given string.
- * If you plan to use this methode on a regular basis think about creating CurlHolder
+ * If you plan to use this methode on a regular basis think about creating a CurlHolder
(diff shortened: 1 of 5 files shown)


cpr/util.cpp
```diff
@@ -144,5 +144,20 @@ std::string urlEncode(const std::string& s) {
     return holder.urlEncode(s);
 }
 
+/**
+ * Creates a temporary CurlHolder object and uses it to unescape the given string.
+ * If you plan to use this methode on a regular basis think about creating CurlHolder
```
COM8, line 149:
```suggestion
 * If you plan to use this methode on a regular basis think about creating a CurlHolder
```

COM8 (changes requested):
Only one small typo. Great addition. Thanks!

COM8 (approved):

PR #462, opened 2020-10-22
add guards for curl version

xloem:
I haven't actually been able to compile 1.5.2 for my system and projects yet (still using 1.4.0), but while attempting to I bumped into this, which might be missing.

### DIFF
--- a/cpr/response.cpp
+++ b/cpr/response.cpp
@@ -13,8 +13,16 @@ Response::Response(std::shared_ptr<CurlHolder> curl, std::string&& p_text,
     char* url_string{nullptr};
     curl_easy_getinfo(curl_->handle, CURLINFO_EFFECTIVE_URL, &url_string);
     url = Url(url_string);
+#if LUBCURL_VERSION_NUM < 0x075500
+    double downloaded_bytes_double, uploaded_bytes_double;
+    curl_easy_getinfo(curl_->handle, CURLINFO_SIZE_DOWNLOAD, &downloaded_bytes_double);
+    curl_easy_getinfo(curl_->handle, CURLINFO_SIZE_UPLOAD, &uploaded_bytes_double);
+    downloaded_bytes = downloaded_bytes_double;
+    uploaded_bytes = uploaded_bytes_double;
+#else
     curl_easy_getinfo(curl_->handle, CURLINFO_SIZE_DOWNLOAD_T, &downloaded_bytes);
     curl_easy_getinfo(curl_->handle, CURLINFO_SIZE_UPLOAD_T, &uploaded_bytes);
+#endif
     curl_easy_getinfo(curl_->handle, CURLINFO_REDIRECT_COUNT, &redirect_count);
 }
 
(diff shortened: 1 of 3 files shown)


cpr/response.cpp
```diff
@@ -13,8 +13,16 @@ Response::Response(std::shared_ptr<CurlHolder> curl, std::string&& p_text,
     char* url_string{nullptr};
     curl_easy_getinfo(curl_->handle, CURLINFO_EFFECTIVE_URL, &url_string);
     url = Url(url_string);
+#if LUBCURL_VERSION_NUM < 0x075500
```
COM8, line 16:
```suggestion
#if LUBCURL_VERSION_NUM < 0x070401
```
Based on the docs for `CURLINFO_SIZE_DOWNLOAD` and `CURLINFO_SIZE_UPLOAD`, shouldn't this be `0x070401`, since it's available since `7.4.1`.

https://curl.haxx.se/libcurl/c/CURLINFO_SIZE_DOWNLOAD.html

cpr/session.cpp
```diff
@@ -365,7 +365,9 @@ void Session::Impl::SetSslOptions(const SslOptions& opts) {
 #endif
     curl_easy_setopt(curl_->handle, CURLOPT_SSL_VERIFYPEER, opts.verify_peer ? ON : OFF);
     curl_easy_setopt(curl_->handle, CURLOPT_SSL_VERIFYHOST, opts.verify_host ? 2L : 0L);
+#if LIBCURL_VERSION_NUM >= 0x074100
```
COM8, line 368:
```suggestion
#if LIBCURL_VERSION_NUM >= 0x072900
```
Available since 7.41.0 -> 0x07, 0x29, 0x00
Curl uses hex numbers here ;)

https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYSTATUS.html

https://curl.haxx.se/docs/versions.html

COM8 (changes requested):
Good thing. I think we should add additional CI runs for different curl versions, to prevent something like this.

cpr/session.cpp
xloem: ha thanks

cpr/response.cpp
xloem: `CURLINFO_SIZE_DOWNLOAD_T` is available since 7.55.0 https://curl.haxx.se/libcurl/c/CURLINFO_SIZE_DOWNLOAD_T.html .  I need to convert to hex.

xloem: I converted to hex; I didn't add a case for < 7.4.1 because it doesn't look like cpr is designed to support that version, and I haven't reviewed other areas of code or tested with the version.

COM8 (approved):

COM8:
Great additions. Thanks.
Is this PR from your side ready to be merged?

xloem:
As far as I know.

I have other issues on the queue to test this latest version locally, but the changes all look harmless and the tests pass.

PR #473, opened 2020-11-09
Update verbose.h

Zwolf8:
Removed a (unused?) cstdbool include that caused C++17 deprecation errors when building on MSVC.

[https://en.cppreference.com/w/cpp/header/cstdbool](Source)


COM8 (approved):
Great, thanks!

PR #475, opened 2020-11-15
Fixed implicit conversion

GiuseppeCesarano:
std::map automatically const-ify the key type, so this changed for each
caused an inplicit conversion.

### DIFF
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -3,7 +3,7 @@
 namespace cpr {
 std::string Cookies::GetEncoded(const CurlHolder& holder) const {
     std::stringstream stream;
-    for (const auto& item : map_) {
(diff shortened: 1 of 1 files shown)


cpr/cookies.cpp
```diff
@@ -3,7 +3,7 @@
 namespace cpr {
 std::string Cookies::GetEncoded(const CurlHolder& holder) const {
     std::stringstream stream;
-    for (const std::pair<std::string, std::string>& item : map_) {
+    for (const auto& item : map_) {
```
COM8, line 6: Since I hate `auto`, how about this:
```suggestion
    for (const std::pair<const std::string, std::string>& item : map_) {
```

COM8 (changes requested):
Good idea!

COM8 (approved):
Great. Thanks :)

PR #479, opened 2020-11-18
Changed session implementation

### LINKED ISSUE #476: Concerns on session.cpp implementation
Looking at the source code of [session.cpp](https://github.com/whoshuu/cpr/blob/master/cpr/session.cpp) i noticed that move constructors are manually implemented and no particular behavior is involved, so i would suggest to remove the blank destructor, and allow the compiler to automatically generate those for us. Note that if is decided to keep things the way they are now, those constructors should anyway be declared as `noexcept`. There also others small things, like the usage of iterator based for instead of range based one like in all the other code base and a [call to std::move() on an c_str()](https://github.com/whoshuu/cpr/blob/47cc92045247c429037324cb4eee67ede81c35cf/cpr/session.cpp#L186) which is already trivially copyable type since is a `char*`.

GiuseppeCesarano:
This implementation allows the compiler to generate the needed
constructor for the class, so there is no more need to implement them.
This commit also removes a std::move call on a trivially copyable type
and rewrites a for loop to use the range syntax. Closes #476

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -16,73 +16,6 @@ namespace cpr {
 constexpr long ON = 1L;
 constexpr long OFF = 0L;
 
-class Session::Impl {
-  public:
-    Impl();
-
-    void SetUrl(const Url& url);
-    void SetParameters(const Parameters& parameters);
-    void SetParameters(Parameters&& parameters);
-    void SetHeader(const Header& header);
-    void SetTimeout(const Timeout& timeout);
-    void SetConnectTimeout(const ConnectTimeout& timeout);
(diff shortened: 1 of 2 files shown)


GiuseppeCesarano:
Please note that the CI is failing due something not related with my patch, all the tests still pass.

COM8:
Sorry, I closed the wrong issue with my commit.
Could you please rebase your changes.

GiuseppeCesarano:
I'm closing this PR since 96e33bae7a5282ae5a3cdac22c83499e0848d856 does exactly what this PR was trying to do.

PR #480, opened 2020-11-18
Changed " " to ' ' for single character

GiuseppeCesarano:
std::string has a `find(char)` overload. Using it instead of passing a string literal of 1 character.


COM8:
I already fixed this one. Thanks.

PR #488, opened 2020-12-12
Add connection pool option

cleaton:
This is a re-opening of PR #330

### DIFF
--- a/test/abstractServer.cpp
+++ b/test/abstractServer.cpp
@@ -51,6 +51,11 @@ static void EventHandler(mg_connection* conn, int event, void* event_data) {
 
         case MG_EV_HTTP_REQUEST: {
             AbstractServer* server = static_cast<AbstractServer*>(conn->mgr->user_data);
+            auto port_cstr = (char*) malloc(256);
+            auto r = mg_conn_addr_to_str(conn, port_cstr, 256, MG_SOCK_STRINGIFY_REMOTE | MG_SOCK_STRINGIFY_PORT);
+            int port = atoi(port_cstr);
+            server->AddConnection(port);
+            free(port_cstr);
             server->OnRequest(conn, static_cast<http_message*>(event_data));
         } break;
 
@@ -78,6 +83,18 @@ void AbstractServer::Run() {
 
     // Notify the main thread that we have shut down everything:
     server_stop_cv.notify_all();
+}
+
+void AbstractServer::AddConnection(int remote_port) {
+    unique_connections.insert(remote_port);
+}
+
+size_t AbstractServer::GetConnectionCount() {
+    return unique_connections.size();
+}
+
+void AbstractServer::ResetConnectionCount() {
+    unique_connections.clear();
 }
 
 static const std::string base64_chars =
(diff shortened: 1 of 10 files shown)


test/abstractServer.cpp
```diff
@@ -51,6 +51,11 @@ static void EventHandler(mg_connection* conn, int event, void* event_data) {
 
         case MG_EV_HTTP_REQUEST: {
             AbstractServer* server = static_cast<AbstractServer*>(conn->mgr->user_data);
+            auto port_cstr = (char*) malloc(256);
```
COM8, line 54: Please do not use auto where ever possible.

`malloc` is C, consider replacing it with `std::array`.

COM8 (commented):
Thanks for the PR. With a bit of additional work, this will be quiet the nice addition to CPR.

test/abstractServer.cpp
cleaton: The mongoose webserver mg_conn_addr_to_str requires a char* buffer as input as it's written in C. Would it be better I use stack allocation for the char array?

I'll remove auto for the obvious types.

cleaton:
Currently the osx tests are failing with `libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`. I have not been able to reproduce it on linux so far.

test/abstractServer.cpp
COM8: You can get the underlying `char*` from a `std::array` with the `std::array::data()` call: https://en.cppreference.com/w/cpp/container/array/data

In this case it's better tu use stack memory here sine than we do not have to care about free any more.

COM8:
> Currently the osx tests are failing with `libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`. I have not been able to reproduce it on linux so far.

Oh, how I love macOS.
I will have a look at it.

cleaton:
> > Currently the osx tests are failing with `libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`. I have not been able to reproduce it on linux so far.
> 
> Oh, how I love macOS.
> I will have a look at it.

I am wondering if make_shared for the mutex is failing to initialize the underlying mutex implementation?

COM8:
To me it looks like a use after free, but I do not know where.
I will try enable stack traces in gtest so we have a clear indication where this is happening later today.

COM8:
> > > Currently the osx tests are failing with `libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`. I have not been able to reproduce it on linux so far.
> > 
> > 
> > Oh, how I love macOS.
> > I will have a look at it.
> 
> I am wondering if make_shared for the mutex is failing to initialize the underlying mutex implementation?

This could be a reason.

zodinyac:
Hi, @cleaton! What is current state of this PR? Will you finalize it?

cleaton:
Hi @zodinyac! At the moment I am working on other projects so I won't have time to work on this PR anytime soon. I'll leave it here in case anyone wants to take over and resolve the macos test issue.

bokket:
Hi, @cleaton .Why not try to use the `std::shared_mutex` in c++17 ?
I think this lock is more useful than `std::shared_ptr<std::mutex>` inside a pool of multiple possible connect objects. After all, this is the native synchronization primitive.
https://stackoverflow.com/questions/46452973/difference-between-stdmutex-and-stdshared-mutex

COM8:
Closing this for now since a lot changed since back then.

PR #489, opened 2020-12-15
Removed explict settings of HTTP type for Get and Post requets

Kupercraft:
Setting http request types supported by Curl is dicouraged according to the documentation.
See:
https://curl.se/libcurl/c/CURLOPT_CUSTOMREQUEST.html
https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/
for more details.

This patch might be applicable to other http requests currently supported.
But I don't think that my understanding of the current usage is good enough for making those changes.

Thanks.


COM8 (approved):
Great thanks.

PR #492, opened 2020-12-20
Problem 491 was fixed.

MateuszMiekicki:
The problem was the lack of linking of necessary libraries for Windows. The problem occurred during the construction of the tests, at least there it was noticed.


COM8:
Great, thanks.

PR #497, opened 2021-01-06
Add job to check compilation on 18.04 with system curl

prudhomm:
see whoshuu/cpr#496 for more details. 
cpr does not build with system curl on 18.04


COM8:
Thanks for the PR.
Closing this since I changed stuff and added this my self.

PR #499, opened 2021-01-09
V1.5.2 pre

OmriBaso:
Add an options to get the response data as bytes and not as text decoded.
Here is an example from a python code that uses Pycurl

Actually its a library i wrote my self called besocurl

You can see the code here: https://github.com/OmriBaso/besoCurl
If you dont want to go into my repo and see the code he is a demonstartion 
image

### DIFF
--- a/cpr/curlholder.cpp
+++ b/cpr/curlholder.cpp
@@ -2,8 +2,7 @@
 #include <cassert>
 
 namespace cpr {
-// NOLINTNEXTLINE (cert-err58-cpp)
-std::mutex curl_easy_init_mutex_;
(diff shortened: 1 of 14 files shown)


COM8:
You re right, this should be added to CPR.
This is especially usefull in case you need custom encodings.
But I don't think your PR is adding anything for this. Especially why dou you want to merge `v1.5.2_pre` into `master`?
All of these changes (as far as I'm aware) are already present in `master`.

COM8:
Closing this since there does not seam to be any activity for some time. Feel free to reopen in case this is still an issue.

Issue #495, opened 2021-01-06
Segfault when running tests


EmRowlands:
I am packaging cpr for Gentoo, but I have run into an issue with the tests: most of them segfault in my environment. I'm not particularly familiar with the source (I'm packaging as a dependency) so I ask here for help.

I don't think it's a missing dependency (everything builds correctly and `ldd` shows nothing obviously wrong), and I don't think it's my CFLAGS.

Log files are below:
Test log: [LastTest.log](https://github.com/whoshuu/cpr/files/5775625/LastTest.log)
Build log/system configuration: [dev-cpp:cpr-1.5.2:20210106-120325.log](https://github.com/whoshuu/cpr/files/5775656/dev-cpp.cpr-1.5.2.20210106-120325.log)

Basic system information:
```
sh dash 0.5.11.2
ld GNU ld (Gentoo 2.34 p6) 2.34.0
distcc 3.3.3 x86_64-pc-linux-gnu [disabled]
app-shells/bash:          5.0_p18::gentoo
dev-util/cmake:           3.17.4-r1::gentoo
sys-devel/binutils:       2.34-r2::gentoo
sys-devel/gcc:            10.2.0-r5::gentoo
sys-devel/gcc-config:     2.3.2-r1::gentoo
sys-devel/make:           4.3::gentoo
sys-kernel/linux-headers: 5.9::gentoo (virtual/os-headers)
sys-libs/glibc:           2.32-r3::gentoo
```

COM8:
Interesting one. Thanks for reporting.
To me it looks like you are using `cpr-1.5.2`.
Could you please try building the master branche of cpr and running all tests on your system?

EmRowlands:
Test log: [LastTest.log](https://github.com/whoshuu/cpr/files/5781698/LastTest.log)
Build log/system configuration: 
[dev-cpp:cpr-9999:20210107-125556.log](https://github.com/whoshuu/cpr/files/5781702/dev-cpp.cpr-9999.20210107-125556.log)

Seems to be the same result.

COM8:
Do you know if there exists any up to date Gentoo live image?

EmRowlands:
Gentoo doesn't really have a live image, but you can find "stage 4" tarballs which can be unzipped and used as a chroot.
See here for more info: https://wiki.gentoo.org/wiki/Stage_tarball#Stage_4

I also did some investigation myself. I'm not really familiar with mongoose, but it appears `mg_bind` is returning a null pointer.

[...]

PR #502, opened 2021-01-16
Fix segfaults in tests (#495)

EmRowlands:
When the port in `GetPort()` is already in use, `mg_bind` returns `null`,
leading to a segfault in `mg_set_protocol_http_websocket`. This PR
causes an easily readable error when this occurs. Since this is caused
by a library function, `errno` is used to get the precise error.

This PR also changes the ports used in the test suite to be in the
ephemeral range, making it much less likely that a collision occurs.

Closes: #495

### DIFF
--- a/test/httpServer.cpp
+++ b/test/httpServer.cpp
@@ -1,5 +1,6 @@
 #include "httpServer.hpp"
 #include <string>
+#include <system_error>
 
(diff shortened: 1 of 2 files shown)


COM8 (approved):
Good job. Thanks.

PR #512, opened 2021-02-08
Support CURLSSLOPT_NO_REVOKE

dirkvdb:
I need to set the CURLSSLOPT_NO_REVOKE option when connecting through the company proxy to avoid revocation errors.
(Curl using WINSSL)


COM8 (approved):
Thanks for this PR. Thanks.

PR #521, opened 2021-02-26
Updated curl from 7.69.1 to 7.75.0

COM8:


COM8:
Closed in favour of #529

PR #525, opened 2021-03-05
Add tests for static builds

KingKili:
Added CI tests also for static builds, it is possible that this tests are still failing


KingKili:
@COM8 Can you see why the CI does not run?

KingKili:
> @COM8 Can you see why the CI does not run?

Never mind found it myself

PR #529, opened 2021-03-13
CMake SSL and Windows SSL fixes

COM8:
This PR refactores the CMake SSL integration and aims to fix SLL and especially Windows SSL issues.
Changes:
* Updated curl from `7.69.1` to `7.75.0` to allow loading the root CA for OpenSSL under Windows (makes #521 obsolete)
* Added zlib or more concrete [zlib-ng](https://github.com/zlib-ng/zlib-ng) as a dependency since it's required by curl and a pain to install on Windows
* Fixed Windows OpenSSL build
* Renamed all exposed CMAKE CPR variables with "CPR_" as prefix
* CPR will automatically detect your system and try to set the appropriate SSL backend for you (WinSSL or OpenSSL)

#### With WinSSL
```bash
mkdir build
cd build
cmake .. -DCPR_BUILD_TESTS_SSL=OFF # SSL test are only supported with OpenSSL
cmake --build .
```

#### With OpenSSL
```bash
mkdir build
cd build
cmake ..  -DCPR_FORCE_OPENSSL_BACKEND=ON # Disable auto detect and force OpenSSL
cmake --build .
```


COM8:
@KingKili could you please review this PR since I changed a lot on the build system.

CMakeLists.txt
```diff
@@ -29,32 +28,90 @@ endmacro()
...
-cpr_option(USE_OPENSSL "Use OpenSSL code. Experimental" ON)
-cpr_option(USE_WINSSL "Use WIN_SSL backend. Experimental" OFF)
+cpr_option(CPR_USE_SYSTEM_GTEST "If ON, this project will look in the system paths for an installed gtest library. If none is found it will use the build in one." OFF)
+cpr_option(CPR_FORCE_USE_SYSTEM_CURL "If enabled we will use the curl lib already installed on this system." OFF)
+cpr_option(CPR_ENABLE_SSL "Enables or disables the SSL backend. Required to perform HTTPS requests." ON)
+cpr_option(CPR_DETECT_SSL_BACKEND "Automatically detects the available SSL backends and uses the recommend for your system in case multiple are available. (WinSSL - Windows, OpenSSL - Linux, Mac, ...)" ON)
+cpr_option(CPR_FORCE_OPENSSL_BACKEND "Force to use the OpenSSL backend. For this to function set CPR_DETECT_SSL_BACKEND to OFF." OFF)
+cpr_option(CPR_FORCE_WINSSL_BACKEND "Force to use the WinSSL backend. For this to function set CPR_DETECT_SSL_BACKEND to OFF." OFF)
+cpr_option(CPR_BUILD_TESTS "Set to ON to build cpr tests." ON)
+cpr_option(CPR_BUILD_TESTS_SSL "Set to ON to build cpr ssl tests" ${CPR_BUILD_TESTS})
```
KingKili, line 40: I think it is a bad idea to rename the variables, since it will lead to problems in depending projects and especially package manager.

CMakeLists.txt
```diff
@@ -29,32 +28,90 @@ endmacro()
...
-endif()
-
-if ((NOT WIN32) AND USE_WINSSL)
-    message(FATAL_ERROR "Only use WINSSL with windows")
-endif()
-
-
 include(GNUInstallDirs)
 include(FetchContent)
+include(ExternalProject)
```
KingKili, line 45: Is this really required, I cannot see any use of it

CMakeLists.txt
```diff
@@ -29,32 +28,90 @@ endmacro()
...
+        elseif(CPR_FORCE_WINSSL_BACKEND)
+            if(WINSSL_AVAILABLE)
+                message(STATUS "Using WinSSL.")
+                set(SSL_BACKEND_USED "WinSSL")
+            else()
+                message(FATAL_ERROR "CPR_FORCE_OPENSSL_BACKEND enabled but we were not able to find OpenSSL!")
+            endif()
+        endif()
+    endif()
+endif()
```
KingKili, line 113: Is this massive detection logic necessary?
Can't we just rely on setting the requested configuration according to the user request?

CMakeLists.txt
```diff
@@ -73,46 +130,66 @@ else ()
...
         set(CMAKE_USE_OPENSSL ON CACHE INTERNAL "" FORCE)
     endif()
 
     # Show progress of FetchContent:
     set(FETCHCONTENT_QUIET OFF CACHE INTERNAL "" FORCE)
     FetchContent_Declare(curl
-                         URL                    https://github.com/curl/curl/releases/download/curl-7_69_1/curl-7.69.1.tar.xz
-                         URL_HASH               SHA256=03c7d5e6697f7b7e40ada1b2256e565a555657398e6c1fcfa4cb251ccd819d4f # the file hash for curl-7.69.1.tar.xz
+                         URL                    https://github.com/curl/curl/releases/download/curl-7_75_0/curl-7.75.0.tar.xz
+                         URL_HASH               SHA256=fe0c49d8468249000bda75bcfdf9e30ff7e9a86d35f1a21f428d79c389d55675 # the file hash for curl-7.75.0.tar.xz
```
KingKili, line 191: Looks good, but we should discuss if we want to link statically against libcurl and zlib

KingKili (changes requested):

CMakeLists.txt
gbMichelle: A project I'm working on had to switch to this branch a few days ago because OpenSSL wasn't working.

This piece of the cmakelist really threw me for a spin. I think having both `CPR_DETECT_SSL_BACKEND` and `CPR_FORCE_OPENSSL_BACKEND` as separate vars can initially be confusing.

I'd personally suggest allowing users to set a var that will directly affect `SSL_BACKEND_USED` as I initially admittedly tried to set the backend by doing `set(SSL_BACKEND_USED "OpenSSL")`

COM8: Also it hurts to say, I think you guys are right.
I would suggest removing the `CPR_DETECT_SSL_BACKEND` variable and fall back to it in case the user did not set either `CPR_FORCE_OPENSSL_BACKEND` or `CPR_FORCE_WINSSL_BACKEND`.
This allows new users to get up and running quickly without having to think about the backend and the complexity is reduced for advanced users.
Any thoughts?

gbMichelle: I think that would be the right move. `CPR_FORCE_OPENSSL_BACKEND` and `CPR_FORCE_WINSSL_BACKEND` only ever do anything if `CPR_DETECT_SSL_BACKEND` is set to `OFF`.

I see no intentional reason to have `CPR_DETECT_SSL_BACKEND` `ON` while either `CPR_FORCE_OPENSSL_BACKEND` or `CPR_FORCE_WINSSL_BACKEND` are `ON`.

The `find_package` for OpenSSL should probably also only be in the `CPR_FORCE_OPENSSL_BACKEND` block as I think that will actually count as an error causing configure to be considered incomplete if not found. Which would be an issue if someone intends to use WinSSL and doesn't have OpenSSL available.
Some people will also have custom `find<PackageName>.cmake` scripts and those will be triggered by using `find_package`

COM8: Done.

CMakeLists.txt
COM8: Yes, but now we have a clear way to show which variables are offered by cpr.

CMakeLists.txt
COM8: Yes, this can be changed in a separate PR.

PR #531, opened 2021-03-24
Do not move targets in to a bin if cpr is a SubProject

KingKili:


COM8:
Nice. Thanks.

KingKili:
> Nice. Thanks.

The pleasure is mine

Issue #530, opened 2021-03-15
Reserve space beforehand for large download


Rockybilly:
Greetings,

It would appear currently, the response body is returned using a std::string but no reserving space method is provided to the calling application.

If a large download is requested, this would cause many unnecessary reallocs and memmove's, that would decrease the performance of this program.

It is true that in stuff like chunked encoding, the size cannot be known before the download, but there should be an option for this if the caller does not have any memory issues and just wants performance.

    std::string response_string;
    std::string header_string;
    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, cpr::util::writeFunction);
    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response_string);
    curl_easy_setopt(curl, CURLOPT_HEADERDATA, &header_string);

If I am not mistaken, this is where the stuff happens. It would be beneficial, if the response_string was reservable.

COM8:
Yes, this would be a nice addition.
Would you like to create a PR for this?

jsouquie:
Following the merge of PR #533, it is now easy to implement this feature.

As we discussed previously in #533, it would be exposed through a method of the Session class. Something like `void Session::ReserveResponseText(size_t capacity)`.

One additional decision to make: should the user be required to call this method before each request, or should the capacity  be "sticky" and apply to all requests made using this session (this implies that the capacity has to be stored as a member of Session) ? I have a slight preference for making it not sticky.

Please tell me if that design and naming is OK with you, and I'll create the PR.

```cpp
cpr::Session session;
for (  ... )
{
    session.SetUrl(url);
    session.ReserveResponseText(256*1024); // not sticky: must be called for each iteration
    std::string result = session.Get().text;
}
```

COM8:
@jsouquie I also agree with you. Making it sticky, does make sense.
Things we also have to think about is, what happens in case we receive more than the reserved amount of data?
I think we should handle this by increasing the buffer (or let it `std::string` handle if we use `std::string::reserve`).
Never the less we should print something to `std::err` in case this happens and `cpr::Debug` is enabled.

Besides that `ReserveResponseText` sounds good to me.

Rockybilly:
[...]

PR #533, opened 2021-03-28
Split a request preparation from its execution

jsouquie:
I'm currently prototyping the use of cpr with multiple parallel requests in a single thread, through the curl_multi_socket_action() API.
This means I need to decouple the request execution (curl_easy_perform) from its preparation and completion.
Would you consider upstreaming the included changes that allow this decoupling ?

Notes for further:
* having response_string_ as an attribute of Session should allow a trivial implementation of #530.
* there is currently a lot of similarity between makeRequest() and makeDownloadRequest(). I'm not sure if the few differences are intentional or a result of copy-paste gone wrong. The prepareCommon() method I introduced would be a good way to reduce copy-paste decay if the differences were not intentional.


jsouquie:
I forgot to comment about how these new API are used. It look something like this:
```
cpr::Url url = cpr::Url{"http://www.httpbin.org/get"};
cpr::Parameters parameters = cpr::Parameters{{"hello", "world"}};
cpr::Session session;
session.SetUrl(url);
session.SetParameters(parameters);

session.PrepareGet();
CURLcode cc = my_curl_multi_engine(session.GetCurlHolder()->handle);
cpr::Response r = session.Complete( cc )

std::cout << r.url << std::endl;
```

COM8 (changes requested):
Thanks for contributing.
I really like your changes here.
Only a few minor things:
* Could you please create a PR against the `gh-pages` branch and add documentation for your feature (chapter [`Session Object`](https://whoshuu.github.io/cpr/advanced-usage.html#session-objects)).
* Add support and documentation for #530.

COM8:
* And add a couple of test cases so we can make sure everything works like expected if somebody uses only the new API.

jsouquie:
I'm thinking about #530: as I said, the implementation is trivial after this PR, but I'm unsure about how to expose this capability.
Adding a `SetResponseCapacity(const ResponseCapacity& capacity)` and a matching `SetOption()` seems a bit off to me. Mainly because all other options are directly related to curl itself, whereas this one is related to cpr only.

COM8 (approved):
Looks good. Thanks.

COM8:
> 
> 
> I'm thinking about #530: as I said, the implementation is trivial after this PR, but I'm unsure about how to expose this capability.
> Adding a `SetResponseCapacity(const ResponseCapacity& capacity)` and a matching `SetOption()` seems a bit off to me. Mainly because all other options are directly related to curl itself, whereas this one is related to cpr only.

I agree something like `SetResponseCapacity(const ResponseCapacity& capacity)` does not make sense here. I'd say this should be something that can only be set in a `cpr::Session` object directly.
So I suggest we push this back and ignore it in this PR.

COM8:
Great work.

PR #537, opened 2021-04-02
Fix subproject check

KingKili:


COM8 (approved):

KingKili:
I don't think this is my fault 🙀

PR #544, opened 2021-04-11
Make Session objets movable for users

jsouquie:
Session's move constructor and move assignment operator were not usable from the public API because they were defined inline in the header. This was causing an "use of undefined type Impl" from unique_ptr.


COM8 (approved):
Well done. Thanks.

PR #549, opened 2021-04-18
Added darwin-ssl

tallalnparis4ev:
Hello,

I added darwin-ssl to the CMake file so that Mac users don't have to pass in -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl every time. I'm not sure if I did this the best way, but it works.


COM8 (changes requested):
Nice addition. Thanks.
Only one thing is missing.
Could you please add a `CPR_FORCE_DARWINSSL_BACKEND` option to make it consistent with the `CPR_FORCE_WINSSL_BACKEND` and `CPR_FORCE_OPENSSL_BACKEND` options.

COM8:
I added two new CI runs for you, so you can validate your options

tallalnparis4ev:
No worries, thanks for investigating.

I got a question though, look at the error message in the following code snippet:
elseif(CPR_FORCE_WINSSL_BACKEND)
            if(WIN32)
                message(STATUS "Using WinSSL.")
                set(SSL_BACKEND_USED "WinSSL")
            else()
                message(FATAL_ERROR "CPR_FORCE_OPENSSL_BACKEND enabled but we were not able to find OpenSSL!")
            endif()

Should it not be something like "FATAL_ERROR "CPR_FORCE_WINSSL_BACKEND enabled on non-Windows machine!"?

Or is my understanding of the force option wrong?

COM8:
> No worries, thanks for investigating.
> 
> I got a question though, look at the error message in the following code snippet:
> elseif(CPR_FORCE_WINSSL_BACKEND)
> if(WIN32)
> message(STATUS "Using WinSSL.")
> set(SSL_BACKEND_USED "WinSSL")
> else()
> message(FATAL_ERROR "CPR_FORCE_OPENSSL_BACKEND enabled but we were not able to find OpenSSL!")
> endif()
> 
> Should it not be something like "FATAL_ERROR "CPR_FORCE_WINSSL_BACKEND enabled on non-Windows machine!"?
> 
> Or is my understanding of the force option wrong?

Yes, this is wrong. You are right.

tallalnparis4ev:
> > No worries, thanks for investigating.
> > I got a question though, look at the error message in the following code snippet:
> > elseif(CPR_FORCE_WINSSL_BACKEND)
> > if(WIN32)
> > message(STATUS "Using WinSSL.")
> > set(SSL_BACKEND_USED "WinSSL")
> > else()
> > message(FATAL_ERROR "CPR_FORCE_OPENSSL_BACKEND enabled but we were not able to find OpenSSL!")
> > endif()
> > Should it not be something like "FATAL_ERROR "CPR_FORCE_WINSSL_BACKEND enabled on non-Windows machine!"?
> > Or is my understanding of the force option wrong?
> 
> Yes, this is wrong. You are right.

Ah, that conditional can never be reached tho, this happens beforehand :P
if(CPR_FORCE_WINSSL_BACKEND AND (NOT WIN32))
        message(FATAL_ERROR "WinSSL is only available on Windows! Use OpenSSL instead (CPR_FORCE_OPENSSL_BACKEND).")
    endif()

tallalnparis4ev:
Sorry, forgot to pull ur changes :/

I removed the repeated check for if we are on Windows and added the feature for forcing DarwinSSL, I hope.

ollien (approved):
LGTM! Builds and works.

COM8 (approved):
Looks good. Thanks!

PR #550, opened 2021-04-19
support a  general SetOption API :  SetOption(int option, void* value)

byronhe:
Some times，we must set some rare curl options ， like：

```
CURLOPT_OPENSOCKETFUNCTION
CURLOPT_SOCKOPTFUNCTION 
CURLOPT_DNS_CACHE_TIMEOUT
```

it would be nice to have a general API to do such things.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -63,6 +63,7 @@ class Session::Impl {
     void SetUnixSocket(const UnixSocket& unix_socket);
     void SetVerbose(const Verbose& verbose);
     void SetSslOptions(const SslOptions& options);
+    int SetOption(int option, void* value);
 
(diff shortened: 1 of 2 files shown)


COM8:
Thanks for contributing!
In theory this is a great addition, but something like this is already possible by using`session.GetCurlHolder()`.
An example can be found here:
https://github.com/whoshuu/cpr/blob/63bfcc6292bba06f9806994ee5474baadff48254/test/session_tests.cpp#L877-L899

So I don't think we should expose two interfaces to the user to archive the same result.
What are you thoughts on this?

byronhe:
OK, `session.GetCurlHolder()` is great, that is exactly what i need.

PR #558, opened 2021-04-24
Fix MSVC build; linker flags for sanitizer builds

ollien:
When adding the latest cpr to Conan, we ran into a problem where we were unable to build when using a `shared` build. It seems that some versions of MSVC are picky about having linker flags defined, and we'd get this in the output log.

```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_SHARED_LINKER_FLAGS_THREADSAN
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_SHARED_LINKER_FLAGS_ADDRSAN
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_SHARED_LINKER_FLAGS_ALLSAN
```

The logs of this CI run can be found here https://github.com/conan-io/conan-center-index/pull/5194#issuecomment-822038253, and that whole thread contains discussion around the issue. If I'm honest, I'm not fully familiar with MSVC. or at least, not enough to fully understand why this is the case. Anyway, we found that adding linker flags (and just sharing the ones from the debug build) seemed to do the trick.

### DIFF
--- a/cmake/sanitizer.cmake
+++ b/cmake/sanitizer.cmake
@@ -16,6 +16,7 @@ if(THREAD_SANITIZER_AVAILABLE)
 endif()
 set(CMAKE_C_FLAGS_THREADSAN "${CMAKE_C_FLAGS_DEBUG} ${THREAD_SAN_FLAGS}" CACHE INTERNAL "Flags used by the C compiler during thread sanitizer builds." FORCE)
 set(CMAKE_CXX_FLAGS_THREADSAN "${CMAKE_CXX_FLAGS_DEBUG} ${THREAD_SAN_FLAGS}" CACHE INTERNAL "Flags used by the C++ compiler during thread sanitizer builds." FORCE)
+set(CMAKE_SHARED_LINKER_FLAGS_THREADSAN "${CMAKE_SHARED_LINKER_FLAGS_DEBUG}" "Flags used for the linker during thread sanitizer builds" FORCE)
 
 # Address sanitizer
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Nice. Thanks!
I will prepare a 1.6.1 release, so you can continue updating the Conan package.

COM8:
I think something broke. Could you please fix sanitizers for linux then I will create a 1.6.2 release of cpr.

ollien:
```
 [ 62%] Building C object _deps/curl-build/lib/CMakeFiles/libcurl.dir/vauth/cleartext.c.o
  g++: error: ;Flags: No such file or directory
  g++: error: used: No such file or directory
  g++: error: for: No such file or directory
  g++: error: the: No such file or directory
  g++: error: linker: No such file or directory
  g++: error: during: No such file or directory
  g++: error: most: No such file or directory
  g++: error: possible: No such file or directory
  g++: error: sanitizer: No such file or directory
  g++: error: builds;FORCE: No such file or directory
  ```
  
  :sweat_smile: That's interesting. Will take a look

ollien:
Builds locally for me with this change. Looks like I missed `CACHE INTERNAL` oops. Looks like it should work... feel free to kick the pipeline when you get a chance.

COM8:
Thanks for your help!

COM8:
https://github.com/whoshuu/cpr/releases/tag/1.6.2

PR #559, opened 2021-04-25
Fixed session copy/move bug i.e added desctructor in Session::Impl class

EMHussain:
issue [553](https://github.com/whoshuu/cpr/issues/553) fixed.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -24,6 +24,7 @@ constexpr long OFF = 0L;
 class Session::Impl {
   public:
     Impl();
+    ~Impl();
 
     void SetUrl(const Url& url);
     void SetParameters(const Parameters& parameters);
@@ -135,6 +136,7 @@ Session::Impl::Impl() : curl_(new CurlHolder()) {
 #endif
 #endif
 }
+Session::Impl::~Impl() {}
 
 void Session::Impl::SetUrl(const Url& url) {
     url_ = url;


COM8 (approved):
Great. Thanks!
Could you pleas and one test case for this, so w we check in the future for this?

cpr/session.cpp
```diff
@@ -24,6 +24,7 @@ constexpr long OFF = 0L;
 class Session::Impl {
   public:
     Impl();
+    ~Impl();
```
COM8, line 27: Any specific reason why `~Impl() = default;` is not working here?

EMHussain: I did not try `~Impl() = default;` i think  would be fine. 

- to fix this bug properly `Impl` class required a custom deleter required.

**reason**:std::unique_ptr The size of the type needs to be statically detected in static_assert(sizeof(Impl)>0, But our Impl Is a pre declared type.  It's impossible to calculate , So it leads to an error.
I create a commit to fix.

COM8:
I'm unable to reproduce this, since the following works for me:
**CXX 11**
```c++
std::unique_ptr<cpr::Session> session(new cpr::Session);
```

**CXX 14**
```c++
std::unique_ptr<cpr::Session> session = std::make_unique<cpr::Session>();
```

So could you please provide a test case to reproduce this behavior?

COM8:
Closing this for inactivity.
Feel free to reopen in case this issue still persists.

PR #562, opened 2021-04-26
Clang tidy

COM8:
Adds support for enabling clang-tidy and adds a separate CI run for it.


COM8:
@KingKili can I get a review from you on this one?

KingKili (changes requested):
I'd prefer such things sorted alphabetically, it drastically improves readability.

CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ cpr_option(CPR_ENABLE_SSL "Enables or disables the SSL backend. Required to perf
 cpr_option(CPR_FORCE_OPENSSL_BACKEND "Force to use the OpenSSL backend. If CPR_FORCE_OPENSSL_BACKEND, CPR_FORCE_DARWINSSL_BACKEND, and CPR_FORCE_WINSSL_BACKEND are set to to OFF, cpr will try to automatically detect the best available SSL backend (WinSSL - Windows, OpenSSL - Linux, DarwinSSL - Mac ...)." OFF)
 cpr_option(CPR_FORCE_WINSSL_BACKEND "Force to use the WinSSL backend. If CPR_FORCE_OPENSSL_BACKEND, CPR_FORCE_DARWINSSL_BACKEND, and CPR_FORCE_WINSSL_BACKEND are set to to OFF, cpr will try to automatically detect the best available SSL backend (WinSSL - Windows, OpenSSL - Linux, DarwinSSL - Mac ...)." OFF)
 cpr_option(CPR_FORCE_DARWINSSL_BACKEND "Force to use the DarwinSSL backend. If CPR_FORCE_OPENSSL_BACKEND, CPR_FORCE_DARWINSSL_BACKEND, and CPR_FORCE_WINSSL_BACKEND are set to to OFF, cpr will try to automatically detect the best available SSL backend (WinSSL - Windows, OpenSSL - Linux, DarwinSSL - Mac ...)." OFF)
+cpr_option(CPR_ENABLE_LINTING "Set to ON to enable clang linting. If enabled, CPR_STATIC_ANALYZE has to be disabled." OFF)
+cpr_option(CPR_STATIC_ANALYZE "Set to ON to enable the GCC 10 static analysis. If enabled, CPR_ENABLE_LINTING has to be disabled." OFF)
```
KingKili, line 43: I think GCC static analysis should be a separate PR

CMakeLists.txt
```diff
@@ -47,7 +49,20 @@ include(GNUInstallDirs)
 include(FetchContent)
 include(cmake/code_coverage.cmake)
 include(cmake/sanitizer.cmake)
-include(cmake/gcc_analyze.cmake)
+include(cmake/clear_variable.cmake)
+
+# Linting and Static Analyze
+if(CPR_ENABLE_LINTING AND CPR_STATIC_ANALYZE)
+    message(FATAL_ERROR "Linting and the GCC static analysis can not be enabled at the same time! Disable either CPR_STATIC_ANALYZE or CPR_ENABLE_LINTING.")
+endif()
```
KingKili, line 57: I think CPR should be fine with clang-tidy and gcc-analyze running in parallel. (It is only a RAM problem for bigger Projects)

CMakeLists.txt
```diff
@@ -47,7 +49,20 @@ include(GNUInstallDirs)
...
+if(CPR_ENABLE_LINTING)
+    include(cmake/clang-tidy.cmake)
+endif()
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Werror")
+endif()
```
KingKili, line 65: This has nothing todo with clang-tidy. Either remove it or do a separate PR

CMakeLists.txt
```diff
@@ -270,13 +299,17 @@ if(CPR_BUILD_TESTS)
...
 add_subdirectory(cpr)
 add_subdirectory(include)
 
 if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND CPR_BUILD_TESTS)
+    # Disable linting for test since they are currently not up to the standard
+    clear_variable(DESTINATION CMAKE_CXX_CLANG_TIDY BACKUP CMAKE_CXX_CLANG_TIDY_BKP)
     enable_testing()
     add_subdirectory(test)
+    restore_variable(DESTINATION CMAKE_CXX_CLANG_TIDY BACKUP CMAKE_CXX_CLANG_TIDY_BKP)
 endif()
```
KingKili, line 315: I think we should also disable gcc-analyze for subprojects

cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,13 @@
+if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+    find_program(CLANG_TIDY_EXECUTABLE NAMES clang-tidy-7 clang-tidy)
```
KingKili, line 2: clang-tidy-7 is 5 mayor releases old, we should either search also for clang-tidy-{0 ... 12}, or remove it completely

cpr/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 cmake_minimum_required(VERSION 3.15)
 
+if(CPR_STATIC_ANALYZE)
+    include(../cmake/gcc_analyze.cmake)
```
KingKili, line 4: Using .. is always ugly, we could instead also work with `CMAKE_MODULE_PATH`.

cmake/clang-tidy.cmake
```diff
@@ -0,0 +1,13 @@
...
+    mark_as_advanced(CLANG_TIDY_EXECUTABLE)
+
+    if (${CLANG_TIDY_EXECUTABLE})
+        message(FATAL_ERROR "Clang-tidy not found")
+    else()
+        message(STATUS "Enabling clang-tidy")
+        set(CMAKE_CXX_CLANG_TIDY "${CLANG_TIDY_EXECUTABLE};-warnings-as-errors=*")
```
KingKili, line 9: Using warning-as-errors=* is a very strict policy, one could also reason to omit it (or do only specific checks).

KingKili (changes requested):
Overall a decent start (the cmake code somehow look very familiar to me). But I have added a few comments with suggestions for change.

cmake/clang-tidy.cmake
COM8: Done.

cmake/clang-tidy.cmake
COM8: I would keep it enabled that way, since linting is for development on cpr. There we than can deal with those.

PR #569, opened 2021-05-20
Fix shadowed variable warnings

craigscott-crascit:
I've tried to follow the existing naming convention already used by some files where some instances of this problem have been fixed previously.


COM8 (approved):
Great work. Thanks!

COM8:
The CI failing with `error: no member named 'sleep_for' in namespace 'std::this_thread'` is already fixed in #562.

Issue #567, opened 2021-05-14
Don't force setting FETCHCONTENT_QUIET, it's a user control


craigscott-crascit:
The top level `CMakeLists.txt` file currently contains these lines:

```cmake
    # Show progress of FetchContent:
    set(FETCHCONTENT_QUIET OFF CACHE INTERNAL "" FORCE)
    FetchContent_Declare(curl
                         URL                    https://github.com/curl/curl/releases/download/curl-7_75_0/curl-7.75.0.tar.xz
                         URL_HASH               SHA256=fe0c49d8468249000bda75bcfdf9e30ff7e9a86d35f1a21f428d79c389d55675 # the file hash for curl-7.75.0.tar.xz
                         USES_TERMINAL_DOWNLOAD TRUE)   # <---- This is needed only for Ninja to show download progress
    FetchContent_MakeAvailable(curl)
```

The `FETCHCONTENT_QUIET` cache variable is meant as a user control, projects should not modify it. If you want to enforce QUIET behavior for this particular dependency, you can set it as a normal non-cache variable here and it will apply to the `FetchContent_MakeAvailable(curl)` call without affecting other parent or sibling scopes.

At the moment, if another project adds `cpr` as a dependency using `FetchContent`, they have to undo the change to the `FETCHCONTENT_QUIET` cache variable to honor any user setting (most such projects probably won't go to that effort, so users will suffer).

craigscott-crascit:
Same thing in `cmake/zlib_external.cmake` as well.

COM8:
You are right!
Would you like to create a PR for this?

PR #570, opened 2021-05-20
Don't forcibly override user setting for FETCHCONTENT_QUIET

craigscott-crascit:
By forcing FETCHCONTENT_QUIET to be OFF, it means any project that
consumes cpr will end up with all of its dependencies having non-quiet
output. This variable is meant to be a user control, projects should not
set it.

Fixes: #567


COM8 (approved):
Thanks!

PR #573, opened 2021-05-27
Proxy authentication

### LINKED ISSUE #248: support Proxy Authentication
Currently, it looks like Proxy Authentication is not supported, it would be nice if someone added code to add support for it

https://curl.haxx.se/libcurl/c/CURLOPT_PROXYAUTH.html

curl setting: CURLOPT_PROXYUSERPWD

jaskinn:
This is another attempt to resolve #248. It is possible to have different authentication credentials per protocol. It follows the same pattern as `Proxies`.

I'm open to suggestions on what tests to add.


include/cpr/proxyauth.h
```diff
@@ -0,0 +1,14 @@
...
+
+namespace cpr {
+class ProxyAuth : public Authentication {
+  public:
+    ProxyAuth(const std::string& username, const std::string& password) : Authentication{username, password} {}
+};
+
+} // namespace cpr
+
+#endif
```
COM8, line 14: Just a simple new line missing

```suggestion
#endif

```

COM8 (changes requested):
Well done. Thanks!
Just a single new line missing.

Regarding the tests:
Are you aware of a simple way to test this with perhaps an easy proxy setup?
I know there are a couple of C++ proxy implementations we could ship with our unit tests.
But I'm not a fan of introducing dependencies to those. Even if it's only when running tests.

COM8:
Besides that, please create a PR for the gh-pages repo, for a bit of documentation.

jaskinn:
I'm unaware of a proxy implementation we could use. I had a quick search and only found examples that appeared to require quite a bit of work.

jaskinn:
I have changed this PR to be more like the original PR (copied the existing proxy map implementation to enable a proxy per protocol/scheme).

This does make it a bit "wordy" for the library user, so if there are strong opinions, happy to change.

jaskinn:
I noticed that the cURL doco for [CURLOPT_PROXYUSERPWD ](https://curl.se/libcurl/c/CURLOPT_PROXYUSERPWD.html )stated that it needed to be URL encoded. I had never noticed that before, but it might explain why I have had some seemingly random failures out in the wild.

jaskinn:
> 
> 
> Besides that, please create a PR for the gh-pages repo, for a bit of documentation.

Working on that now.

jaskinn:
> 
> 
> Well done. Thanks!
> Just a single new line missing.
> 
> Regarding the tests:
> Are you aware of a simple way to test this with perhaps an easy proxy setup?
> I know there are a couple of C++ proxy implementations we could ship with our unit tests.
> But I'm not a fan of introducing dependencies to those. Even if it's only when running tests.

For this PR, I manually tested via squid. Not really appropriate but I feel better that it works as expected.

test/proxy_auth_tests.cpp
```diff
@@ -0,0 +1,98 @@
...
+#define HTTPS_PROXY "192.168.0.180:3128"
+#define PROXY_USER "u$er"
+#define PROXY_PASS "p@ss"
+
+using namespace cpr;
+
+// TODO: These should be fixed after a source code implementation of a proxy
+#if defined(false)
+TEST(ProxyAuthTests, SingleProxyTest) {
+    Url url{"https://www.httpbin.org/get"};
```
COM8, line 20: Please do not use external URLs for tests here.
Since those will fail in case we run those test cases offline.

Please use the `HttpServer` for this.
You can find examples for this inside the `get_tests.cpp` file.

COM8 (changes requested):
I'm fine with disabling the tests for now.
Looks good, except the URL used.

COM8 (approved):

COM8:
Now MacOS is unhappy:
```
'CFURLCreateDataAndPropertiesFromResource' is deprecated: first deprecated in macOS 10.9 - For resource data, use the CFReadStream API. For file resource properties, use CFURLCopyResourcePropertiesForKeys. [-Wdeprecated-declarations]
```

jaskinn:
Hopefully, I haven't missed anything else. Ready to run tests again.

COM8 (approved):

COM8:
Now we are ready to go!

PR #576, opened 2021-06-03
Documentation for ProxyAuthentication #248

jaskinn:
Doco is so hard - this took way longer than I had hoped. :)


COM8 (approved):
Looks good. Thanks!

PR #597, opened 2021-07-22
add userdata param to callback function

zydxhs:
Add UserData to the callback function to satisfy concurrent calls in a multi-threaded environment.


include/cpr/callback.h
```diff
@@ -12,44 +12,61 @@ class ReadCallback {
...
+    ReadCallback(cpr_off_t p_size, std::function<bool(intptr_t userdata, char* buffer, size_t& size)> p_callback, intptr_t p_userdata = 0)
+            : userdata(p_userdata), size{p_size}, callback{std::move(p_callback)} {}
+    bool operator()(char* buffer, size_t& size) const {
+        return callback(userdata, buffer, size);
+    }
 
+    intptr_t userdata;
     cpr_off_t size{};
-    std::function<bool(char* buffer, size_t& size)> callback;
+    std::function<bool(intptr_t userdata, char* buffer, size_t& size)> callback;
```
COM8, line 25: Since `userdata` is an optional argument, we should move it to the last position everywhere to keep it consistent.

COM8 (changes requested):
Nice work!
Just change the order and we are ready to go.

Also please create a PR for the `gh-pages` branche with an update for the documentation:
https://whoshuu.github.io/cpr/advanced-usage.html#setting-callbacks

COM8:
The output from the MacOS CI run.

```
[ RUN      ] CallbackPatchTests.CallbackPatchLambdaTextTest
  10: /Users/runner/work/cpr/cpr/test/callback_tests.cpp:658: Failure
  10: Expected equality of these values:
  10:   future.wait_for(zero)
  10:     Which is: 4-byte object <01-00 00-00>
  10:   std::future_status::ready
  10:     Which is: 4-byte object <00-00 00-00>
```

zydxhs:
@COM8 What do I need to do to solve the problem?

1.  documentation
2. `callback.h` conflicts

include/cpr/callback.h
zydxhs: The userdata parameter, which must appear in the callback function, so, I don't think it's a problem right.

COM8:
The definition of `ProgressCallback` changed in https://github.com/whoshuu/cpr/commit/cfb2a3e42e10e5a1a6e16de22504740b6ab1c6ad. Please try to rebase your changes.
https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a

To update the docs, please checkout the `gh-pages` banche, perform your changes and create an additional PR.

include/cpr/callback.h
COM8: Yes, but please change the order, so `userdata` is always the last parameter.

zydxhs: All right

zydxhs:
Now I have changed userdata to the last, and merged the "whoshuu/master" with  cfb2a3e.

zydxhs:
I reintroduced a PR https://github.com/whoshuu/cpr/pull/601

PR #599, opened 2021-07-27
add interface GetDownloadFileLength

zydxhs:
Some HTTP servers don't support the HEAD method to get the length of the file, so I added the GetDownloadFileLength interface to get the length of the file before the actual download starts. 

Unnecessary memory allocation can be avoided in some scenarios.

example :
``` c++
void *global_buff = nullptr;
int64_t global_down_length = 0;

bool write_data(std::string data)
{
    memcpy(global_buff + global_down_length, data.data(), data.size());
    global_down_length += data.size();
}

int main() {
    cpr::Session session;
    session.SetUrl(cpr::Url{"http://xxx/file"});
    // memory alloc
    global_buff = malloc(session.GetDownloadFileLength());
    session.Download(cpr::WriteCallback{write_data});
    // do something
    free(global_buff);
    global_buff = nullptr;
}
```

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -66,7 +66,7 @@ class Session::Impl {
     void SetVerbose(const Verbose& verbose);
     void SetSslOptions(const SslOptions& options);
 
-    int64_t GetDownloadFileLength();
+    cpr_off_t GetDownloadFileLength();
     Response Delete();
     Response Download(const WriteCallback& write);
     Response Download(std::ofstream& file);
@@ -233,16 +233,14 @@ void Session::Impl::SetUserAgent(const UserAgent& ua) {
 void Session::Impl::SetPayload(Payload&& payload) {
     hasBodyOrPayload_ = true;
     const std::string content = payload.GetContent(*curl_);
-    curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDSIZE_LARGE,
-                     static_cast<curl_off_t>(content.length()));
+    curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDSIZE_LARGE, static_cast<curl_off_t>(content.length()));
     curl_easy_setopt(curl_->handle, CURLOPT_COPYPOSTFIELDS, content.c_str());
 }
 
 void Session::Impl::SetPayload(const Payload& payload) {
     hasBodyOrPayload_ = true;
     const std::string content = payload.GetContent(*curl_);
-    curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDSIZE_LARGE,
-                     static_cast<curl_off_t>(content.length()));
+    curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDSIZE_LARGE, static_cast<curl_off_t>(content.length()));
     curl_easy_setopt(curl_->handle, CURLOPT_COPYPOSTFIELDS, content.c_str());
 }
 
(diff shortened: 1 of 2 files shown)


cpr/session.cpp
```diff
@@ -66,6 +66,7 @@ class Session::Impl {
     void SetVerbose(const Verbose& verbose);
     void SetSslOptions(const SslOptions& options);
 
+    int64_t GetDownloadFileLength();
```
COM8, line 69:
```suggestion
    cpr_off_t GetDownloadFileLength();
```

cpr/session.cpp
```diff
@@ -493,6 +494,30 @@ void Session::Impl::PrepareDelete() {
     prepareCommon();
 }
 
+int64_t Session::Impl::GetDownloadFileLength() {
```
COM8, line 497:
```suggestion
cpr_off_t Session::Impl::GetDownloadFileLength() {
```

cpr/session.cpp
```diff
@@ -493,6 +494,30 @@ void Session::Impl::PrepareDelete() {
...
+	curl_off_t downloadFileLenth = -1;
```
COM8, line 498:
```suggestion
	cpr_off_t downloadFileLenth = -1;
```

cpr/session.cpp
```diff
@@ -802,6 +827,7 @@ void Session::SetOption(const Verbose& verbose) { pimpl_->SetVerbose(verbose); }
 void Session::SetOption(const UnixSocket& unix_socket) { pimpl_->SetUnixSocket(unix_socket); }
 void Session::SetOption(const SslOptions& options) { pimpl_->SetSslOptions(options); }
 
+int64_t Session::GetDownloadFileLength() { return pimpl_->GetDownloadFileLength(); }
```
COM8, line 830:
```suggestion
cpr_off_t Session::GetDownloadFileLength() { return pimpl_->GetDownloadFileLength(); }
```

include/cpr/session.h
```diff
@@ -119,6 +119,7 @@ class Session {
     void SetOption(const UnixSocket& unix_socket);
     void SetOption(const SslOptions& options);
 
+    int64_t GetDownloadFileLength();
```
COM8, line 122:
```suggestion
    cpr_off_t GetDownloadFileLength();
```

COM8 (changes requested):
Good idea!
Thanks.
Please replace `int64_t` with `cpr_off_t`, which is defined here:
https://github.com/whoshuu/cpr/blob/cfb2a3e42e10e5a1a6e16de22504740b6ab1c6ad/include/cpr/cprtypes.h#L13-L16

It is a simple wrapper for `curl_off_t` to prevent applications from having to link against libcurl.

Besides that, please create an additional PR for documentation against the `gh-pages` branche.

zydxhs:
Now I have changed int64_t/curl_off_t to cpr_off_t.

COM8:
`ubuntu-18-gcc-shared-ssl-system-curl` and `ubuntu-18-gcc-static-ssl-system-curl` are currently expected to fail.

COM8 (approved):

zydxhs:
> `ubuntu-18-gcc-shared-ssl-system-curl` and `ubuntu-18-gcc-static-ssl-system-curl` are currently expected to fail.

@COM8 Please tell me, what should I do?

COM8:
> > `ubuntu-18-gcc-shared-ssl-system-curl` and `ubuntu-18-gcc-static-ssl-system-curl` are currently expected to fail.
> 
> @COM8 Please tell me, what should I do?

Nothing. They are broken. It should fix it self.

> 
> > Besides that, please create an additional PR for documentation against the gh-pages branche.
> 
> I didn't find the documentation for the Download interface in gh-pages, how should I add it?

Please create a `DOWNLOAD Requests` section next to the [`PUT and PATCH Requests`](https://whoshuu.github.io/cpr/advanced-usage.html#put-and-patch-requests) section. Add a small download example and then in a sub-section add your example and documentation.

PR #612, opened 2021-08-17
Add PIC option

zydxhs:
Using the option CPR_USE_PIC, add the option '-fpic' to GCC when generating.so/.a.

This allows CPR to be used in .so .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ cpr_option(CPR_FORCE_DARWINSSL_BACKEND "Force to use the DarwinSSL backend. If C
 cpr_option(CPR_ENABLE_LINTING "Set to ON to enable clang linting." OFF)
 cpr_option(CPR_BUILD_TESTS "Set to ON to build cpr tests." ON)
 cpr_option(CPR_BUILD_TESTS_SSL "Set to ON to build cpr ssl tests" ${CPR_BUILD_TESTS})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -60,6 +61,11 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Werror")
 endif()
 
+if(CPR_USE_PIC)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
COM8, line 65: Any reason, why the `CMAKE_POSITION_INDEPENDENT_CODE` is not being used here?

https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html

COM8 (changes requested):
Thanks for creating this PR.

CMakeLists.txt
zydxhs: Excuse me for not knowing this variable.

I'll change it later.

COM8 (approved):
Thanks!

COM8:
Hmmm...
I agree @Leon0402 .

@zydxhs do you have any specific use case, why you need this option?

COM8:
Since no explanation why this option is required has been provided, I'm reverting this PR.

Issue #300, opened 2018-11-30
System provided proxy is ignored by setting CURLOPT_PROXY to "" when no proxy is set


EliSchleifer:
By default the cpr will ignore system proxy if it is not specified explicitly this is because setting CURLOPT_PROXY to "" disabled default proxy for system

COM8:
Would you like to create a PR for this?

PR #617, opened 2021-08-22
Respect system proxy configurations

SpriteOvO:
Issue: #300

Tested `cpr::Download` on `Windows` only and works fine for me.


COM8 (approved):
```
Setting the proxy string to "" (an empty string) will explicitly disable the use of a proxy, even if there is an environment variable set for it.
```
Taken from the curl docs: https://curl.se/libcurl/c/CURLOPT_PROXY.html

Thanks for fixing this.

PR #637, opened 2021-09-15
Code of Conduct

COM8:
This PR adds a Code of Conduct.
This issue will be merged on October 15. 2021 in case there are no problems with it.

Let me know if there is something you think that's wrong or missing here.


whoshuu:
:+1:

PR #641, opened 2021-09-20
Added exception throwing for bad HTTP code in response

sixdenip:
Added a class method to Response that throws when the status_code is a 4xx or 5xx.

### DIFF
--- a/cpr/response.cpp
+++ b/cpr/response.cpp
@@ -43,4 +43,17 @@ std::vector<std::string> Response::GetCertInfo() {
 
     return info;
 }
+
+void Response::ThrowForStatus(){
+    std::string err_msg = "";
+    if (400 <= status_code && status_code < 500) {
+        err_msg = "Client error : " + std::to_string(status_code) + " with content : " + text;
+    } else if (500 <= status_code && status_code < 600) {
+        err_msg = "Server error : " + std::to_string(status_code) + " with content : " + text;
+    }
+
+    if (err_msg.size()>0)
+        throw cpr::Exception::BadResponse(err_msg, status_code);
+}
+
 } // namespace cpr
--- a/include/cpr/cpr.h
+++ b/include/cpr/cpr.h
@@ -4,6 +4,7 @@
 #include "cpr/api.h"
 #include "cpr/auth.h"
 #include "cpr/cprtypes.h"
+#include "cpr/exceptions.h"
 #include "cpr/interface.h"
 #include "cpr/redirect.h"
 #include "cpr/response.h"
--- a/include/cpr/response.h
+++ b/include/cpr/response.h
@@ -8,10 +8,12 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <string>
 
 #include "cpr/cookies.h"
 #include "cpr/cprtypes.h"
 #include "cpr/error.h"
+#include "cpr/exceptions.h"
 #include "cpr/ssl_options.h"
 #include "cpr/util.h"
 
@@ -44,6 +46,7 @@ class Response {
     Response(std::shared_ptr<CurlHolder> curl, std::string&& p_text, std::string&& p_header_string,
              Cookies&& p_cookies, Error&& p_error);
     std::vector<std::string> GetCertInfo();
+    void ThrowForStatus();
     Response(const Response& other) = default;
     Response(Response&& old) noexcept = default;
     ~Response() noexcept = default;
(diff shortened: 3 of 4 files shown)


COM8:
Looks good so far. Thanks.
The following is still missing:
* The actual exception calls
* Testcases
* Documentation. For this please create a PR here: https://github.com/libcpr/docs

cpr/response.cpp
```diff
@@ -43,4 +43,17 @@ std::vector<std::string> Response::GetCertInfo() {
 
     return info;
 }
+
+void Response::ThrowForStatus(){
+    std::string err_msg = "";
```
COM8, line 48:
```suggestion
    std::string err_msg;
```

cpr/response.cpp
```diff
@@ -43,4 +43,17 @@ std::vector<std::string> Response::GetCertInfo() {
...
+    if (400 <= status_code && status_code < 500) {
+        err_msg = "Client error : " + std::to_string(status_code) + " with content : " + text;
+    } else if (500 <= status_code && status_code < 600) {
+        err_msg = "Server error : " + std::to_string(status_code) + " with content : " + text;
+    }
+
+    if (err_msg.size()>0)
```
COM8, line 55:
```suggestion
    if (!err_msg.empty())
```

cpr/response.cpp
```diff
@@ -43,4 +43,17 @@ std::vector<std::string> Response::GetCertInfo() {
...
+        throw cpr::Exception::BadResponse(err_msg, status_code);
```
COM8, line 56:
```suggestion
        throw cpr::Exception::BadResponse(std::move(err_msg), status_code);
```

include/cpr/exceptions.h
```diff
@@ -0,0 +1,19 @@
...
+#include <exception>
+#include <string>
+
+namespace cpr {
+    namespace Exception {
+        class BadResponse : public std::exception {
+            public:
+                int http_code;
+                std::string message;
+                BadResponse(const std::string& msg, int http_code) : http_code(http_code), message(msg) {}
```
COM8, line 12:
```suggestion
                BadResponse(std::string&& msg, int http_code) : http_code(http_code), message(std::move(msg)) {}
```

include/cpr/exceptions.h
```diff
@@ -0,0 +1,19 @@
+#ifndef CPR_CPR_EXCEPTIONS_H
+#define CPR_CPR_EXCEPTIONS_H
+#include <exception>
+#include <string>
+
+namespace cpr {
+    namespace Exception {
+        class BadResponse : public std::exception {
+            public:
+                int http_code;
```
COM8, line 10:
```suggestion
                const int http_code;
```

include/cpr/exceptions.h
```diff
@@ -0,0 +1,19 @@
...
+                std::string message;
```
COM8, line 11:
```suggestion
                const std::string message;
```

include/cpr/exceptions.h
```diff
@@ -0,0 +1,19 @@
...
+                BadResponse(const std::string& msg, int http_code) : http_code(http_code), message(msg) {}
+                ~BadResponse() {}
```
COM8, line 13:
```suggestion
                ~BadResponse() = default;
```

test/exceptions_tests.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <string>
+
+#include "cpr/exceptions.h"
+#include <cpr/cpr.h>
+
+using namespace cpr;
+
+// HTTP code generator
+Url url{"https://httpstat.us/"};
```
COM8, line 11: Please do not use an external URL, since tests should not depend on an internet connection.
The best would be to add a modified case for this one:
https://github.com/libcpr/cpr/blob/eccea39e4b112a088ae665eda5854cc366f86128/test/httpServer.cpp#L30-L39
Where you pass the code, you want to receive as header.
https://github.com/libcpr/cpr/blob/eccea39e4b112a088ae665eda5854cc366f86128/test/httpServer.cpp#L193

COM8 (changes requested):

COM8:
I do not plan to actively throw exceptions.
This PR is a convenience function for throwing exceptions for http codes.
If something internally goes wrong, this call will also throw an exception since the `status_code` is 0 then.

COM8:
In a further PR we could extend this functionality a bit by including additional information when the `status_code` is 0.

PR #646, opened 2021-09-27
Fix CMake paths for cpr used with add_subdirectory

rlalik:
When using cpr with git submodule and `add_subdirectory()` function, the
```cmake
    configure_package_config_file(${CMAKE_SOURCE_DIR}/cmake/cprConfig.cmake.in
                                  "${CMAKE_BINARY_DIR}/cpr/cprConfig.cmake"
                                  INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/cpr)
```
evaluates `CMAKE_SOURCE_DIR` to the top-level project, not the cpr source location.

From https://cmake.org/cmake/help/latest/command/project.html it can be fixed by replacing `CMAKE_SOURCE_DIR` with `PROJECT_SOURCE_DIR` which evaluates to the `CMAKE_SOURCE_DIR` of cpr is a top-level project or to `CMAKE_SOURCE_DIR/<cpr or other subdirname>` if cpr is add as `add_subdirectory(cpr)` (or any other directory name).

For consistence, CMAKE_BUILD_DIR is replaced with PROJECT_BUILD_DIR.


KingKili:
Great catch thanks!
We should probably add some CI runs for Integration testing.

COM8:
> We should probably add some CI runs for Integration testing.
Agree.
This could be something we can do during [Hacktoberfest](https://hacktoberfest.digitalocean.com/).
We then can combine this with a couple of minimal demo repos for all ways one could use cpr.

PR #648, opened 2021-10-03
MIT for verything except the test directory (GPLv3)

COM8:
This PR aims to fix the license compatibility with mongoose.
[Mongoose](https://github.com/cesanta/mongoose) is provided under [GPLv2](https://github.com/cesanta/mongoose/blob/master/LICENSE) and is only consumed (linked against) by the tests.
So all tests are now under GPLv3, while everything else is still under MIT.

This should only matter in case you are linking against tests directly.
Everything else stays the same.

Also I updated the copyright information in the MIT license.

**I will keep this one open until the 10. October 2021 in case there are any concerns.**

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,10 @@
-The MIT License (MIT)
+This license applies to everything except the contents of the "test"
+directory and its subdirectories.
 
-Copyright (c) 2017 Huu Nguyen
+MIT License
+
+Copyright (c) 2017-2021 Huu Nguyen
+Copyright (c) 2021 Fabian Sauter
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -1,6 +1,10 @@
-The MIT License (MIT)
+This license applies to everything except the contents of the "test"
+directory and its subdirectories.
 
-Copyright (c) 2017 Huu Nguyen
+MIT License
+
+Copyright (c) 2017-2021 Huu Nguyen
+Copyright (c) 2021 Fabian Sauter
```
KingKili, line 7: I think the copyright should belong to libcpr not you. But at the end of the day I am not a Layer 🤷.

KingKili (commented):
Apart from my comment, I am fine. If we enforce dynamic linkage against mongoose we should be also fine with the old license, and eventually avoid using two licenses concurrently.

LICENSE
COM8: Agree.
Based on: https://fairuse.stanford.edu/overview/faqs/copyright-ownership/
and: http://haacked.com/archive/2006/01/26/WhoOwnstheCopyrightforAnOpenSourceProject.aspx/

Since we do not have any statement when people are contributing, stating something like „I hereby assign copyright in this code to the project, to be licensed under the same terms as the rest of the code.“ the copyright for their code stays with them.

So we should change it to something like: „libcpr and many other contributors“. This should be fine, since the term „libcpr“ refers indirectly to us maintainers and all other contributors.

Also I don‘t think switching away from this dual license will be possible any time soon, since every copyright owner has to agree with this change.

PR #649, opened 2021-10-03
Version macro

COM8:
Inspired by: https://github.com/rwinlib/libcurl/blob/v7.64.1/include/curl/curlver.h

Adds the following version macros to use:
```c++
/**
 * CPR version as a string.
 **/
#define CPR_VERSION "1.7.0"

/**
 * CPR version split up into parts.
 **/
#define CPR_VERSION_MAJOR 1
#define CPR_VERSION_MINOR 7
#define CPR_VERSION_PATCH 0

/**
 * CPR version as a single hex digit.
 * it can be split up into three parts:
 * 0xAABBCC
 * AA: The current CPR major version number in a hex format.
 * BB: The current CPR minor version number in a hex format.
 * CC: The current CPR patch version number in a hex format.
 *
 * Examples:
 * '0x010702' -> 01.07.02 -> CPR_VERSION: 1.7.2
 * '0xA13722' -> A1.37.22 -> CPR_VERSION: 161.55.34
 **/
#define CPR_VERSION_NUM 0x010702
```

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -34,6 +34,7 @@ target_sources(cpr PRIVATE
     cpr/interface.h
     cpr/redirect.h
     cpr/http_version.h
+    cpr/cprver.h
 )
 
 install(DIRECTORY cpr DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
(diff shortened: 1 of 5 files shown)


KingKili:
Overall it is fine, but I prefer constexpr variables, since those are a little more robust regarding compilers

nightlark:
The downside to using constexpr variables is that they aren't useable in preprocessor conditionals, which is one of the main places a way to check the version would be used.

COM8:
Then I'd say we keep it as macros.

PR #650, opened 2021-10-04
Disable SSL options for curl build when CPR_ENABLE_SSL is off

nightlark:
This PR makes so that disabling SSL can be done by just setting `CPR_ENABLE_SSL` to off. The current behavior prior to this change is that to disable SSL, `CPR_ENABLE_SSL` needs to be set to off along with finding and setting the correct `CMAKE_USE_*` variable to off. An extended description is after the break/line.

I also changed a comment about using environment variables to control options, since the current behavior is overriding the value of an option, not actually setting the value for the option in the CMake cache (the changed value doesn't persist if a new shell gets opened, or the environment variable get unset).

---

With the way the curl build system is currently set up, the `SSL_ENABLED` variable is set based on the `CMAKE_USE_*` options and isn't used to determine if curl should be built with SSL support. It is also setting the `SSL_ENABLED` local variable, so whatever cpr is setting in the cache for `SSL_ENABLED` is effectively being ignored.

By default curl sets the `CMAKE_USE_OPENSSL` option to ON, so when `CPR_ENABLE_SSL` is set to OFF we also need to set `CMAKE_USE_OPENSSL` to OFF.

The other `CMAKE_USE_*` options are also set to OFF, since it makes the behavior a bit cleaner if a subsequent CMake configure disables the `CPR_ENABLE_SSL` option (no tracking down which `CMAKE_USE_*` option also needs disabling).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,25 +176,28 @@ else()
     if (CPR_ENABLE_SSL)
         set(SSL_ENABLED ON CACHE INTERNAL "" FORCE)
         set(CURL_CA_PATH "auto" CACHE INTERNAL "" FORCE)
+
+        if(SSL_BACKEND_USED STREQUAL "WinSSL")
+            set(CMAKE_USE_SCHANNEL ON CACHE INTERNAL "" FORCE)
+        endif()
+
+        if(SSL_BACKEND_USED STREQUAL "OpenSSL")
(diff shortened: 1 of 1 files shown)


nightlark:
I also changed a comment about using environment variables to control options, since the current behavior is overriding the value of an option, not actually setting the value for the option in the CMake cache (the changed value doesn't persist if a new shell gets opened, or the environment variable get unset).

Issue #654, opened 2021-10-06
CMake configuration failure when used as subproject


thomasgt:
## Description
I recently experienced a CMake configuration error when using cpr as a subproject. In this case, I was pulling the project in using CMake's FetchContent, but I suspect this could happen when including cpr as a Git submodule.

## Expected Behavior
Successful CMake configuration and generation

## Actual Behavior
The error shown below:
```
CMake Error: File D:/my_project_source_dir/cmake/cprConfig.cmake.in does not exist.
CMake Error at C:/Program Files/CMake/share/cmake-3.20/Modules/CMakePackageConfigHelpers.cmake:342 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  build-RelWithDebInfo/_deps/pathcore_cpr-src/cpr/CMakeLists.txt:51 (configure_package_config_file)
```

## Possible Fix
`cpr/CMakeLists.txt` uses `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` where it should use `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR`, respectively.

## Steps to Reproduce
Try configuring a project that integrates cpr as a subproject using CMake's FetchContent facility.

This is how I am fetching cpr (I am fetching my fork, but that shouldn't matter):
```
FetchContent_Declare(
  pathcore_cpr
  GIT_REPOSITORY https://github.com/pathcore/cpr.git
  GIT_TAG add-virtual-destructors
)

# check if population has already been performed
FetchContent_GetProperties(pathcore_cpr)
if(NOT pathcore_cpr_POPULATED)
  # fetch the content using previously declared details
  FetchContent_Populate(pathcore_cpr)

  # set custom variables, policies, etc.
  set(CPR_BUILD_TESTS OFF CACHE BOOL "Set to ON to build cpr tests.")
  set(CPR_FORCE_USE_SYSTEM_CURL ON CACHE BOOL "If enabled we will use the curl lib already installed on this system.")

  # bring the populated content into the build
  add_subdirectory(${pathcore_cpr_SOURCE_DIR} ${pathcore_cpr_BINARY_DIR})
endif()

FetchContent_GetProperties(pathcore_cpr)
```

## Your Environment
Microsoft Windows 11 Pro (10.0.22000 Build 22000)
CMake v3.20.5
Visual Studio 2019 (MSVC 19.29.30038.1)

PR #652, opened 2021-10-06
Add virtual destructors to classes with virtual functions

thomasgt:
Resolves #651

PR #655, opened 2021-10-06
Fix CMake paths if subproject

thomasgt:
Resolves libcpr/cpr#654

This branch is based off of another PR branch, so it should be merged after https://github.com/libcpr/cpr/pull/652


KingKili:
Seems to be a duplicate of #646. But since 646 is stalling I will merge this PR. Thanks!

Issue #420, opened 2020-07-09
How could I use MBedTLS for SSL?


SeanOMik:
Is there a way I could use MBedTLS for curl? I tried to do this:
```cmake
set(USE_WINSSL OFF)
set(USE_OPENSSL OFF)
set(CMAKE_USE_WINSSL OFF)
set(CMAKE_USE_OPENSSL OFF)
set(CMAKE_USE_MBEDTLS ON)
```
But when I create a get request, I get no response with this error: `Cert verify failed: BADCERT_NOT_TRUSTED` 

Is there a way I could use MBedTLS with CPR?

COM8:
Currently I do not think we support `MBedTLS`, only `OpenSSL` and `WinSSL`.
This should be possible to add through.

SeanOMik:
Okay I'll just use WinSSL. I'm also having issues with OpenSSL on Linux, but I'll create a separate issue for it.

COM8:
@KingKili do you have time to add `MBedTLS` support?

COM8:
https://github.com/libcpr/cpr/pull/714 adds a mew option to enable Mbed TLS.

PR #657, opened 2021-10-16
Update curl from 7.75.0 to 7.79.1

nightlark:
Bump curl version -- numerous bugfixes, and support for mbedtls 3 (possibly a start towards working on #420).


COM8:
Thanks!

PR #658, opened 2021-10-17
Use CMake libcpr version number to generate cprver.h

nightlark:
Here's an idea that builds on #649 and makes so that the cprver.h file is generated based on the version in the CMake project command -- this would make so the version number only needs updating in one place when it comes time to make a new release.

Thoughts/ideas/changes?


COM8:
A really nice idea!
That makes it much easier to maintain.

PR #671, opened 2021-10-30
updated the conan example in the README

nidhaloff:
I needed to make an http client in c++ and I found this lib yesterday. The API is great, I tried it and kept getting an error using conan. It was frustrating and I thought of trying another lib. Today I opened an issue and give it another try, the solution was very simple and it was solved by adding ABI definition to my cmake setup.

I updated the conan example in the README to notify future users that this can be the issue. Instead of getting frustrated, users can uncomment that line responsible for the ABI config and try again.


KingKili:
Thanks, great Idea!

KingKili (approved):

PR #675, opened 2021-11-05
CI: Add cppcheck

Garfield96:
This PR adds cppcheck to the CI workflow. 

Currently, the source files are specified explicitly. Instead, `compile_commands.json` generated by cmake could be used. However, this would also include the dependencies, which is not desired and slows down the analysis.

@COM8


cmake/cppcheck.cmake
```diff
@@ -0,0 +1,9 @@
+find_program(CMAKE_CXX_CPPCHECK NAMES cppcheck)
+if(CMAKE_CXX_CPPCHECK)
+    list(APPEND CMAKE_CXX_CPPCHECK 
+        "--error-exitcode=1"
+        "--enable=warning,style"
+        "--force" 
+        "--inline-suppr"
+        "--suppressions-list=${CMAKE_SOURCE_DIR}/CppCheckSuppressions.txt")
```
Garfield96, line 8: The input files are not specified, hence this does nothing.

cpr/curl_container.cpp
```diff
@@ -9,6 +9,8 @@ CurlContainer<T>::CurlContainer(const std::initializer_list<T>& containerList)
 template <class T>
 void CurlContainer<T>::Add(const std::initializer_list<T>& containerList) {
     for (const T& element : containerList) {
+        // To use std::transform like suggested we would have to upgrade to cpp17
+        // cppcheck-suppress useStlAlgorithm
```
Garfield96, line 13: We could lower the standard in the cppcheck flags, which should make this suppression unnecessary.

.github/workflows/ci.yml
```diff
@@ -58,6 +58,29 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2.2.0
+      with:
+        submodules: true
+    - name: Update package list
+      run: sudo apt update
+    - name: Install cppcheck
+      run: sudo apt install cppcheck
```
Garfield96, line 71: Github complains about `apt`:
```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```
Maybe we should exchange it with `apt-get`.

cmake/cppcheck.cmake
COM8: CMake automatically adds all source files here.

Garfield96: I'm not sure about that. But it looks like cppcheck is not even invoked during the build since there is no output in the build logs and the build should fail, because there are still some open issues.

cpr/curl_container.cpp
COM8: Good point.

.github/workflows/ci.yml
COM8: I would keep it that way since we are not using any complex apt operations and in my eyes this is the more further proof command.

cmake/cppcheck.cmake
COM8: Are you sure. Works for me. Did you enable it with `cmake .. -DCPR_ENABLE_CPPCHECK=ON`?

KingKili:
I strongly discourage the addition of cpp check. In my opinion it does not add any benefit over clangtidy and clang format (am I missing something). Instead this will lead to additional hassle for contributers, and unnesesery no lint comments. Therefore I would opt to not include it.

cmake/cppcheck.cmake
COM8: And the CI run also has the output for it:
```bash
[...]
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__TINYC__...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__TURBOC__...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__VAX;__VMS...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__VMS...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__VXWORKS__;__WATCOMC__...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__WATCOMC__...
  Checking /home/runner/work/cpr/cpr/cpr/curlholder.cpp: CPR_ENABLE_CPPCHECK=1;CPR_ENABLE_SSL=1;cpr_EXPORTS=1;NDEBUG=1;__xlc__...
[...]
```

Garfield96:
The benefit of cppcheck is that it finds different issues, e.g. there are a bunch of unused variables, etc. which were not found by clang-tidy. As far as I can tell,  no additional no-lint comments are needed.

cmake/cppcheck.cmake
Garfield96: You're right. I got confused because the output looks different from what I expected. A disadvantage of the current approach is that cppcheck is invoked for each source file separately. Should you ever consider searching for unused functions, this will be a problem.

Issue #382, opened 2020-05-20
C++17/C++20 support


COM8:
We should update the complete library to `C++20` or at least `C++17` ready.

ZaMaZaN4iK:
I cannot understand the reason? Why? Do you want port whole CPR project to C++17 or even C++20? Remember, that a lot of users don't have even C++17 in their applications. So if you update the library to such new C++ standard, a lot of users will not be able to use CPR.

Do you need any specific stuff from C++17/20?

My personal point of view - that's acceptable to grade someday to C++17, but for now that's totally unacceptable grade the library to C++20 (since it's even officially released and C++ compilers have poor support for C++20). Please  - don't upgrade cpr to C++20.

Chlorie:
> I cannot understand the reason? Why? Do you want port whole CPR project to C++17 or even C++20? Remember, that a lot of users don't have even C++17 in their applications. So if you update the library to such new C++ standard, a lot of users will not be able to use CPR.
> 
> Do you need any specific stuff from C++17/20?
> 
> My personal point of view - that's acceptable to grade someday to C++17, but for now that's totally unacceptable grade the library to C++20 (since it's even officially released and C++ compilers have poor support for C++20). Please - don't upgrade cpr to C++20 now.
> 
> Thank you.

I think what they meant was to make CPR better integrated with C++17/20 if your build settings support those, like with version detection macros so that if you use C++17/20 CPR will support new fancy things (like coroutines?) but if your compiler don't have the support those parts of the library are simply not accessible (disabled by preprocessor). I don't see how this will be a problem for people using older compilers.

COM8:
My plan is to make use of `C++17` features like [`std::opt`](https://en.cppreference.com/w/cpp/utility/optional) and the new default deleter for [`std::unique_ptr`](https://de.cppreference.com/w/cpp/memory/unique_ptr) and [`std::shared_ptr`](https://de.cppreference.com/w/cpp/memory/shared_ptr).
Besides that create a better [RAII](https://en.cppreference.com/w/cpp/language/raii) wrapper for curl sessions to properly support multithreaded access.
Perhaps even offer support for [lambdas](https://de.cppreference.com/w/cpp/language/lambda) and [coroutines](https://en.cppreference.com/w/cpp/language/coroutines).

Since it's rather early for `C++20` features in such a library, this will be optional stuff and compilations should be possible with  `C++17` **and** `C++20`.

[...]

PR #682, opened 2021-11-27
use std::string_view for writecallback

shaforostoff:
to avoid unnecessary memory allocations

### DIFF
--- a/include/cpr/callback.h
+++ b/include/cpr/callback.h
@@ -40,13 +40,13 @@ class WriteCallback {
   public:
     WriteCallback() = default;
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
-    WriteCallback(std::function<bool(std::string data, intptr_t userdata)> p_callback, intptr_t p_userdata = 0) : userdata(p_userdata), callback(std::move(p_callback)) {}
-    bool operator()(std::string data) const {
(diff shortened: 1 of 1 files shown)


COM8:
Thanks for contributing.
The problem with `std::string_view`is, it's [cpp17](https://en.cppreference.com/w/cpp/header/string_view) and this lib is cpp11.
So to merge this change you could add an alternative call which is only available if the compiler supports it. Here we could use the `__cplusplus` macro.

shaforostoff:
is it a hard requirement to still support c++11? do you know of any cpr users that have such requirement?
 
there are many more places which could be improved with std::string_view, for example cpr::util::split could be optimized to only allocate memory for std::vector, also cpr::util::parseHeader

KingKili:
If you have a look at #382, you can see that users are very hesitant against updating to newer versions. Therefore the plan is currently to move first to c++ 14.

shaforostoff:
i can do it, but it would be a pain to maintain. Please ping me when cpr will move to c++17, and I will create MR with more string_view related optimizations

COM8:
> i can redo MR with ifdef's, but it would be a pain to maintain. Please ping me when decision is made to move cpr to c++17, and I will create MR with more string_view related optimizations (split, parseCookies, parseHeader)

Yes, I don't think this would be a good way to go.

PR #690, opened 2021-12-15
CMake: install dll in bin folder

SpaceIm:
fix install destination of dll (broken since https://github.com/libcpr/cpr/pull/645)


KingKili:
Thanks!

KingKili (approved):

COM8:
@SpaceIm do you need a new CPR release with this fix included, since you are working on an updated conan package?

SpaceIm:
If it's just for conan, no it's fine, thanks. I've already included the patch, and it will be pulled out for the packaging of the next version of CPR.

PR #692, opened 2021-12-19
[clang-tidy] don't default to warnings as errors

Green-Sky:
I had issues with a new clang-tidy version (12), discovering new issues with the code, with cpr as a submodule. So the build of my project failed because clang-tidy uses the cpr .clang-tidy file.

I propose this change, since your CI runner [overrides](https://github.com/libcpr/cpr/blob/master/cmake/clang-tidy.cmake#L9) the setting anyway and this makes it integration/future proof, so projects using cpr as a submodule don't just fail because clang enhances clang-tidy with new features. :smirk: 

The generated error you probably want to investigate:
```
cpr/cpr/util.cpp:28:8: error: function 'parseHeader' has cognitive complexity of 31 (threshold 25) [readability-function-cognitive-complexity,-warnings-as-errors]
Header parseHeader(const std::string& headers, std::string* status_line, std::string* reason) {
       ^
cpr/cpr/util.cpp:34:9: note: +1, including nesting penalty of 0, nesting level increased to 1
        while (std::getline(stream, line, '\n')) {
        ^
cpr/cpr/util.cpp:39:5: note: +1, including nesting penalty of 0, nesting level increased to 1
    for (std::string& line : lines) {
    ^
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,6 +27,6 @@ Checks: '*,
 -llvmlibc-implementation-in-namespace,
 -llvmlibc-restrict-system-libc-headers
 '
-WarningsAsErrors: '*'
+WarningsAsErrors: ''
 HeaderFilterRegex: 'src/*.hpp'
 FormatStyle: file


COM8:
I'm sad to say, but I'm not a fan of that since this would disable errors when linting locally and working on cpr.
Any reason why one could not just disable linting for CPR in your project with: https://github.com/libcpr/cpr/blob/fc0e0f86ce418733c2e7bbd5b71b923d30406cb8/CMakeLists.txt#L314-L319

Or you also could specify your own `.clang-tidy` file.

Green-Sky:
I see your concerns, but your `.clang-tidy` file supersedes mine. If you found a setting i could set in my `.clang-tidy` file to eg. disable linting of certain subfolders, that would work for me too.

COM8:
Taken from `clang-tidy -help`

> [...] .clang-tidy file located in the closest parent directory of the source file.

So I think there is something wrong with the way you invoke clang-tidy. It only searches for config files in parent directories.

Green-Sky:
Yes, because I use `cpr` as a submodule, and inject clang-tidy via `CMAKE_CXX_CLANG_TIDY` in the parent's `CMakeLists.txt` it also calls clang-tidy for your projects files. And your `.clang-tidy` is of cause "[...] located in the closest parent directory of the source file."


Also! I forgot to ask whether I should create a Issue for the warning that clang-tidy actually throws. (the real issue imho)

COM8:
> Also! I forgot to ask whether I should create a Issue for the warning that clang-tidy actually throws. (the real issue imho)
Oh totally forgot about this. Yes, please!

KingKili:
I also opt to not disable warning as errors, since the mentioned issue is fixed since [#9b2bda734dffdd76096e1364290c58fe3a30c59a](https://github.com/libcpr/cpr/commit/9b2bda734dffdd76096e1364290c58fe3a30c59a).

Green-Sky:
Ok, so since the issue seems to be resolved, I will not open an issue.

Cheers

PR #696, opened 2022-01-17
Fix build with OpenSSL on Ubuntu bionic

TheAssassin:
Not sure why the `SSL` component is checked (seems superfluous compared to HTTPS), but this check does not work on Ubuntu bionic (cURL 7.58). As far as I can see, it can be removed safely, as long as the HTTPS check is retained.


COM8 (approved):
Agree. Thanks!

PR #698, opened 2022-01-31
DownloadAsync added

tomaszewskip:
DownloadAsync uses std::launch::async to Download a remote file to a local file system.


COM8 (changes requested):
Could you please a a `DownloadAsyncTests` in [`async_tests.cpp`](https://github.com/libcpr/cpr/blob/master/test/async_tests.cpp)? You can mostly copy it from [here](https://github.com/libcpr/cpr/blob/master/test/download_tests.cpp).

COM8 (approved):
Thanks!

PR #699, opened 2022-01-31
KeyBlob introduced

tomaszewskip:
Options let to set an SSL private key using a path to a file only.
Such approach prevents against setting the private key, which is stored
in a secure storage and is extracted using an algorithm.

Patch introduces KeyBlob, which translates to the CURLOPT_SSLKEY_BLOB directly.
It can be set using SslOptions::SetOption(const ssl::KeyBlob&) API.
It is assumed that blob content is PEM.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <cstdlib>
+#include <cstring>
 #include <fstream>
 #include <functional>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


COM8:
Thanks, could you please create a PR against the [docs](https://github.com/libcpr/docs) repo updating the related docs: https://docs.libcpr.org/advanced-usage.html#https-options?

Also there are a couple clang-tidy issues:
```
  /home/runner/work/cpr/cpr/cpr/session.cpp:490:7: error: uninitialized record type: 'blob' [cppcoreguidelines-pro-type-member-init,-warnings-as-errors]
        struct curl_blob blob;
        ^
                             {}
  /home/runner/work/cpr/cpr/cpr/session.cpp:490:7: error: uninitialized record type: 'blob' [hicpp-member-init,-warnings-as-errors]
        struct curl_blob blob;
        ^
                             {}
  /home/runner/work/cpr/cpr/cpr/session.cpp:491:19: error: do not use const_cast [cppcoreguidelines-pro-type-const-cast,-warnings-as-errors]
        blob.data = const_cast<char *>(options.key_blob.c_str());
                    ^
```

cpr/session.cpp
```diff
@@ -486,6 +486,17 @@ void Session::Impl::SetSslOptions(const SslOptions& options) {
         if (!options.key_pass.empty()) {
             curl_easy_setopt(curl_->handle, CURLOPT_KEYPASSWD, options.key_pass.c_str());
         }
+    } else if (!options.key_blob.empty()) {
+      struct curl_blob blob;
```
COM8, line 490:
```suggestion
      curl_blob blob{};
```
Should fix the clang tidy issue and since we are using C++, we don't need the `struct` keyword here.

cpr/session.cpp
```diff
@@ -486,6 +486,17 @@ void Session::Impl::SetSslOptions(const SslOptions& options) {
...
+      blob.data = const_cast<char *>(options.key_blob.c_str());
```
COM8, line 491: Don't use a `const_cast` here. A solution would be to store a mutable copy of `key_blob` in `KeyBlob`.

COM8 (changes requested):

cpr/session.cpp
tomaszewskip: Is there a way to run checks by hand using cmake?

COM8: In case you have clang-tidy installed on your system you can do something like this:
```bash
mkdir build
cd build
cmake -DCPR_ENABLE_LINTING=ON ..
cmake --build .
```

COM8: Here you can find a list of all options available: https://github.com/libcpr/cpr/blob/da381bb2afe4e36b18b1fe206bc06951d8fdc06e/CMakeLists.txt#L36-L49

cpr/session.cpp
tomaszewskip: Requested changes resolved in d68d6ee.

cpr/session.cpp
```diff
@@ -486,6 +487,36 @@ void Session::Impl::SetSslOptions(const SslOptions& options) {
         if (!options.key_pass.empty()) {
             curl_easy_setopt(curl_->handle, CURLOPT_KEYPASSWD, options.key_pass.c_str());
         }
+    } else if (!options.key_blob.empty()) {
+        struct ScopedCharArray {
```
COM8, line 491: Why don't we use something like this here:
```c++
std::string str{"Hello world!"};
char* strPtr = &(str[0]);
```

COM8 (changes requested):

COM8 (approved):
Looks good now. Thanks for your help!

PR #701, opened 2022-02-01
75 - basic range support added

tomaszewskip:
Basic support for ranges. Tests are disabled as they are connecting to the cURL server, which supports range header. Tested with Download only.

### DIFF
--- a/test/download_tests.cpp
+++ b/test/download_tests.cpp
@@ -29,6 +29,54 @@ TEST(DownloadTests, DownloadGzip) {
     EXPECT_EQ(cpr::ErrorCode::OK, response.error.code);
 }
 
+TEST(DownloadTests, RangeTestWholeFile) {
+#if 0
+    const int64_t download_size = 277186;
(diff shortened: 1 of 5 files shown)


test/download_tests.cpp
```diff
@@ -29,6 +29,54 @@ TEST(DownloadTests, DownloadGzip) {
...
+    session.SetSslOptions(sslOpts);
+    cpr::Response response = session.Download(cpr::WriteCallback{write_data, 0});
+    EXPECT_EQ(200, response.status_code);
+    EXPECT_EQ(cpr::ErrorCode::OK, response.error.code);
+    EXPECT_EQ(download_size, response.downloaded_bytes);
+#endif
+}
+
+TEST(DownloadTests, RangeTestLowerLimit) {
+#if 0
```
COM8, line 49: Tests look good, but I'm not a fan of tests that require internet here. Could you have a look at implementing rages support for our [httpServer](https://github.com/libcpr/cpr/blob/master/test/httpServer.hpp) we use for testing?

COM8 (changes requested):
Nice, thanks! Only the thing with the tests.

COM8 (approved):
Awesome. Thanks!

PR #707, opened 2022-03-15
Fix for certificate verification on android

twisted89:
Fixes the CA path for android, without it cpr::VerifySsl(true) always fails with "unable to get local issuer certificate ".


KingKili (approved):
Looks good to me. Thanks!

COM8 (approved):
Thanks!

PR #712, opened 2022-03-20
Option to reserve space for the response string beforehand

### LINKED ISSUE #530: Reserve space beforehand for large download

COM8:
Closes #530

Example:
```c++
cpr::Session session;
session.SetUrl(cpr::Url{"http://xxx/file"});
session.ResponseStringReserve(1024 * 512); // Reserve space for at least 1024 * 512 characters
cpr::Response r = session.Get();
```

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -70,6 +70,7 @@ class Session::Impl {
     void SetRange(const Range& range);
 
     cpr_off_t GetDownloadFileLength();
+    void ResponseStringReserve(size_t size);
     Response Delete();
     Response Download(const WriteCallback& write);
     Response Download(std::ofstream& file);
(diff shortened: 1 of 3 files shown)


COM8:
Docs: https://github.com/libcpr/docs/pull/10

PR #716, opened 2022-03-23
Not including curl.h in headers

### LINKED ISSUE #700: cpr.h includes windows.h on Windows
On windows, `cpr.h` includes `curl.h`, which itself includes `windows.h`.

## Description
The `windows.h` header is a very annoying header, which defines a number of common words/expressions as macros (such as `min`, `max`, `CopyFile`, etc.). It is usually a bad idea to include it in a header, as it then pollutes all files including said header. For example, it prevents us from creating our own function called `CopyFile`, unless we explicitly `#undef CopyFile` after including `cpr.h`. 

## Expected Behavior
Including `cpr.h` should only include CPR types, and should not define annoying macros.

## Actual Behavior
Including `cpr.h` includes `windows.h` :( Here is one of the include chains I could spot:
 - `cpr/cpr.h` -> `cpr/response.h` -> `curl/curl.h` -> `winsock2.h` -> `windows.h`

## Possible Fix
Including `curl.h` should be avoided in CPR headers.

## Steps to Reproduce
1. Include `cpr.h` on a Windows development environment
2. Watch the macros being defined! (e.g. `CopyFile` -> `CopyFileA`)

## Context
We were including `cpr.h` in one of our core headers, which means our entire code base had to live with `windows.h` being included. We since managed to move the `#include cpr.h` to just the `.cpp` files that needed it, but it required some annoyance (e.g., can't forward-declare `cpr::Header` because it is a typedef).

## Your Environment
* Version used: 1.6.2
* Where did you get it from (e.g. conan, vcpkg, master, ...): github
* Operating System and version: Windows 10 Pro
* Link to a small example:
```c++
void CopyFile() {  /*...*/ }

#include "cpr/cpr.h"

int main() {
    CopyFile(); // compiler error
}
```

COM8:
To prevent transitively including `windows.h` on windows when including `cpr.h` or `cprtypes.h`, `curl.h` includes got replaced or removed.

Fixes #700

### DIFF
--- a/include/cpr/cprtypes.h
+++ b/include/cpr/cprtypes.h
@@ -1,7 +1,7 @@
 #ifndef CPR_CPR_TYPES_H
 #define CPR_CPR_TYPES_H
 
-#include <curl/curl.h>
+#include <curl/system.h>
 #include <initializer_list>
 #include <map>
 #include <memory>
--- a/include/cpr/response.h
+++ b/include/cpr/response.h
@@ -3,7 +3,6 @@
 
 #include <cassert>
 #include <cstdint>
-#include <curl/curl.h>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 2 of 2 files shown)


cschreib-ibex:
Thanks for the PR, will be great to have this :) Weirdly, this is advertised as part of the 1.8.0 release, which was just released, but this PR isn't merged?

COM8:
I guess I failed that one 🤦‍♂️.
I guess 1.8.1 will be released shortly.

PR #717, opened 2022-03-24
Fix overwriting of ca bundle

### LINKED ISSUE #445: SSL certificate problem
SSL certificate problem: self signed certificate in certificate chain
SSL certificate problem: unable to get local issuer certificate

Leon0402:
Closes #445


COM8:
Thanks!

PR #726, opened 2022-04-16
Improve usability for ResponseStringReserve

WorkingRobot:
The current interface for ResponseStringReserve doesn't allow `cpr::Get` and other similar functions to make use of it. The current implementation here is backwards compatible with previous versions, at the cost of making the current function redundant. Any feedback is appreciated.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -68,6 +68,7 @@ class Session::Impl {
     void SetInterface(const Interface& iface);
     void SetHttpVersion(const HttpVersion& version);
     void SetRange(const Range& range);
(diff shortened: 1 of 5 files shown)


COM8 (changes requested):
Thanks! This is a valid point.
Cloud you please also add a call which takes `ReserveSize` for the session impl.
So this function is overloaded and we keep a "consistent" interface there.

Also, please update the docs here:
https://docs.libcpr.org/advanced-usage.html#session-objects

For this create a PR here:
https://github.com/libcpr/docs

COM8 (approved):
Looks good. Thanks!

PR #728, opened 2022-04-22
Make sure mutex is properly initialized

cordbleibaum:
I encountered access violations in a static debug build included in a Windows DLL for another application. The mutex crashed on use due to not being initialized. This is avoidable by using an init function.


COM8 (approved):
Aha the good old static problem :)
Thank you!

PR #738, opened 2022-05-04
Use UpdateHeader to allow header updates from different sources

### LINKED ISSUE #265: Better Session::SetHeader() implementation
Currently, `SetHeader()` method sets completely new HTTP header every time `cpr::Header` arg is set in `cpr::Post` (or whatever). If we set several different args like that, they will get overriden.

It would be nice to have a possibility to set headers from different sources, for example when injecting params and headers to the request by different helpers.

    template<class ...Ts>
    cpr::Response myPost(Ts&& ...ts) {
        return cpr::Post(
            std::forward<Ts>(ts)...,
            cpr::Header{ { "Authorization", "PuppyMaster" } });
    }
    
    auto resp = myPost(
        cpr::Url{ "http://service.com/puppies" },
        cpr::Header{ { "Content-Type", "application/json" } }
        cpr::Body{ myPuppy });

This code will make only "Injected: Header" header be sent, "Content-Type" header is ignored and the web service we invoke gets grumpy.

This can be easily fixed. I may prepare a PR next week if you want to be an expected behavior.

If the current behavior is expected, how else could I handle the issue? I.e. how can I inject request headers **easily**?

simon-berger:
Closes #265

Makes it possible to set (and update) headers from different sources, for example when injecting params and headers to the request by different helpers.

### DIFF
--- a/include/cpr/api.h
+++ b/include/cpr/api.h
@@ -24,10 +24,31 @@ using AsyncResponse = std::future<Response>;
 
 namespace priv {
 
+template <bool processed_header, typename CurrentType>
+void set_option_internal(Session& session, CurrentType&& current_option) {
+    session.SetOption(std::forward<CurrentType>(current_option));
+}
+
+template <>
+inline void set_option_internal<true, Header>(Session& session, Header&& current_option) {
+    // Header option was already provided -> Update previous header
+    session.UpdateHeader(std::forward<Header>(current_option));
+}
+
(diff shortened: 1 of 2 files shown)


COM8 (approved):
Nice one!
Please add an example in the docs here: https://docs.libcpr.org/advanced-usage.html#request-headers

simon-berger:
Docs: https://github.com/libcpr/docs/pull/15

PR #742, opened 2022-05-10
Improved range requests and added support for multiple ranges

### LINKED ISSUE #75: Range request support
Can't find anything about support for this in the options, or docs.

Doable via CURLOPT_RANGE.

simon-berger:
Closes #75 

Improves range requests introduced in #701 and added support for multiple ranges in a single request ([multipart ranges](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#multipart_ranges)).

### DIFF
--- a/include/cpr/range.h
+++ b/include/cpr/range.h
@@ -26,7 +26,7 @@ class MultiRange {
     const std::string str() const {
         std::string multi_range_string{};
         for (Range range : ranges) {
-            multi_range_string += ((multi_range_string.empty()) ? "" : ",") + range.str();
+            multi_range_string += ((multi_range_string.empty()) ? "" : ", ") + range.str();
         }
         return multi_range_string;
     }
--- a/include/cpr/range.h
+++ b/include/cpr/range.h
@@ -11,6 +11,10 @@ class Range {
 
     std::int64_t resume_from = 0;
     std::int64_t finish_at = 0;
+
+    const std::string str() const {
+        return std::to_string(resume_from) + "-" + std::to_string(finish_at);
+    }
 };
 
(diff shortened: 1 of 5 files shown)


include/cpr/range.h
```diff
@@ -11,8 +11,30 @@ class Range {
 
     std::int64_t resume_from = 0;
     std::int64_t finish_at = 0;
+
+    const std::string str() const {
+        std::string from_str = (resume_from == -1) ? "" : std::to_string(resume_from);
```
COM8, line 16: Where is `resume_from` set to `-1`? Since I was thinking about replacing `resume_from == -1` with `resume_from < 0`.
This is a safer alternative to check for ranges while we do not ECC-RAM everywhere :)

COM8:
Besides the one comment, it looks good. Thanks!

include/cpr/range.h
simon-berger: -1 is the value in the documentation for creating partial ranges. Nevertheless, < 0 is the safer alternative.

COM8:
This happens sometimes. I guess their agent is broken. I restarted the failed tests.
```
error: downloading 'https://github.com/google/googletest/archive/release-1.11.0.tar.gz' failed
            status_code: 22
            status_string: "HTTP response code said error"
            log:
            --- LOG BEGIN ---
              Trying 192.30.255.112:443...
  
    Connected to github.com (192.30.255.112) port 443 (#0)
```

simon-berger:
Thanks 👍

COM8 (approved):
Ready to go. Thanks!

Issue #683, opened 2021-11-29
How to disable post request compression


deadash:
## Description
When I use a post request to upload a file, if the file is a bit big, it will be automatically compressed, and I need to upload it as it is.

## Expected Behavior
Upload without compression

## Actual Behavior
Will be compressed by gzip, the file header is pk

## Possible Fix
Don't know how to do it, maybe curl has the answer?

COM8:
As far as I'm aware curl does not automatically compress files for upload. But perhaps it does it in case the server announces gzip support.

What happens if you do something like:
```c++
Session session;
session.SetUrl(url);
curl_easy_setopt(session.GetCurlHolder()->handle, CURLOPT_ACCEPT_ENCODING, "identity");
// [...] Your other code for the request
```

deadash:
The client sends a POST request directly, how does it know whether the server supports it? Did the server reject it after sending it?

The following is to view the sent header
Accept-Encoding: deflate, gzip

Through your tips, after I modified the code
Accept-Encoding: identity

Lines of code related to this， https://github.com/libcpr/cpr/blob/bcb7729dd6d930820fb858f63c97ad559241b789/cpr/session.cpp#L749

COM8:
Sorry, I do not understand your reply there.

I understood:
* This fixed it.
* We are enabling compression for send data with the line you are highlighting.

deadash:
Yes, your answer fixed it.

I hope this setting can be easily set. Just like setting Proxies.

leviliangtw:
I'd like to take this issue.

barisconur:
I would like to work on this issue.

KingKili:
@leviliangtw @barisconur, can you both agree on one of you? Not that you both create the same solution.

leviliangtw:
@barisconur, have you started this issue already?
I think I have an idea for it.

barisconur:
@leviliangtw I have not started yet. Do you have a chance to implement and test your idea?

leviliangtw:
@barisconur Yes, let me get the chance. 
@KingKili I take this issue.

leviliangtw:
@KingKili The PR [#746](https://github.com/libcpr/cpr/pull/746) is ready for this issue

COM8:
Closing since this has been solved #746 .

PR #746, opened 2022-05-17
Add method SetAcceptEncoding for customized Accept-Encoding header (#683)

leviliangtw:
1. New class for managing header Accept-Encoding refers to proxies structure.cpp
2. New method to set header Accept-Encoding
3. New unit test to verify header Accept-Encoding

### DIFF
--- a/cpr/accept_encoding.cpp
+++ b/cpr/accept_encoding.cpp
@@ -1,0 +1,25 @@
+#include "cpr/accept_encoding.h"
+
+#include <algorithm>
+#include <iterator>
+#include <numeric>
+
+namespace cpr {
+
+AcceptEncoding::AcceptEncoding(const std::initializer_list<AcceptEncodingMethods>& methods) {
+    methods_.clear();
+    std::transform(methods.begin(), methods.end(), std::back_inserter(methods_), [&](cpr::AcceptEncodingMethods method) { return cpr::AcceptEncodingMethodsStringMap.at(method); });
+}
+
+AcceptEncoding::AcceptEncoding(const std::initializer_list<std::string>& string_methods) : methods_{string_methods} {}
+
+bool AcceptEncoding::empty() const noexcept {
+    return methods_.empty();
+}
+
+const std::string AcceptEncoding::getString() const {
+    std::string methodsString = std::accumulate(std::next(methods_.begin()), methods_.end(), methods_[0], [](std::string a, std::string b) { return std::move(a) + ", " + std::move(b); });
+    return methodsString;
+}
+
+} // namespace cpr
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -76,6 +76,8 @@ class Session::Impl {
     void SetRange(const Range& range);
     void SetMultiRange(const MultiRange& multi_range);
     void SetReserveSize(const ReserveSize& reserve_size);
+    void SetAcceptEncoding(AcceptEncoding&& accept_encoding);
+    void SetAcceptEncoding(const AcceptEncoding& accept_encoding);
 
     cpr_off_t GetDownloadFileLength();
     void ResponseStringReserve(size_t size);
@@ -115,6 +117,7 @@ class Session::Impl {
     Proxies proxies_;
     ProxyAuthentication proxyAuth_;
     Header header_;
+    AcceptEncoding acceptEncoding_;
     /**
      * Will be set by the read callback.
      * Ensures that the "Transfer-Encoding" is set to "chunked", if not overriden in header_.
@@ -600,6 +603,14 @@ void Session::Impl::SetReserveSize(const ReserveSize& reserve_size) {
     ResponseStringReserve(reserve_size.size);
 }
 
+void Session::Impl::SetAcceptEncoding(const AcceptEncoding& accept_encoding) {
+    acceptEncoding_ = accept_encoding;
+}
+
+void Session::Impl::SetAcceptEncoding(AcceptEncoding&& accept_encoding) {
+    acceptEncoding_ = std::move(accept_encoding);
+}
+
 void Session::Impl::PrepareDelete() {
     curl_easy_setopt(curl_->handle, CURLOPT_HTTPGET, 0L);
     curl_easy_setopt(curl_->handle, CURLOPT_NOBODY, 0L);
@@ -830,8 +841,13 @@ void Session::Impl::prepareCommon() {
 
 #if LIBCURL_VERSION_MAJOR >= 7
 #if LIBCURL_VERSION_MINOR >= 21
-    /* enable all supported built-in compressions */
-    curl_easy_setopt(curl_->handle, CURLOPT_ACCEPT_ENCODING, "");
+    if (acceptEncoding_.empty()) {
+        /* enable all supported built-in compressions */
+        curl_easy_setopt(curl_->handle, CURLOPT_ACCEPT_ENCODING, "");
+    }
+    else {
+        curl_easy_setopt(curl_->handle, CURLOPT_ACCEPT_ENCODING, acceptEncoding_.getString().c_str());
+    }
 #endif
 #endif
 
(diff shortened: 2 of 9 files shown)


COM8:
```
 1: [----------] 5 tests from BasicAuthenticationTests
  1: [ RUN      ] BasicAuthenticationTests.BasicAuthenticationSuccessTest
  1: /home/runner/work/cpr/cpr/test/abstractServer.cpp:108:20: runtime error: left shift of 492624220 by 6 places cannot be represented in type 'int'
  1: HEADERS: Authorization: Basic dXNlcjpwYXNzd29yZA==
  1: User-Agent: curl/7.81.0
  1: Accept-Encoding: deflate, gzip
  1: Content-Type: text/html
  1: [       OK ] BasicAuthenticationTests.BasicAuthenticationSuccessTest (1 ms)
  1: [ RUN      ] BasicAuthenticationTests.BasicBearerSuccessTest
  1: HEADERS: Authorization: ***
  1: User-Agent: curl/7.81.0
  1: Accept-Encoding: deflate, gzip
  1: Content-Type: text/html
```

cpr/accept_encoding.cpp
```diff
@@ -0,0 +1,23 @@
+#include "cpr/accept_encoding.h"
+
+namespace cpr {
+
+AcceptEncoding::AcceptEncoding(const std::initializer_list<AcceptEncodingMethods>& methods) : methods_{methods} {}
+
+bool AcceptEncoding::isEmpty() const {
```
COM8, line 7: All public methods have to start with an upper case letter.

include/cpr/accept_encoding.h
```diff
@@ -0,0 +1,40 @@
...
+enum class AcceptEncodingMethods {
+  identity,
+  deflate, 
+  zlib, 
+  gzip,
+};
+
+class AcceptEncoding {
+  public:
+    std::map<AcceptEncodingMethods, std::string> MethodsString {
```
COM8, line 21: Although I like the way you are solving this here, curl allows even more encodings: https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

Since this is something that changes based on what is build into curl, in my eyes we need at least an additional option allowing the user to specify their own encoding via a string name.

COM8 (changes requested):
Thanks for looking into this.

cpr/accept_encoding.cpp
leviliangtw: I would like to change its name to `has`, just like the implementation in proxies.cpp:
```c++
bool Proxies::has(const std::string& protocol) const {
    return hosts_.count(protocol) > 0;
}
```

COM8: This is also possible.

test/abstractServer.cpp
```diff
@@ -105,7 +105,7 @@ std::string AbstractServer::Base64Decode(const std::string& in) {
             break;
         }
         // NOLINTNEXTLINE (cppcoreguidelines-avoid-magic-numbers)
-        val = (val << 6) + T[c];
+        val = ((uint)val << (uint)6) + T[c];
```
COM8, line 108: Please don't use C-style casts
```suggestion
        val = (static_cast<uint>(val) << static_cast<uint>(6)) + T[c];
```

include/cpr/accept_encoding.h
```diff
@@ -0,0 +1,41 @@
...
+enum class AcceptEncodingMethods {
+  identity,
+  deflate, 
+  zlib, 
+  gzip,
+};
+
+class AcceptEncoding {
+  public:
+    std::map<AcceptEncodingMethods, std::string> MethodsStringMap {
```
COM8, line 21: Can't this map be `static` and `const`?

include/cpr/accept_encoding.h
```diff
@@ -0,0 +1,41 @@
...
+      {AcceptEncodingMethods::deflate, "deflate"}, 
+      {AcceptEncodingMethods::zlib, "zlib"}, 
+      {AcceptEncodingMethods::gzip, "gzip"} 
+    };
+
+    AcceptEncoding() = default;
+    AcceptEncoding(const std::initializer_list<AcceptEncodingMethods>& methods);
+    AcceptEncoding(const std::initializer_list<std::string>& methods);
+
+    bool has() const;
```
COM8, line 32: I thought about it again and I prefer `empty()` here as name.

cpr/session.cpp
```diff
@@ -792,8 +803,13 @@ void Session::Impl::prepareCommon() {
 
 #if LIBCURL_VERSION_MAJOR >= 7
 #if LIBCURL_VERSION_MINOR >= 21
-    /* enable all supported built-in compressions */
-    curl_easy_setopt(curl_->handle, CURLOPT_ACCEPT_ENCODING, "");
+    if (acceptEncoding_.has()) {
```
COM8, line 806: Shouldn't this be:
```suggestion
    if (!acceptEncoding_.has()) {
```

cpr/accept_encoding.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+AcceptEncoding::AcceptEncoding(const std::initializer_list<std::string>& string_methods) : methods_{string_methods} {
+}
+
+bool AcceptEncoding::has() const {
+    return methods_.empty();
+}
+
+const std::string AcceptEncoding::getString() const {
+    std::string methodsString = std::accumulate(std::next(methods_.begin()), methods_.end(), methods_[0],
```
COM8, line 21: `std::accumulate` is CPP20, but cpr supports cpp11. Please replace it.

cpr/accept_encoding.cpp
```diff
@@ -0,0 +1,26 @@
+#include "cpr/accept_encoding.h"
+#include <algorithm>
+#include <numeric>
+
+namespace cpr {
+
+AcceptEncoding::AcceptEncoding(const std::initializer_list<AcceptEncodingMethods>& enum_methods) {
+    methods_.clear();
+    std::transform(enum_methods.begin(), enum_methods.end(), std::back_inserter(methods_),
```
COM8, line 9: `std::transform` is too new. It got introduced in cpp17.

COM8 (changes requested):
This one goes into the right direction. Only a few things to sort out. Also please rebase your changes on master: https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa

cpr/session.cpp
leviliangtw: Sorry for that XD
Now I changed it to empty()

include/cpr/accept_encoding.h
leviliangtw: Sure!

include/cpr/accept_encoding.h
leviliangtw: I would investigate it and find a best practice for that

test/abstractServer.cpp
leviliangtw: Got it!

cpr/accept_encoding.cpp
leviliangtw: I have investigated `std::accumulate` and `std::transform` already; however, the cppcheck throws the error messages to advise me to use that two methods. 
Could you give me some suggestions?

leviliangtw: BTW, `std::transform` is used in curl_container.cpp

leviliangtw: I tried another way to pass Cppcheck: wrap the raw loop into another function and then make them more readable. 
Reference: https://belaycpp.com/2021/06/22/dont-use-raw-loops/

COM8: Turns out I was wrong there. They both are available in cpp11. They just changed the signature in later versions and extended it. Sorry for the confusion. This means it can stay how it is.

COM8 (approved):

COM8:
msvc on Windows reports:
```
  D:\a\cpr\cpr\cpr\accept_encoding.cpp(8,24): error C2039: 'back_inserter': is not a member of 'std' [D:\a\cpr\cpr\build\cpr\cpr.vcxproj]
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.32.31326\include\vector(24): message : see declaration of 'std' [D:\a\cpr\cpr\build\cpr\cpr.vcxproj]
  D:\a\cpr\cpr\cpr\accept_encoding.cpp(8,37): error C3861: 'back_inserter': identifier not found [D:\a\cpr\cpr\build\cpr\cpr.vcxproj]
```

cpr/accept_encoding.cpp
leviliangtw: OK, I would like to keep it with `std::accumulate` and `std::transform` for simplicity.

COM8 (approved):
Looks good now, thanks!
Please create a PR for a bit of documentation here: https://github.com/libcpr/docs

COM8 (approved):

COM8:
D:\a\cpr\cpr\test\abstractServer.cpp(108,28): error C2061: syntax error: identifier 'uint' [D:\a\cpr\cpr\build\test\test_server.vcxproj]

leviliangtw:
> ```
> D:\a\cpr\cpr\test\abstractServer.cpp(108,28): error C2061: syntax error: identifier 'uint' [D:\a\cpr\cpr\build\test\test_server.vcxproj]
> ```

I think I should not modify this file.

PR #748, opened 2022-05-24
Support for CURLOPT_LOCALPORT and CURLOPT_LOCALPORTRANGE

jmhersc:
This adds support for the curl_easy_setopt options CURLOPT_LOCALPORT and CURLOPT_LOCALPORTRANGE.
Unfortunately there are no test cases as I was not able to build them, but I tried it out and captured the
communication with tcpdump and it looks like it works.

I just noticed that the new headers local_port.h and local_port_range.h are not yet being installed. Will update soon.


COM8 (changes requested):
Thanks for taking the time and working on this PR. Only a few things need to be done:
* Please add your new headers to `cpr.h`
* Please create a bit of documentation here: https://github.com/libcpr/docs
* Please add a few test cases. What you could do is write them locally, push your changes and then see if the CI fails.

COM8:
> Thank you for looking into this. Do you think it would make sense to add default constructors to cpr::LocalPort and cpr::LocalPortRange for setting the libcurl default values? Like in cpr::Interface?

In my eyes no, since I see no use case, where this would be useful. Once you have set the port range once, why would you want to reset it again for the same session - ignoring the fact that this would not have any effect on the session.

COM8 (changes requested):
Thank you. What you cloud do to test the already occupied port is, create on session, where you specify lest say port 12345 as local port. Then you create a second session with the same port which should fail.

jmhersc:
> Thank you. What you cloud do to test the already occupied port is, create on session, where you specify lest say port 12345 as local port. Then you create a second session with the same port which should fail.

I was thinking about just using the same port the http server is using, that should be occupied already. When using two session I'd have to deal with the situation that the first session already fails. If it is safe to assume the http server and the "local" session are using the same interface that should work?

COM8:
This is also possible.

jmhersc:
> This is also possible.

Ok. If there is anything else to do, please ask.

COM8:
Docs PR: https://github.com/libcpr/docs/pull/19#pullrequestreview-987216235

COM8 (approved):
Thanks for your PR. As soon as the CI runs through, we are ready to go!

jmhersc:
> Thanks for your PR. As soon as the CI run through, we are ready to go!

Thank you!

> Thanks for your PR. As soon as the CI run through, we are ready to go!

My pleasure!

PR #750, opened 2022-05-30
Add CaBuffer to enable loading of CA certificates from buffers

### LINKED ISSUE #483: Loading certificates from buffer
Hi,
I wanted to load certificates from a buffer instead of from a file, CURL seems to be able to do it, but I couldn't find any way to do it with CPR.
Is there any way to do this?

simon-berger:
Add the SSL option `CaBuffer` to allow loading of CA certificates directly from string buffers (closes #483 ).

The option is only enabled if OpenSSL is used as ssl backend and at least version 7.11.0 of libcurl is used.

The implementation uses [CURLOPT_SSL_CTX_FUNCTION](https://curl.se/libcurl/c/CURLOPT_SSL_CTX_FUNCTION.html) to specify the callback function and [CURLOPT_SSL_CTX_DATA](https://curl.se/libcurl/c/CURLOPT_SSL_CTX_DATA.html) to specify its parameters.


simon-berger:
I did some local tests today and as far as I was able to see, the loading of CA certificates from buffers works. 

At a later point in time, other SSL libraries could probably also be integrated (via another PR), but I do not have any examples or background knowledge for this.

Therefore, I would be happy about a review / feedback :D

cpr/session.cpp
```diff
@@ -558,6 +563,14 @@ void Session::Impl::SetSslOptions(const SslOptions& options) {
     if (!options.ca_path.empty()) {
         curl_easy_setopt(curl_->handle, CURLOPT_CAPATH, options.ca_path.c_str());
     }
+#if SUPPORT_CURLOPT_SSL_CTX_FUNCTION
+#ifdef OPENSSL_BACKEND_USED
```
COM8, line 567: In case the `OPENSSL_BACKEND_USED` is not used, please throw an error explaining why OpenSSL is required.
This might help:
https://gcc.gnu.org/onlinedocs/cpp/Diagnostics.html

COM8 (changes requested):
Looks good, thank you. Please add your new includes also to `cpr.hpp`.

cpr/session.cpp
simon-berger: I think I don't understand exactly what you mean.
As I understand it, it would look something like this:
```c++
#if SUPPORT_CURLOPT_SSL_CTX_FUNCTION
#ifdef OPENSSL_BACKEND_USED
...
#else
#error "CURLOPT_SSL_CTX_FUNCTION is only supported for OpenSSL"
#endif
```
However, this would mean that one could no longer compile CPR without OpenSSL.

simon-berger:
The docs are here: https://github.com/libcpr/docs/pull/20

cpr/session.cpp
COM8: Ah, did not fully get the `SUPPORT_CURLOPT_SSL_CTX_FUNCTION`. Then it is fine!

COM8 (approved):

Issue #745, opened 2022-05-17
Add support for uploading arbitrary file parts with CURLOPT_RESUME_FROM


simon-berger:
**Is your feature request related to a problem? Please describe.**
Currently it is not possible to upload arbitrary parts of a file, as [CURLOPT_RANGE](https://curl.se/libcurl/c/CURLOPT_RANGE.html) should not be used for HTTP PUT uploads, as it conflicts with other options. Instead [CURLOPT_RESUME_FROM](https://curl.se/libcurl/c/CURLOPT_RESUME_FROM.html) should be used together with [CURLOPT_INFILESIZE](https://curl.se/libcurl/c/CURLOPT_INFILESIZE.html).

**Describe the solution you'd like**
Implement the [CURLOPT_RESUME_FROM](https://curl.se/libcurl/c/CURLOPT_RESUME_FROM.html) option in order to support the upload of arbitrary file parts.

**Describe alternatives you've considered**
I am not aware of any alternatives.

**Additional context**
If you like this feature I can create a pull request for it.

COM8:
Yes, this would be a nice to have feature!

simon-berger:
Then I will work on it in the next few weeks and create a PR when I'm done

simon-berger:
I think we can close this issue, as libcurl / http will not support this (see: #753).

PR #753, opened 2022-06-03
Add support for uploading arbitrary file parts with CURLOPT_RESUME_FROM

simon-berger:
Issue: #745 

I have implemented the basics for uploading arbitrary file parts with CURLOPT_RESUME_FROM, like it is described [here](https://curl.se/mail/lib-2019-05/0012.html). 

Unfortunately, I have encountered a problem that I have not yet been able to solve on my own:
As implemented, all local tests are successful and header values are set correctly. However, I had to comment out the setting of the CURLOPT_RESUME_FROM option in `session.cpp`. If the option is set, requests with a set UploadRange cannot connect to the Http server. More precisely, the execution will then hang at `curl_easy_perform(curl_->handle)` in the `makeRequest()` method of `session.cpp`.

The problems occurred not only with the internal test http server, but also with a local Python server and external providers.

Unfortunately, I am currently stuck at this point and would be happy if someone could help me with this or has a suggestion what the problem might be. One of my assumptions is that the Http server simply does not support the upload of arbitrary parts and therefore the connect phase is somehow not successful.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -611,7 +611,8 @@ void Session::Impl::SetUploadRange(const UploadRange& upload_range) {
     curl_off_t filesize = upload_range.getFilesize();
 
     // Create upload range header
-    std::array<char, 256> range_header;
+    constexpr int max_header_char_count{256};
+    std::array<char, max_header_char_count> range_header{};
     sprintf(range_header.data(),
             "Content-Range: bytes %" CURL_FORMAT_CURL_OFF_T
             "-"
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -1005,6 +1005,7 @@ void Session::SetOption(const MultiRange& multi_range) { pimpl_->SetMultiRange(m
 void Session::SetOption(const ReserveSize& reserve_size) { pimpl_->SetReserveSize(reserve_size.size); }
 void Session::SetOption(const AcceptEncoding& accept_encoding) { pimpl_->SetAcceptEncoding(accept_encoding); }
 void Session::SetOption(AcceptEncoding&& accept_encoding) { pimpl_->SetAcceptEncoding(std::move(accept_encoding)); }
+void Session::SetOption(const UploadRange& upload_range) { pimpl_->SetUploadRange(upload_range); }
 
 cpr_off_t Session::GetDownloadFileLength() { return pimpl_->GetDownloadFileLength(); }
 void Session::ResponseStringReserve(size_t size) { pimpl_->ResponseStringReserve(size); }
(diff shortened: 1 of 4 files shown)


COM8:
I also think this might me a case of not supported. What happens in case you watch the events here: https://github.com/libcpr/cpr/blob/2157a3126c2ef1b76f339dedd6d7585e39651512/test/abstractServer.cpp#L26

Does it trigger any of those?
Did you try "Patch" instead of "Put"?
That's what I have found: https://stackoverflow.com/questions/20969331/standard-method-for-http-partial-upload-resume-upload

simon-berger:
Thanks for your reply @COM8. I will have a look at it next week 👍

simon-berger:
> I also think this might me a case of not supported. What happens in case you watch the events here:
> 
> https://github.com/libcpr/cpr/blob/2157a3126c2ef1b76f339dedd6d7585e39651512/test/abstractServer.cpp#L26
> 
> Does it trigger any of those?

Only the time event is triggered:
 https://github.com/libcpr/cpr/blob/2157a3126c2ef1b76f339dedd6d7585e39651512/test/abstractServer.cpp#L44


> Did you try "Patch" instead of "Put"?

Yes I tried ... same issue.

simon-berger:
I have done more tests and could not fix the problem for http. I think that although you can use the CURLOPT_RESUME_FROM option for http, very few servers support it.

The question now is how to handle the feature. One possibility would be to leave the CURLOPT_RESUME_FROM option commented out (delete it), and solve the partial uploads using only the "Content-Range" header and CURLOPT_INFILESIZE_LARGE. This would be sufficient in my opinion due to the lack of a specification for this case. Another alternative would be to set the option, but then I can't do any tests in the current state. Or discard the PR and leave the feature out.

What do you think about that @COM8?

COM8:
Since only `MG_EV_TIMER` gets triggered, I'm almost tempted to say this is a curl issue. Could you please have a look with wireshark, if curl even tries to establish a http connection?

simon-berger:
Of course. I will take a look.

COM8:
Then I might say, we have discovered a bug in curl.
What you can do now is, try updating curl: https://github.com/libcpr/cpr/blob/dd715f6aa9d2692d31712a9ac5662ebc943298f2/CMakeLists.txt#L246-L250

If this also does not help, please create an issue for this at curl.

simon-berger:
After discussion with the curl team, one should probably no longer use CURLOPT_RESUME_FROM for http requests. Unfortunately, the corresponding curl documentation suggested this, which led to my confusion.

If we still want to implement the partial uploads feature, I think it would be a possibility to set only the CURLOPT_INFILESIZE and instead of CURLOPT_RESUME_FROM set the content-range header manually (like it is already done in the current implementation). Since there is no http standard for this, and we can document it clearly, this would save a potential user at least a few steps.
Or do we not want to introduce such a feature at the current time, discard the PR and eventually wait for a better support / working example implementation of curl?

COM8:
In this case I'd say we should discard this PR since I don't see a clear use case for that. In case someone actually needs this, he then can start working on a new PR with this as a reference.
But thanks for looking into this!

simon-berger:
Ok, thanks for the response. I close the PR.

PR #756, opened 2022-06-08
Add async methods to the Session object interface

### LINKED ISSUE #70: Interface of Session object doesn't provide Async methods
Currently it is not possible to call *Async or *Callback methods on the session object. I believe that this, combined with issue #69, makes it impossible to do write async code in combination with SSL keepalive in a practical way.

simon-berger:
Close #70 

## Async Requests
Implemented the following async request methods for the session interface:
```c++
AsyncResponse GetAsync();
AsyncResponse DeleteAsync();
AsyncResponse DownloadAsync(const WriteCallback& write);
AsyncResponse DownloadAsync(std::ofstream& file);
AsyncResponse HeadAsync();
AsyncResponse OptionsAsync();
AsyncResponse PatchAsync();
AsyncResponse PostAsync();
AsyncResponse PutAsync();
```
These functions are internally based on the previous async implementation and ensure that the Session::Impl object is not deleted during use via a shared pointer in the lambda capture.
Note: The async request is directly made on the passed Session object and not on a copy of it.

## Async Callbacks
Furthermore I implemented a draft for the async callback function `GetCallback` on Session in the last [commit](https://github.com/libcpr/cpr/commit/eba14ceea3c67cfd0df820086585662e078d32fa). However, for this I had to move the definition of the Session::Impl class from session.cpp into session. This is neccessary, as the required template functions for `GetCallback` have to be implemented directly in the header. Since these template functions have to access functions from Session::Impl, I couldn't figure out another way with C++11.

If someone has a better idea how to implement the callbacks without moving the Session::Impl definition I will be happy to implement it.

If there is no alternative idea, there are two options I guess:
1. If moving Session::Impl is not an option we want to merge, I will revert the corresponding commit and finish the PR just for the AsyncRequest methods.
2. If the current implementation is ok, I will extend the PR with the other callback methods and finalize it.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -86,12 +86,6 @@ Session::Session() : curl_(new CurlHolder()) {
 #endif
 }
 
-Session::Session(Session&& /*old*/) = default;
-
-Session::~Session() = default;
-
-Session& Session::operator=(Session&& old) noexcept = default;
-
 Response Session::makeDownloadRequest() {
     assert(curl_->handle);
 
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -11,6 +11,7 @@
 
 #include <curl/curl.h>
 
+#include "cpr/async.h"
 #include "cpr/cprtypes.h"
 #include "cpr/interceptor.h"
 #include "cpr/util.h"
@@ -28,7 +29,7 @@ constexpr long ON = 1L;
 // NOLINTNEXTLINE(google-runtime-int)
 constexpr long OFF = 0L;
 
-class Session::Impl {
+class Session::Impl : public std::enable_shared_from_this<Session::Impl> {
   public:
     explicit Impl(Session* psession);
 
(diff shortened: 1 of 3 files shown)


COM8:
Thanks for having a look at this. Currently I do not see a specific reason why moving `Session::Impl` to the header is too bad. But before I give my final OK on that, I'd like to have a deeper look at this first. Sadly I won't have time for this until next Monday.

COM8:
In my eyes we can remove `Session::Impl` completely. It's wrapper functionality is not needed in my eyes any more. This the would allow us to reduce code duplication and we do not have to move it into the header file.
Besides that, the way you implemented `GetCallback` looks good although I have not too much experience with `std::future`.

@KingKili Do you have more experience with that?

simon-berger:
As far as I understand the project structure, I also don't think that Session::Impl is needed.
I could accordingly remove Session::Impl right away with this PR or, if that would be too many changes, I could also create another PR in which I remove Impl and continue working on this PR after a rebase.

COM8:
Yes, please remove it in this PR since in my eyes this is a part of your change.

COM8:
Please rebase on master. Then I will have a look at it.

simon-berger:
Rebase completed

COM8 (approved):
Looks good. Thanks.

tlammi:
What is the rationale behind deleting Session move constructor? The commit message only states what but not why.

simon-berger:
If I remember correctly, the compiler could not generate the move constructor after integrating Session::Impl. If you want to know it more precisely, I would have to look at it again in detail.

Issue #752, opened 2022-06-01
cpr.h includes windows.h on Windows (again)


cschreib-ibex:
See https://github.com/libcpr/cpr/issues/700 for the first bug report on this issue. It was fixed for a while, but unfortunately https://github.com/libcpr/cpr/commit/c0a412230be7ce1fced4b8c3a6fb2b8b9f7dd746 introduced a new inclusion of curl.h, so the problem appeared again. 

Here's the chain:
 - `cpr/cpr.h` -> `cpr/cookies.h` -> `cpr/curlholder.h` -> `curl/curl.h`

## Description
See https://github.com/libcpr/cpr/issues/700.

## Expected Behavior
See https://github.com/libcpr/cpr/issues/700.

## Actual Behavior
See https://github.com/libcpr/cpr/issues/700.

## Possible Fix
See https://github.com/libcpr/cpr/issues/700.

## Steps to Reproduce
See https://github.com/libcpr/cpr/issues/700.

## Context
See https://github.com/libcpr/cpr/issues/700.

## Your Environment
- Version used: 1.8.1
- Where did you get it from (e.g. conan, vcpkg, master, ...): conan
- Operating System and version: Windows 10 Pro
- Link to a small example:
```c++
void CopyFile() {  /*...*/ }

#include "cpr/cpr.h"

int main() {
    CopyFile(); // compiler error
}
```

COM8:
Thanks for the heads up on this one. I will have a look at it shortly!

COM8:
I don't think this is something we can fix if you include `cpr/cpr.h` since we require the `curl.h` include inside `cpr/curlholder.h` for the `CURL` type, which is not being able to be forward declared.

I have created a PR to reduce the number of transitive curl.h includes in cpr here: https://github.com/libcpr/cpr/pull/758
Could you please try this branch and do not include `cpr/cpr.h`, but instead try `cpr/api.h`.

COM8:
Ping @cschreib-ibex

cschreib-ibex:
Sorry, I was on honeymoon! I just gave it a go, and it seems there is still an issue:
 - `cpr/api.h` -> `cpr/response.h` -> `cpr/ssl_options.h` -> `curl/curl.h`
   It looks like the include to `cpr/ssl_options.h` can just be removed from `response.h`.

 - `cpr/api.h` -> `cpr/session.h` -> `cpr/ssl_options.h` -> `curl/curl.h`
[...]

PR #758, opened 2022-06-08
Reduced the number of curl.h includes

COM8:
This PR reduces the number of transitive `curl.h` includes in cpr for https://github.com/libcpr/cpr/issues/752


COM8:
Closing this one for now since it does not seam to be "relevant" any more. Make noise if it still is.

Issue #581, opened 2021-06-17
Canot use File or Buffer convienince classes in a Post Body


RoryO:
This is a continuation of #212 . That issue states to use an http form. However, when working with APIs there is a use case for a post body with raw binary data, specified by `application/octet-stream`. For example, the [Microsoft Cognitive Services](https://westcentralus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-ga/operations/56f91f2e778daf14a499f21b) states:

> Request body
> Input passed within the POST body. Supported input methods: raw image binary or image URL. 

Explicitly stating that the post body is either an image url, when using `application/json` or the raw image data when using `application/octet-stream`. In this case it would be convenient to do

```c++
Response r = Post(Url(base_url),
                  Header({
                    {"Ocp-Apim-Subscription-Key", api_key},
                    {"Content-Type", "application/octet-stream"}
                  }),
                  Body(File("test.jpg")));
```
However, the template for Post does not accept a File or a Buffer. Instead, you have to read the file into memory manually and use the pointer and size in the Body constructor. This works perfectly fine

```
char *data = 0;
int handle = 0;
handle = open("test.jpg", O_RDONLY);
assert(handle != 0);
struct stat s = {};
fstat(handle, &s);
size = s.st_size;
data = static_cast<char *>(malloc(s.st_size));
read(handle, data, s.st_size);
assert(data);
close(handle);

Response r = Post(Url(base_url),
                  Header({
                    {"Ocp-Apim-Subscription-Key", api_key},
                    {"Content-Type", "application/octet-stream"}
                  }),
                  Body(data, s.st_size));

```

Since passing a raw pointer to the Body constructor works in this case, and there are real world use cases for raw binary data in a post request, I suggest the template to Body can also accept the File and Buffer convenience classes like Multipart does.

COM8:
Yes, sure. Sounds reasonable.
Would you like to create a PR for this?

RoryO:
Sure, I’ll take a stab at it.

leviliangtw:
I'd like to take this issue.

KingKili:
Sure go ahead. Just let me know when PR is ready. Thanks in advance!

leviliangtw:
@KingKili @COM8 I created a new PR[#763](https://github.com/libcpr/cpr/pull/763).

PR #763, opened 2022-06-17
Add support of file and buffer for the POST Body (#581)

leviliangtw:
1. Make struct Buffer and File independent to Multipart. 
2. Add support of struct Buffer and File to class Body. 
3. Modify CMakeLists.txt.

### DIFF
--- a/include/cpr/buffer.h
+++ b/include/cpr/buffer.h
@@ -1,0 +1,33 @@
+#ifndef CPR_BUFFER_H
+#define CPR_BUFFER_H
+
+#include <string>
+
+namespace cpr {
+
+struct Buffer {
+    using data_t = const unsigned char*;
+
+    template <typename Iterator>
+    Buffer(Iterator begin, Iterator end, std::string&& p_filename)
+            // Ignored here since libcurl reqires a long.
+            // There is also no way around the reinterpret_cast.
+            // NOLINTNEXTLINE(google-runtime-int, cppcoreguidelines-pro-type-reinterpret-cast)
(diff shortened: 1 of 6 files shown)


include/cpr/buffer.h
```diff
@@ -0,0 +1,33 @@
...
+    using data_t = const unsigned char*;
+
+    template <typename Iterator>
+    Buffer(Iterator begin, Iterator end, std::string&& p_filename)
+            // Ignored here since libcurl reqires a long.
+            // There is also no way around the reinterpret_cast.
+            // NOLINTNEXTLINE(google-runtime-int, cppcoreguidelines-pro-type-reinterpret-cast)
+            : data{reinterpret_cast<data_t>(&(*begin))}, datalen{static_cast<long>(std::distance(begin, end))}, filename(std::move(p_filename)) {
+        is_random_access_iterator(begin, end);
+        static_assert(sizeof(*begin) == 1, "only byte buffers can be used");
```
COM8, line 18:
```suggestion
        static_assert(sizeof(*begin) == 1, "Only byte buffers can be used.");
```

COM8 (changes requested):
Looks good. Thanks! Only two minor changes.

leviliangtw:
@COM8 Some tests of CI cannot be passed due to SSL problems, I have no idea what it is. 
However, I think they are not relative to what I modified in my code.

COM8 (approved):
Looks good now. Thanks.
The Windows/MacOS CI is currently broken. Somehow installing OpenSSL fails. This is something I have to create an issue for.

Issue #216, opened 2017-09-13
Multi-part file uploading via buffer does not work properly.


ninearif:
According to this feature  "Add support for POST file uploads from in-memory buffers " https://github.com/whoshuu/cpr/pull/129.

This support will allow you upload data from the buffer.
so when we do something like :

```
(1)
std::vector<unsigned char> content{'h', 'e', 'l', 'l', 'o', ' ', 'w', 'o', 'r', 'l', 'd'};
auto url = Url{"https://httpbin.org/post"};
auto response = cpr::Post(url, Multipart{
          { "fileContent", 
             Buffer{content.begin(), content.end(), "test_file.txt"}
          }});
```
 
and  :

```
(2)
std::vector<unsigned char> content{'h', 'e', 'l', 'l', 'o', ' ', 'w', 'o', 'r', 'l', 'd'};
auto url = Url{"https://httpbin.org/post"};
auto buff = Buffer{content.begin(), content.end(), "test_file.txt"};
auto response = cpr::Post(url, Multipart{
          { "fileContent", 
             buff
          }});
```

ps: I use httpbin.org to check my request details.

The expected behavior of POST request from both (1) and (2) should give the same result. They should be sent **fileContent** as a **file** not a **form** but only a code in (2) is work.The problem of (1) is  **fileContent** was sent as a **form** (It should be a **file**).

I'm not sure, I think this code might be related to this issue.

https://github.com/whoshuu/cpr/blob/b5758fbc88021437f968fe5174f121b8b92f5d5c/cpr/session.cpp#L29-L30
https://github.com/whoshuu/cpr/blob/b5758fbc88021437f968fe5174f121b8b92f5d5c/cpr/session.cpp#L184
https://github.com/whoshuu/cpr/blob/b5758fbc88021437f968fe5174f121b8b92f5d5c/cpr/session.cpp#L217

ninearif:
Here are request details that show a different result between code (1) and (2).
```
(1) Result:
{
  "args": {}, 
  "data": "", 
  "files": {
  }, 
  "form": {
    "fileContent": "....DATA....", 
  }, 
  "headers": {...}, 
  "json": null, 
  "origin": "...", 
  "url": "http://www.httpbin.org/post"
}
```

```
(2) Result:
{
  "args": {}, 
  "data": "", 
  "files": {
    "fileContent": "....DATA....", 
  }, 
  "form": {
  }, 
  "headers": {...}, 
  "json": null, 
  "origin": "...", 
[...]

PR #764, opened 2022-06-17
Add unit tests for file uploading using buffer of rvalue/lvalue reference (#216)

leviliangtw:
1. Use the `OnRequestFormPost` handler of the mock HTTP server to judge whether it uploads a file or not. 
    - `httpserver.cpp`:
        ```c++=422
        if (!filenames["x"].empty()) {
            response = std::string{
                    "{\n"
                    "  \"x\": " +
                    filenames["x"] + "=" + forms["x"] +
                    "\n"
                    "}"};
        }
        ```
2. Write two unit tests using struct `Buffer` of rvalue and lvalue reference.


COM8 (approved):
Thanks

PR #765, opened 2022-06-20
C++17 support

simon-berger:
# Work in progress
**Do not merge**

# Issue
- Link: #382 
- Update CPR to C++17
- Make use of new standard features and improve the implementation

# Planned enhancements / Progress

- [ ] Implement [std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) support (e.g. for cpr::Url as proposed by @[HellsingDarge](https://github.com/HellsingDarge))
- [ ] Improve lambda functions:
  - [ ] Type determination with auto
  - [ ] Initialization of function-local constants directly in captures
- [ ] Use of [std::optional](https://en.cppreference.com/w/cpp/utility/optional) (e.g. for ranges which are only partially defined)
- [ ] Use the new [filesystem](https://en.cppreference.com/w/cpp/filesystem) library
- [ ] Use [std::byte](https://en.cppreference.com/w/cpp/types/byte) instead of char / unsigned char
- [ ] Support [std::chrono_literals](https://en.cppreference.com/w/cpp/chrono/operator%22%22h) for timeouts
- [ ] Use auto as return type for functions without decltype
- [ ] Use of [make_unique()](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique)
- [ ] Investigate lambda capture of *this (might be interesting for async requests)

# Optional improvements
I will not implement the following list until I receive confirmation:
- Replace `const std::string&` with `std::string_view` in (every) file
  - In my opinion this would be more performant and the more modern code. The only drawback I see is that std::string_view has no c_str() method. This would mean that in all places where a C string is needed from a string_view, a string would first have to be created from it.
  - Performance analysis: [link](https://www.cppstories.com/2018/07/string-view-perf/)


simon-berger:
Docs: https://github.com/libcpr/docs/pull/24

simon-berger:
I tried to add a new job for Ubuntu 20.10 to check compiling with a newer GCC version. However, no workflow was triggered by my push. Is this because of the conflicts with the master branch or the pending reviews, or did I make a mistake when creating the job?

COM8:
I'd say this is because of the conflict. But in your forked repo you can also enable GitHub Actions via the tab on top.

include/cpr/async.h
```diff
@@ -19,18 +19,17 @@ class GlobalThreadPool : public ThreadPool {
...
 
 class async {
   public:
-    static void startup(size_t min_threads = CPR_DEFAULT_THREAD_POOL_MIN_THREAD_NUM,
-                 size_t max_threads = CPR_DEFAULT_THREAD_POOL_MAX_THREAD_NUM,
-                 std::chrono::milliseconds max_idle_ms = CPR_DEFAULT_THREAD_POOL_MAX_IDLE_TIME) {
+    static void startup(size_t min_threads = CPR_DEFAULT_THREAD_POOL_MIN_THREAD_NUM, size_t max_threads = CPR_DEFAULT_THREAD_POOL_MAX_THREAD_NUM, std::chrono::milliseconds max_idle_ms = CPR_DEFAULT_THREAD_POOL_MAX_IDLE_TIME) {
         GlobalThreadPool* gtp = GlobalThreadPool::GetInstance();
-        if (gtp->IsStarted()) return;
+        if (gtp->IsStarted())
```
COM8, line 31: Please always use braces around even single line ifs to prevent unexpected behaviour.

COM8 (changes requested):
Looks good. This is basically what I intended it to be.

KingKili (changes requested):
This is what I found so far. But I looked only very coarsely.

COM8:
A small not from my side regarding the rough time line when this PR is expected to be merged: I'm planing to release 1.9.0 by the end of next week. Then this PR will be the first one that gets merged after that.

include/cpr/async.h
simon-berger: I have only formatted this line automatically. Should I still add the brackets @COM8?

CMakeLists.txt
```diff
@@ -10,11 +12,41 @@ if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
...
+        CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX17)
+        CHECK_CXX_COMPILER_FLAG("-std=c++20" COMPILER_SUPPORTS_CXX20)
+    endif()
+    if(COMPILER_SUPPORTS_CXX20)
+        set(CMAKE_CXX_STANDARD 20)
+    elseif(COMPILER_SUPPORTS_CXX17)
+        set(CMAKE_CXX_STANDARD 17)
+    else()
+        message(FATAL_ERROR "Compiler ${CMAKE_CXX_COMPILER} does not support C++17. Please use cpr <= 1.9.0")        
+    endif()
```
KingKili, line 46: Why all this logic. CPR shoult not set CXX standard to 20, if it is not set by the parent project.

CMakeLists.txt
```diff
@@ -10,11 +12,41 @@ if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
...
 endif()
 
 # Avoid the dll boilerplate code for windows
 set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-set(CMAKE_CXX_STANDARD 11)
+
+if (PARENT_CXX_STANDARD)
+    # Don't set CMAKE_CXX_STANDARD if it is already set by parent project
+    if (${PARENT_CXX_STANDARD} LESS 17)
```
KingKili, line 28: I think omitting the `${}` here should be fine

CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+include(CheckCXXCompilerFlag)
```
KingKili, line 1: If we remove all the 17/20 logic, we should not require thies include

CMakeLists.txt
```diff
@@ -10,11 +12,41 @@ if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
     set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "CMake")
     set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR}/bin)
     set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR}/lib)
+else()
+    # Check required c++ standard of parent project
```
KingKili, line 16: I think we should also set the [CXX_STANDARD_REQUIRED](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED) variable

KingKili (changes requested):

CMakeLists.txt
simon-berger: I thought the plan for cpr 1.10.0 was to use C++20 if supported and make C++17 required (like it is described here: https://github.com/libcpr/cpr/issues/382#issue-621811066)?

If I have misunderstood and this is not desired I can of course remove the additional logic.

CMakeLists.txt
simon-berger: I will try this, thanks :)

CMakeLists.txt
simon-berger: It is set here: https://github.com/simon-berger/cpr/blob/5205468e7f087ef992f4f1ddab99514968b7ac48/CMakeLists.txt#L50
```
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

CMakeLists.txt
COM8: Yes, this was the plan since not all compilers are up to standard with cpp20.

include/cpr/async.h
COM8: Yes, please.

CMakeLists.txt
KingKili: Sorry, I oversaw that one

CMakeLists.txt
simon-berger: Ok, no problem :) I will look into this later and fix it.
Just one question: Should there still be a check with `CHECK_CXX_COMPILER_FLAG` if c++17 is supported. Otherwise I don't see an option how to raise an error message in the case that c++17 is not supported.

KingKili: There is no need for this, since CMake will should check this for us

include/cpr/filesystem.h
```diff
@@ -0,0 +1,15 @@
...
+#include <filesystem>
+namespace fs = std::filesystem;
+#elif __has_include("experimental/filesystem")
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
+#error "Failed to include <filesystem> header!"
+#endif
+
+#endif
```
KingKili, line 15: I would prefer to not have this header. But I can't think of any neater solution, … so it is fine

KingKili (approved):
All in all I am happy with your PR. Thanks for the great work!

include/cpr/filesystem.h
simon-berger: Yeah, me neither. But I think it is the best solution for now.

simon-berger:
Rebase finished!

CMakeLists.txt
```diff
@@ -10,11 +10,28 @@ if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
...
 endif()
 
 # Avoid the dll boilerplate code for windows
 set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
-set(CMAKE_CXX_STANDARD 11)
+
+if (PARENT_CXX_STANDARD)
+    # Don't set CMAKE_CXX_STANDARD if it is already set by parent project
+    if (PARENT_CXX_STANDARD LESS 17)
+        message(FATAL_ERROR "cpr ${cpr_VERSION} does not support ${PARENT_CXX_STANDARD}. Please use cpr <= 1.9.0")
```
COM8, line 27:
```suggestion
        message(FATAL_ERROR "cpr ${cpr_VERSION} does not support ${PARENT_CXX_STANDARD}. Please use cpr <= 1.9.x")
```

COM8 (changes requested):
Only one small "typo". Other than that I'm also really happy with the result.

COM8 (approved):

COM8:
OK, we are ready to. I will merge it in the next couple of days when preparing the Repo for the next release as first PR.

include/cpr/timeout.h
```diff
@@ -12,6 +12,8 @@ class Timeout {
     Timeout(const std::chrono::milliseconds& duration) : ms{duration} {}
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
     Timeout(const std::int32_t& milliseconds) : Timeout{std::chrono::milliseconds(milliseconds)} {}
+    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
+    Timeout(const std::chrono::seconds& duration) : ms{1000 * duration.count()} {}
```
Curve, line 16: Alternatively: `ms{std::chrono::milliseconds(duration).count()}`

COM8: Agree. Fixed. Thanks!
https://github.com/libcpr/cpr/commit/2694e2badb84afef791bcaba8a7b0d3b1ab80c15

PR #770, opened 2022-06-28
Implement curl_multi_perform

### LINKED ISSUE #556: performance improvement curl_multi_perform and multiplexing
libcurl support `curl_multi_perform `for asynchronous requests and also offers `curl_multi_fdset` to extract **fd_sets** to handle concurrent requests efficiently. As well as libcurl support **multiplexing and http/2** protocol to reusing a single connection to one server for multiple http requests to improve **performance** of requests-heavy applications. 
I am not able to find some thing like this in cpr or may be I missed that. Please inform if cpr is providing such functionality or cpr not providing it could be good new feature.

simon-berger:
# Progress
- [x] Implement more request methods
- [x] Implement interface which does implicitly create session objects
- [x] Tests
- [x] Documentation

# Issue
- Closes #556 

# Implementation

## Class: CurlMultiHolder
- `cpr::CurlMultiHolder` is similar to `cpr::CurlHolder` and is a RAII wrapper for the used `CURLM` handle which is created using [curl_multi_init](https://curl.se/libcurl/c/curl_multi_init.html)

## Class: MultiPerform
- Implemented new class `cpr::MultiPerform` with the following public interface:
  ```c++
  class MultiPerform {
    public:
      enum class HttpMethod {
        UNDEFINED = 0,
        GET_REQUEST,
        POST_REQUEST,
        PUT_REQUEST,
        DELETE_REQUEST,
        PATCH_REQUEST,
        HEAD_REQUEST,
        OPTIONS_REQUEST,
        DOWNLOAD_REQUEST,
      };
      
      MultiPerform();
      std::vector<Response> Get();
      std::vector<Response> Delete();
      std::vector<Response> Put();
      std::vector<Response> Head();
      std::vector<Response> Options();
      std::vector<Response> Patch();
      std::vector<Response> Post();
      template <typename... DownloadArgTypes>
      std::vector<Response> Download(DownloadArgTypes... args);

      std::vector<Response> Perform();
      template <typename... DownloadArgTypes>
      std::vector<Response> PerformDownload(DownloadArgTypes... args);

      void AddSession(std::shared_ptr<Session>& session, HttpMethod method = HttpMethod::UNDEFINED);
      void RemoveSession(const std::shared_ptr<Session>& session);
  }
  ```
  - `AddSession` and `RemoveSession` can be used to add and remove sessions from the multi perform handle which also locks / unlocks the session for the standalone use without multi handle.
  - The `Perform` function is only interesting when using `AddSession` with the method parameter. If the method parameter is set when adding the respective session, `Perform` will perform the specified HTTP method on the session object. This makes it possible to execute several different request methods in one Multi Perform.
  Note: One can not mix `DOWNLOAD_REQUEST` with any other request type.
  - The functions `Get`, `Delete`, `Put`, `Head`, `Options`, `Patch`, `Post`, and `Download` can be used to perform the respective http request on every added session
- Internally, every `cpr::MultiPerform`  object holds a vector of pairs of added session objects together with their specified request type and a `cpr::CurlMultiHolder`
  ```c++
  std::vector<std::pair<std::shared_ptr<Session>, HttpMethod>> sessions_;
  std::unique_ptr<CurlMultiHolder> multicurl_;
  ```
- When a request is performed the following steps are done:
  1. Call prepare function (e.g. PrepareGet()) on every session object
  2. Do multi perform using [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) and [curl_multi_poll](https://curl.se/libcurl/c/curl_multi_poll.html) until every handle has finished
  3. Get infos and create Response objects for all requests using [curl_multi_info_read](https://curl.se/libcurl/c/curl_multi_info_read.html)

## API Functions for Simplified Usage
- Implemented the following functions to simplify the usage for base cases:
  - `cpr::MultiGet`
  - `cpr::MultiDelete`
  - `cpr::MultiPut`
  - `cpr::MultiHead`
  - `cpr::MultiOptions`
  - `cpr::MultiPatch`
  - `cpr::MultiPost`
- All these functions have the same structure and for example `cpr::MultiGet(...)` can be used to implicitly create a `cpr::CurlMultiHolder` object which automatically creates `cpr::Session` objects, adds them to the multi holder and performs a multi get request on them. In order to set different options for different sessions, one can simply pass multiple tuples (the number of tuples determines how many sessions will be created) containing the session options.

**Example:**
The following request will create two session objects and performs a get request to 'https://example.com'. Moreover, for the first request a timeout is set.
```c++
std::vector<Response> responses = MultiGet(std::tuple<Url, Timeout>{Url{"https://example.com"}, Timeout{1000}}, std::tuple<Url>{Url{"https://example.com"}});

```

### DIFF
--- a/cpr/multiperform.cpp
+++ b/cpr/multiperform.cpp
@@ -46,7 +46,7 @@ void MultiPerform::DoMultiPerform() {
         }
 
         if (still_running) {
-            int timeout_ms = 1000;
+            const int timeout_ms{1000};
             error_code = curl_multi_poll(multicurl_->handle, nullptr, 0, timeout_ms, nullptr);
             if (error_code) {
                 fprintf(stderr, "curl_multi_poll() failed, code %d.\n", static_cast<int>(error_code));
--- a/cpr/multiperform.cpp
+++ b/cpr/multiperform.cpp
@@ -14,6 +14,9 @@ void MultiPerform::AddSession(std::shared_ptr<Session>& session) {
         return;
     }
 
+    // Lock session to the multihandle
+    session->isUsedInMultiPerform = true;
+
     // Add session to sessions_
     sessions_.push_back(session);
 }
--- a/cpr/multiperform.cpp
+++ b/cpr/multiperform.cpp
@@ -27,7 +27,7 @@ void MultiPerform::AddSession(std::shared_ptr<Session>& session, HttpMethod meth
     // Add easy handle to multi handle
     CURLMcode error_code = curl_multi_add_handle(multicurl_->handle, session->curl_->handle);
     if (error_code) {
-        fprintf(stderr, "curl_multi_add_handle() failed, code %d.\n", static_cast<int>(error_code));
+        (void) fprintf(stderr, "curl_multi_add_handle() failed, code %d.\n", static_cast<int>(error_code));
         return;
     }
 
@@ -42,7 +42,7 @@ void MultiPerform::RemoveSession(const std::shared_ptr<Session>& session) {
     // Remove easy handle from multihandle
     CURLMcode error_code = curl_multi_remove_handle(multicurl_->handle, session->curl_->handle);
     if (error_code) {
-        fprintf(stderr, "curl_multi_remove_handle() failed, code %d.\n", static_cast<int>(error_code));
+        (void) fprintf(stderr, "curl_multi_remove_handle() failed, code %d.\n", static_cast<int>(error_code));
         return;
     }
 
@@ -68,7 +68,7 @@ void MultiPerform::DoMultiPerform() {
     do {
         CURLMcode error_code = curl_multi_perform(multicurl_->handle, &still_running);
         if (error_code) {
-            fprintf(stderr, "curl_multi_perform() failed, code %d.\n", static_cast<int>(error_code));
+            (void) fprintf(stderr, "curl_multi_perform() failed, code %d.\n", static_cast<int>(error_code));
             break;
         }
 
@@ -76,7 +76,7 @@ void MultiPerform::DoMultiPerform() {
             const int timeout_ms{1000};
             error_code = curl_multi_poll(multicurl_->handle, nullptr, 0, timeout_ms, nullptr);
             if (error_code) {
-                fprintf(stderr, "curl_multi_poll() failed, code %d.\n", static_cast<int>(error_code));
+                (void) fprintf(stderr, "curl_multi_poll() failed, code %d.\n", static_cast<int>(error_code));
                 break;
             }
         }
@@ -97,7 +97,7 @@ std::vector<Response> MultiPerform::ReadMultiInfo(std::function<Response(Session
             // Find current session
             auto it = std::find_if(sessions_.begin(), sessions_.end(), [&info](const std::pair<std::shared_ptr<Session>, HttpMethod>& pair) { return pair.first->curl_->handle == info->easy_handle; });
             if (it == sessions_.end()) {
-                fprintf(stderr, "Failed to find current session!\n");
+                (void) fprintf(stderr, "Failed to find current session!\n");
                 break;
             }
             std::shared_ptr<Session> current_session = (*it).first;
@@ -157,7 +157,7 @@ void MultiPerform::PrepareSessions() {
                 pair.first->PrepareOptions();
                 break;
             default:
-                fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or download without arguments!\n");
+                (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or download without arguments!\n");
                 return;
         }
     }
@@ -170,7 +170,7 @@ void MultiPerform::PrepareDownloadSession(size_t sessions_index, const WriteCall
             pair.first->PrepareDownload(write);
             break;
         default:
-            fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
+            (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
             return;
     }
 }
@@ -182,7 +182,7 @@ void MultiPerform::PrepareDownloadSession(size_t sessions_index, std::ofstream& 
             pair.first->PrepareDownload(file);
             break;
         default:
-            fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
+            (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
             return;
     }
 }
--- a/cpr/multiperform.cpp
+++ b/cpr/multiperform.cpp
@@ -89,7 +89,7 @@ std::vector<Response> MultiPerform::MakeRequest() {
 }
 
 void MultiPerform::PrepareGet() {
-    for (std::shared_ptr<Session> session : sessions_) {
+    for (std::shared_ptr<Session>& session : sessions_) {
         session->PrepareGet();
     }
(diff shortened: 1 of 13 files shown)


simon-berger:
I just noticed that now, as we are using C++17, I can remove the std14 namespace "hack" in api.h.
I will rebase to master and fix that.

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    // Set download only if neccessary
+    if (method == HttpMethod::DOWNLOAD_REQUEST) {
+        is_download_multi_perform = true;
+    }
+
+    // Add easy handle to multi handle
+    CURLMcode error_code = curl_multi_add_handle(multicurl_->handle, session->curl_->handle);
+    if (error_code) {
+        (void) fprintf(stderr, "curl_multi_add_handle() failed, code %d.\n", static_cast<int>(error_code));
```
COM8, line 31: Please use std::cerr here.

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    // Add session to sessions_
+    sessions_.emplace_back(session, method);
+}
+
+void MultiPerform::RemoveSession(const std::shared_ptr<Session>& session) {
+    // Remove easy handle from multihandle
+    CURLMcode error_code = curl_multi_remove_handle(multicurl_->handle, session->curl_->handle);
+    if (error_code) {
+        (void) fprintf(stderr, "curl_multi_remove_handle() failed, code %d.\n", static_cast<int>(error_code));
```
COM8, line 46: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+void MultiPerform::DoMultiPerform() {
+    // Do multi perform until every handle has finished
+    int still_running{0};
+    do {
+        CURLMcode error_code = curl_multi_perform(multicurl_->handle, &still_running);
+        if (error_code) {
+            (void) fprintf(stderr, "curl_multi_perform() failed, code %d.\n", static_cast<int>(error_code));
```
COM8, line 72: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+            break;
+        }
+
+        if (still_running) {
+            const int timeout_ms{1000};
+            error_code = curl_multi_poll(multicurl_->handle, nullptr, 0, timeout_ms, nullptr);
+            if (error_code) {
+                (void) fprintf(stderr, "curl_multi_poll() failed, code %d.\n", static_cast<int>(error_code));
```
COM8, line 80: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+    int still_running{0};
+    do {
+        CURLMcode error_code = curl_multi_perform(multicurl_->handle, &still_running);
+        if (error_code) {
+            (void) fprintf(stderr, "curl_multi_perform() failed, code %d.\n", static_cast<int>(error_code));
+            break;
+        }
+
+        if (still_running) {
+            const int timeout_ms{1000};
```
COM8, line 77: 1s might be a long timeout. This should rather be something like 250.

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+        int msgq = 0;
+
+        // Read info from multihandle
+        info = curl_multi_info_read(multicurl_->handle, &msgq);
+
+        if (info) {
+            // Find current session
+            auto it = std::find_if(sessions_.begin(), sessions_.end(), [&info](const std::pair<std::shared_ptr<Session>, HttpMethod>& pair) { return pair.first->curl_->handle == info->easy_handle; });
+            if (it == sessions_.end()) {
+                (void) fprintf(stderr, "Failed to find current session!\n");
```
COM8, line 101: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+                break;
+            }
+            std::shared_ptr<Session> current_session = (*it).first;
```
COM8, line 104:
```suggestion
            std::shared_ptr<Session> current_session = it->first;
```

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+                pair.first->PreparePatch();
+                break;
+            case HttpMethod::HEAD_REQUEST:
+                pair.first->PrepareHead();
+                break;
+            case HttpMethod::OPTIONS_REQUEST:
+                pair.first->PrepareOptions();
+                break;
+            default:
+                (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or download without arguments!\n");
```
COM8, line 161: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+void MultiPerform::PrepareDownloadSession(size_t sessions_index, const WriteCallback& write) {
+    std::pair<std::shared_ptr<Session>, HttpMethod>& pair = sessions_[sessions_index];
+    switch (pair.second) {
+        case HttpMethod::DOWNLOAD_REQUEST:
+            pair.first->PrepareDownload(write);
+            break;
+        default:
+            (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
```
COM8, line 174: std::cerr

cpr/multiperform.cpp
```diff
@@ -0,0 +1,272 @@
...
+}
+
+void MultiPerform::PrepareDownloadSession(size_t sessions_index, std::ofstream& file) {
+    std::pair<std::shared_ptr<Session>, HttpMethod>& pair = sessions_[sessions_index];
+    switch (pair.second) {
+        case HttpMethod::DOWNLOAD_REQUEST:
+            pair.first->PrepareDownload(file);
+            break;
+        default:
+            (void) fprintf(stderr, "PrepareSessions failed: Undefined HttpMethod or non download method with arguments!\n");
```
COM8, line 186: std::cerr

cpr/session.cpp
```diff
@@ -28,6 +28,14 @@ constexpr long ON = 1L;
 // NOLINTNEXTLINE(google-runtime-int)
 constexpr long OFF = 0L;
 
+CURLcode Session::DoEasyPerform() {
+    if (isUsedInMultiPerform) {
+        (void) fprintf(stderr, "curl_easy_perform cannot be executed if the CURL handle is used in a MultiPerform.\n");
```
COM8, line 33: std::cerr

include/cpr/api.h
```diff
@@ -51,6 +52,39 @@ void set_option(Session& session, Ts&&... ts) {
     set_option_internal<false, Ts...>(session, std::forward<Ts>(ts)...);
 }
 
+// Idea: https://stackoverflow.com/a/19060157
+template <typename Tuple, std::size_t... I>
+void apply_set_option_internal(Session& session, Tuple&& t, std::index_sequence<I...>) {
```
COM8, line 57: We use camel case for methods.

COM8 (changes requested):
Thanks! After some testing I have done, I agree with your changes. We need only a few formatting changes here and there.

include/cpr/api.h
simon-berger: I have orientated myself here with the notation on the other, already existing set_option functions. Should I change the notation anyway?

COM8: It looks like we are using lower case with '_' for template helper functions. For now this should then be fine, but in the future I will change this.

COM8 (approved):
Well done. Thank you!

Issue #771, opened 2022-06-28
HTTPS Tests are not working


saendigPhilip:
## Description
There are a lot of errors in the HTTPS Test Server. While @simon-berger made me aware of some of them, I encountered a few more during my work on updating the test server (see #757).

### Expected Behavior
The certificates and keys are moved to the test binary path, so the relative paths to them are correct. Certificates and keys are loaded properly.

### Actual Behavior
The [paths to the certificates and keys](https://github.com/libcpr/cpr/blob/master/test/ssl_tests.cpp#L115-L116) are wrong. For some reason, the tests succeed anyway.

### Possible Fix
Install the certificates and keys relative to the test build directory.

## Expired client certificate
The client certificate is expired and needs to be re-generated. 

### Possible Fix
Re-generate the client certificate. Also, the period of validity should be generously extended, as its private key is publicly available anyway.

### Expected Behavior
If I understood that correctly, there should be a PKI, in which the root certificate signs the server and the client certificate.

### Actual Behavior
The verification of the certificates failed. Openssl states that they are self-signed.
Verification check of the server certificate:
```console
$ openssl verify -CAfile ca.cer server.cer
C = XX, L = Default City, O = Default Company Ltd
error 18 at 0 depth lookup: self-signed certificate
error server.cer: verification failed
```
Verification check of the client certificate:
```console
$ openssl verify -CAfile ca.cer client.cer
C = XX, L = Default City, O = Default Company Ltd
error 18 at 0 depth lookup: self-signed certificate
C = XX, L = Default City, O = Default Company Ltd
error 10 at 0 depth lookup: certificate has expired
error client.cer: verification failed
```

### Possible Fix
Set up a CA as described in the [corresponding chapter in the OpenSSL cookbook](https://www.feistyduck.com/library/openssl-cookbook/online/openssl-command-line/private-ca.html).
This would not only solve the issue, but it would also make the generation of certificates for future tests easier.

## Context
As mentioned before, I am currently working on upgrading the mongoose HTTP test server to support mongoose 7. I need to fix those bugs anyway as they cause problems in mongoose 7. Therefore, I would like to work on that as well.

If anyone notices any further problems or has other suggestions for fixes, please mention them here, then I will have a look at them :)

COM8:
Thanks for reporting! Yes, this is an issue and should be fixed. Would you like to have a look at it, once you are done with your mongoose update?

saendigPhilip:
Yes, I'll try to solve it.
[...]

Issue #757, opened 2022-06-08
Update Mongoose to Version 7


COM8:
## Description

With the latest releases of OpenSSL we get a few deprecation warnings, when compiling [mongoose](https://github.com/cesanta/mongoose):
```log
[build] /home/fabian/Documents/Repos/libcpr/cpr/build/_deps/mongoose-src/mongoose.c: In function ‘mg_use_cert’:
[build] /home/fabian/Documents/Repos/libcpr/cpr/build/_deps/mongoose-src/mongoose.c:4758:7: warning: ‘PEM_read_bio_DHparams’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
[build]  4758 |       dh = PEM_read_bio_DHparams(bio, NULL, NULL, NULL);
[build]       |       ^~
[build] In file included from /usr/include/openssl/ssl.h:36,
[build]                  from /home/fabian/Documents/Repos/libcpr/cpr/build/_deps/mongoose-src/mongoose.c:4471:
[build] /usr/include/openssl/pem.h:469:1: note: declared here
[build]   469 | DECLARE_PEM_rw_attr(OSSL_DEPRECATEDIN_3_0, DHparams, DH)
[build]       | ^~~~~~~~~~~~~~~~~~~
[build] /home/fabian/Documents/Repos/libcpr/cpr/build/_deps/mongoose-src/mongoose.c:4767:7: warning: ‘PEM_read_bio_DHparams’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
[build]  4767 |       dh = PEM_read_bio_DHparams(bio, NULL, NULL, NULL);
[build]       |       ^~
[build] /usr/include/openssl/pem.h:469:1: note: declared here
[build]   469 | DECLARE_PEM_rw_attr(OSSL_DEPRECATEDIN_3_0, DHparams, DH)
[build]       | ^~~~~~~~~~~~~~~~~~~
[build] /home/fabian/Documents/Repos/libcpr/cpr/build/_deps/mongoose-src/mongoose.c:4773:7: warning: ‘DH_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
[build]  4773 |       DH_free(dh);
[build]       |       ^~~~~~~
[build] In file included from /usr/include/openssl/dsa.h:51,
[build]                  from /usr/include/openssl/x509.h:37,
[build]                  from /usr/include/openssl/ssl.h:31:
[...]

PR #773, opened 2022-06-29
New certificates for HTTPS tests

saendigPhilip:
## Issue
This pull request solves #771. The HTTPS tests now work with valid certificates and a working PKI.

### New Certificate Authority
* The new certificate hierarchy uses the `openssl ca` commandline tool for the generation of root, server, and client certificates.
* The certificates can be generated using the `generate-certificates.sh` script in `test/data` which uses the new configuration files `root-ca.cnf`, `server.cnf` and `client.cnf`.

### Changes in the certificates
* The client and server certificate are issued by the root certificate, such that `openssl verify -CAfile root-ca.crt server.crt client.crt` succeeds.
* The root certificate has a validity of 10 years. The client and server certificates both have a validity of 5 years.
* The elliptic curve ED25519 is used for public-key encryption, making the certificates smaller, the encryption faster, and everything more up-to-date.

### Adaptions in tests
* The certificates and keys are installed in the subdirectory `data` of the build directory.
* The paths for the certificates and keys in `ssl_tests.cpp` are updated.
* A test case is updated, which expects data from the new certificate.

## Conclusion
* Now, that we have working certificates, I will first of all continue my work on #757 to update mongoose to version 7.
* Possible follow-up: further HTTPS test cases (e.g. request a server with an invalid certificate, thus checking whether libcpr outputs an error in this case).


test/CMakeLists.txt
```diff
@@ -63,11 +63,6 @@ if (ENABLE_SSL_TESTS)
...
     
     # Install all ssl keys and certs:
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:ssl_tests>/data)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/server.pubkey $<TARGET_FILE_DIR:ssl_tests>)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/server.key $<TARGET_FILE_DIR:ssl_tests>)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/server.cer $<TARGET_FILE_DIR:ssl_tests>)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/client.key $<TARGET_FILE_DIR:ssl_tests>)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/client.cer $<TARGET_FILE_DIR:ssl_tests>)
-    add_custom_command(TARGET ssl_tests POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/data/ca.cer $<TARGET_FILE_DIR:ssl_tests>)
+    file(INSTALL data/keys DESTINATION data)
```
COM8, line 66: As far as I'm aware `file(...)` runs during the CMake configuration phase and not when a target is build or has been build.

This might result in issues when using CMake multi build.
@KingKili will test this to make sure everything works as expected.

test/ssl_tests.cpp
```diff
@@ -26,8 +31,7 @@ TEST(SslTests, HelloWorldTestSimpel) {
     std::this_thread::sleep_for(std::chrono::seconds(1));
 
     Url url{server->GetBaseUrl() + "/hello.html"};
-    std::string baseDirPath = server->getBaseDirPath();
-    SslOptions sslOpts = Ssl(ssl::CaPath{baseDirPath + "ca.cer"}, ssl::CertFile{baseDirPath + "client.cer"}, ssl::KeyFile{baseDirPath + "client.key"}, ssl::VerifyPeer{false}, ssl::PinnedPublicKey{baseDirPath + "server.pubkey"}, ssl::VerifyHost{false}, ssl::VerifyStatus{false});
+    SslOptions sslOpts = Ssl(ssl::CaPath{std::move(caCertPath)}, ssl::CertFile{std::move(clientCertPath)}, ssl::KeyFile{std::move(clientKeyPath)}, ssl::VerifyPeer{false}, ssl::PinnedPublicKey{std::move(serverPubKeyPath)}, ssl::VerifyHost{false}, ssl::VerifyStatus{false});
```
COM8, line 34: `std::move(...)` with static variables is no good code style.

COM8 (changes requested):
Thanks for looking into this. Only a few minor things to fix.

test/CMakeLists.txt
KingKili: This does not work with multi build. Please rework this

saendigPhilip: I resolved this by adding some custom post-build commands, similar to the way it was before

saendigPhilip:
The test output contains the following lines for each HTTPS test, which come from curl:
```
[...]
* Server certificate:
*  subject: CN=test-server
*  start date: Jun 29 11:33:07 2022 GMT
*  expire date: Jun 28 11:33:07 2027 GMT
*  issuer: C=GB; O=Example; CN=Root CA
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
[...]
```
What bothers me is the last line, which is an error code indicating that curl could not get the root-ca certificate, even though its path is specified in an argument to the method call.

Executing `curl -v --cacert certificates/root-ca.crt https://127.0.0.1:61937/hello.html` on the test server yields the following output:
```
[...]
* Server certificate:
*  subject: CN=test-server
*  start date: Jun 29 11:33:07 2022 GMT
*  expire date: Jun 28 11:33:07 2027 GMT
*  subjectAltName: host "127.0.0.1" matched cert's IP address!
*  issuer: C=GB; O=Example; CN=Root CA
*  SSL certificate verify ok.
[...]
```
So, this is not an issue with the certificates or with the test server, but rather either has something to do with curl or with the tests. Do you think that this needs some further investigation?

test/ssl_tests.cpp
saendigPhilip: I replaced all occurrences of std::move on static variables by string-concatenations for generating rvalue references in-place, similar to the way it was before. I removed the unused static variables that I introduced. This should resolve the issue

saendigPhilip:
And I have one more question: I rebased my work on the current main branch. This introduced some commits that don't belong to my work on this issue, as [shown by github](https://github.com/libcpr/cpr/pull/773/commits), for example "[Debian files and scripts for debian package build](https://github.com/libcpr/cpr/pull/773/commits/b8194fec561d819fcb2c929b2d0f0f8fc3f9eeac)".
Is this an issue for merging, and should I fix this?

COM8 (approved):
Looks good. Thanks!

COM8:
Regarding the `SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.` this is fine for now. Please create an issue for this, so we can look into this in the future.

Regarding the unwanted commits: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

PR #776, opened 2022-07-04
Securely remove sensitive data from memory

Garfield96:
This PR introduces explicit removal of sensitive data from memory. Once a string holding a password or certificate is destructed, the associated memory gets overridden. This reduces the time frame in which attackers can steal credentials, e.g. by examining a core dump or by exploiting a memory vulnerability.

### DIFF
--- a/cpr/util.cpp
+++ b/cpr/util.cpp
@@ -11,10 +11,10 @@
 #include <string>
 #include <vector>
 
-#if !defined(_Win32)
-#include <cstring>
-#else
+#if defined(_Win32)
 #include <Windows.h>
+#else
+#include <cstring>
 #endif
 
 namespace cpr {
@@ -127,7 +127,7 @@ int progressUserFunction(const ProgressCallback* progress, double dltotal, doubl
 int progressUserFunction(const ProgressCallback* progress, curl_off_t dltotal, curl_off_t dlnow, curl_off_t ultotal, curl_off_t ulnow) {
 #endif
     return (*progress)(dltotal, dlnow, ultotal, ulnow) ? 0 : 1;
-} // namespace util
+}
 
 int debugUserFunction(CURL* /*handle*/, curl_infotype type, char* data, size_t size, const DebugCallback* debug) {
     (*debug)(DebugCallback::InfoType(type), std::string(data, size));
@@ -174,7 +174,7 @@ void secureStringClear(std::string& s) {
 #if defined(__linux__) || defined(__unix__)
     explicit_bzero(&s.front(), s.length());
 #elif defined(_WIN32)
-    SecureZeroMemory(&s.front(), s.length())
+    SecureZeroMemory(&s.front(), s.length());
 #elif defined(__STDC_LIB_EXT1__)
     memset_s(&s.front(), s.length(), 0, s.length());
 #else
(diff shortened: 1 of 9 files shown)


cpr/util.cpp
```diff
@@ -11,6 +11,12 @@
 #include <string>
 #include <vector>
 
+#if !defined(_Win32)
```
COM8, line 14: Please invert this condition, so it's easier to read.

include/cpr/ssl_options.h
```diff
@@ -7,6 +7,7 @@
 
 #include <curl/curl.h>
 
+#include "util.h"
```
COM8, line 10:
```suggestion
#include "cpr/util.h"
```

COM8 (changes requested):
Thanks! Only a few minor things to fix.
Could you please also add this behaviour for the `Bearer` class?
https://github.com/libcpr/cpr/blob/67bf631a15424f20697ec27b009126c0ee5b4484/include/cpr/bearer.h#L30

COM8 (approved):
Looks good now. Thanks!

cpsauer:
We were just getting set up with CPR--and noticed that the pragmas in util.cpp introduced here lead to some build warnings, e.g., `external/cpr/cpr/util.cpp:188:9: warning: unknown pragma ignored [-Wunknown-pragmas]`. Definitely not a big deal--and I'm not sure if there's an easy way to fix that without `-Wno-unknown-pragmas`--but I figured you might want a heads up.

Thanks so much for a wonderful project!
Chris

cpsauer:
Will do. (We're wrapping the build with Bazel atm.)

Thanks for a great, fast reply--and what seems to be an unbelievably great library.

Apple Clang 13.1.6 in this particular instance, though we different ones by platform.

illesD:
With this PR we are losing support for Ubuntu16.
My jenkins/conan build:
```
...
14:53:55  [100%] Building CXX object cpr/CMakeFiles/cpr.dir/redirect.cpp.o
14:53:56  /home/conan/.conan/data/cpr/1.8.4/demo/dev/build/2baf5af34a86eeb65faf92ab36dd577663138661/cpr/cpr/util.cpp: In function ‘void cpr::util::secureStringClear(std::__cxx11::string&)’:
14:53:56  /home/conan/.conan/data/cpr/1.8.4/demo/dev/build/2baf5af34a86eeb65faf92ab36dd577663138661/cpr/cpr/util.cpp:179:42: error: ‘explicit_bzero’ was not declared in this scope
14:53:56       explicit_bzero(&s.front(), s.length());
14:53:56                                            ^
14:53:56  cpr/CMakeFiles/cpr.dir/build.make:270: recipe for target 'cpr/CMakeFiles/cpr.dir/util.cpp.o' failed
14:53:56  make[2]: *** [cpr/CMakeFiles/cpr.dir/util.cpp.o] Error 1
...
```

`explicit_bzero` requires **GLIBC 2.25** and higher, while Ubuntu16 comes with **2.23**.
https://man7.org/linux/man-pages/man3/bzero.3.html
```
illesd@jenkins:~$ ldd --version
ldd (Ubuntu GLIBC 2.23-0ubuntu11) 2.23
```
Could it be possible to replace it with `void bzero(void *s, size_t n);` or with `memset`?

It is crucial that our application supports U16 and above. I really don't want to be stuck on an older release without the option to update/apply fixes.

COM8:
Yes, this is definitely something we need to fix before the next release.

Issue #777, opened 2022-07-05
Refactor the class cpr::Cookies for storing more fields and reasonability


leviliangtw:
## Description
In this [comment](https://github.com/libcpr/cpr/issues/158), @COM8 mentioned a possibility of refactoring `cookies.h` with replacing the internal member `std::map<std::string, std::string>` to something like `std::map<std::string, cpr::Cookie>`. 
By doing that, you could manage the cookies more reasonably with a cookie consisting of kinds of fields instead of a pair string of only name and value. 

## Expected Behavior

```cpp
cpr::Response response = cpr::Get(url);
cpr::Cookies cookies{response.cookies};
for(const auto & cookie : cookies) {
    std::cout << cookie->Domain << ":" 
        << cookie->IncludingSubdomains << ":" 
        << cookie->Path << ":" 
        << cookie->HttpsOnly << ":" 
        << cookie->Expires << ":" 
        << cookie->Name << ":" 
        << cookie->Value << std::endl;
}
```

## Actual Behavior

```cpp
cpr::Response response = cpr::Get(url);
cpr::Cookies cookies{response.cookies};
for(const auto & cookie : cookies) {
    std::cout << cookie->first << ":" << cookie->second << std::endl;
}
```

## Possible Fix

After some investigations, it is much properer to refactor `std::map<std::string, std::string>` into `std::vector<cpr::Cookie>` by creating a new struct `cpr::Cookie` containing fields that `libcurl` returns. 

## References

- [RFC 6265 - HTTP State Management Mechanism](https://datatracker.ietf.org/doc/html/rfc6265)
- [CURLOPT_COOKIE](https://curl.se/libcurl/c/CURLOPT_COOKIE.html)
- [curl - HTTP Cookies](https://curl.se/docs/http-cookies.html)

leviliangtw:
@COM8 #778 is ready for this issue.

Issue #158, opened 2016-12-12
SetCookie just modify the header


dlee-libo:
I noticed that the implementation of Session::SetCookie is based on CURLOPT_COOKIE. And from the document of libcurl: This option sets the cookie header explicitly in the outgoing request(s). If multiple requests are done due to authentication, followed redirections or similar, they will all get this cookie passed on. I think this behaviour is not what we expect although python's requests lib act the same. Or can we provide another API to initialize cookie engine?

mexus:
Could you please elaborate a little bit on this topic? some examples would certainly help..
actually what behaviour do you expect exactly?

dlee-libo:
The problem with CURLOPT_COOKIE is that it just change request headers. Take below code as an example.

client:
```cpp
cpr::Session s;
s.SetCookies({{"status", "off"}, {"name", "debug"}});
s.SetRedirect(true);
s.SetMaxRedirects(cpr::MaxRedirects{5});
s.SetUrl(cpr::Url{"http://127.0.0.1:9999"});
auto r = s.Get();
for (auto &kv : r.cookies) {
    std::cout << kv.first << std::endl;
}
```

server(nodejs snippets):
```javascript
var express = require('express');
var cookieParser = require('cookie-parser');
var app = express();
app.use(cookieParser());
var port = 9999;

app.get('/', (req, res)=> {
    console.log(req.headers);

    if (req.cookies.status == 'off') {
        res.writeHead(302, {
            'Location': 'http://127.0.0.1:' + port + '/',
            'Set-Cookie': 'status=on'
        });
        res.end();
    } else {
        res.end('hello');
    }

});

app.listen(port);

```

What we expect:
Client send get request to server with 'status=off;name=debug' in cookie.
Server respond with status 302 and set client's cookie to 'status=on'.
Client follow server's redirection, issue another request with 'status=on;name=debug' in cookie.
Server respond 'hello'. Client print 'name\nstatus'

What we really get:
Client send get request to server with 'status=off;name=debug' in cookie.
Server respond with status 302 and set client's cookie to 'status=on'.
Client follow server's redirection, issue another request with '**status=on; name=debug; status=off;** ' in cookie.
Server respond 'hello'. Client print '**status**'

According to RFC6265 section 4.2.2

> Although cookies are serialized linearly in the Cookie header,
   servers SHOULD NOT rely upon the serialization order.  In particular,
   if the Cookie header contains two cookies with the same name (e.g.,
[...]

Issue #26, opened 2015-07-20
MSVC support


whoshuu:

smiley:
What version are you targeting? It's nearly-flawless (save for some warnings on CRT functions) on MSVC14.

nshcat:
Same here, works fine with the current version of MSVC. Only thing that has to be done is to disable deprecation warnings using  __/sdl-__ compiler switch.

smiley:
The tests, however, aren't compiling due to a problem with `Response` objects and `std::tuple` in use in `std::bind`. (See #38)

whoshuu:
Hey @smiley, thanks for the information you posted about the builds. I'm going to go ahead and write out separate issues for each error and link them here. Once we get those sorted out, this can get closed.

@nshcat, that's awesome that you got it to work on the latest MSVC! Have you actually linked it to an executable and made requests using it? The library building is different from the library working, just want to know how far you were able to take things.

smiley:
Whoops, looks like my commit message in 1a04ac7 triggered a future close on this issue, once my PR (#38) is accepted. Any way to turn that off on your end?

whoshuu:
@smiley, no worries, I'll fix it up before merging it in.

whoshuu:
#40 and #41 are all sorted out so this is good to go.

PR #778, opened 2022-07-05
Refactor cpr::Cookies for storing more fields (#777)

leviliangtw:
## Issue

- #158 
- #777 

## Changes

- [x] Improved Cookies
- [X] Refactored Unit Tests

### Improved Cookies

Now the cookies received from a server could be easily accessed like:

```cpp
cpr::Response response = cpr::Get(url);
cpr::Cookies cookies{response.cookies};
for(auto cookie : cookies) {
    std::cout << cookie->GetName() << ":" << cookie->GetValue() << std::endl;
}
```

In addition to the cookie name and the value, you could also get the domain, path, expires, and so on:

```cpp
class Cookie {
  public:
    // skip constructors
    const std::string GetDomain() const;
    bool IsIncludingSubdomains() const;
    const std::string GetPath() const;
    bool IsHttpsOnly() const;
    const std::chrono::time_point<std::chrono::system_clock> GetExpires() const;
    const std::string GetExpiresString();
    const std::string GetName() const;
    const std::string GetValue() const;
}
```

If you would like to set cookies from the client-side, you can use the original interface just like the old implementations: 

```cpp
cpr::Session session{};
session.SetUrl(url);
session.SetCookies(cpr::Cookies{
        {"SID", "31d4d96e407aad42"},
        {"lang", "en-US"},
});
```

## Refactored Unit Tests

I corrected and simplified all unit tests concerning cookies. 
Here I only remained 2 mock server-side APIs for receiving normal and empty cookies: 

- OnRequestBasicCookies()
- OnRequestEmptyCookies()

Furthermore, I added another 2 APIs for reflecting `cpr::SetCookies` from the client-side and redirecting HTTP request according to cookies status: 

- OnRequestCookiesReflect()
- OnRequestRedirectionWithChangingCookies()

With the 2 APIs above, I rewrote all unit tests about `cpr::Session::SetCookie` and added new unit tests for HTTP redirection under the situation of changing cookies.

### DIFF
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -1,51 +1,107 @@
 #include "cpr/cookies.h"
 
 namespace cpr {
-std::string Cookies::GetEncoded(const CurlHolder& holder) const {
+const std::string Cookie::GetDomain() const {
+    return domain_;
+}
+
+bool Cookie::IsIncludingSubdomains() const {
+    return includeSubdomains_;
+}
+
+const std::string Cookie::GetPath() const {
+    return path_;
+}
+
+bool Cookie::IsHttpsOnly() const {
+    return httpsOnly_;
+}
+
+
+/**
+ * TODO: Update the implementation using `std::chrono::utc_clock` of C++20
+ */
+const std::chrono::time_point<std::chrono::system_clock> Cookie::GetExpires() const {
+    return std::chrono::system_clock::from_time_t(expires_);
+}
+
+const std::string Cookie::GetExpiresString() const {
+    std::stringstream ss;
+    std::tm tm{};
+#ifdef _WIN32
+    gmtime_s(&tm, &expires_);
+#else
+    gmtime_r(&expires_, &tm);
+#endif
+    ss << std::put_time(&tm, "%a, %d %b %Y %H:%M:%S GMT");
+    return ss.str();
+}
+
+const std::string Cookie::GetName() const {
+    return name_;
+}
+
+const std::string Cookie::GetValue() const {
+    return value_;
+}
+
+const std::string Cookies::GetEncoded(const CurlHolder& holder) const {
     std::stringstream stream;
-    for (const std::pair<const std::string, std::string>& item : map_) {
+    for (const cpr::Cookie& item : cookies_) {
         // Depending on if encoding is set to "true", we will URL-encode cookies
-        stream << (encode ? holder.urlEncode(item.first) : item.first) << "=";
+        stream << (encode ? holder.urlEncode(item.GetName()) : item.GetName()) << "=";
 
         // special case version 1 cookies, which can be distinguished by
         // beginning and trailing quotes
-        if (!item.second.empty() && item.second.front() == '"' && item.second.back() == '"') {
-            stream << item.second;
+        if (!item.GetValue().empty() && item.GetValue().front() == '"' && item.GetValue().back() == '"') {
+            stream << item.GetValue();
         } else {
             // Depending on if encoding is set to "true", we will URL-encode cookies
-            stream << (encode ? holder.urlEncode(item.second) : item.second);
+            stream << (encode ? holder.urlEncode(item.GetValue()) : item.GetValue());
         }
         stream << "; ";
     }
     return stream.str();
 }
 
-std::string& Cookies::operator[](const std::string& key) {
-    return map_[key];
+cpr::Cookie& Cookies::operator[](size_t pos) {
+    return cookies_[pos];
 }
 
 Cookies::iterator Cookies::begin() {
-    return map_.begin();
+    return cookies_.begin();
 }
 
 Cookies::iterator Cookies::end() {
-    return map_.end();
+    return cookies_.end();
 }
 
 Cookies::const_iterator Cookies::begin() const {
-    return map_.begin();
+    return cookies_.begin();
 }
 
 Cookies::const_iterator Cookies::end() const {
-    return map_.end();
+    return cookies_.end();
 }
 
 Cookies::const_iterator Cookies::cbegin() const {
-    return map_.cbegin();
+    return cookies_.cbegin();
 }
 
 Cookies::const_iterator Cookies::cend() const {
-    return map_.cend();
+    return cookies_.cend();
+}
+
+void Cookies::emplace_back(const Cookie& str) {
+    cookies_.emplace_back(str);
+}
+
+void Cookies::push_back(const Cookie& str) {
+    cookies_.push_back(str);
+}
+
+void Cookies::pop_back() {
+    cookies_.pop_back();
 }
 
 } // namespace cpr
--- a/cpr/cookies.cpp
+++ b/cpr/cookies.cpp
@@ -1,4 +1,5 @@
 #include "cpr/cookies.h"
+#include <ctime>
 
 namespace cpr {
 const std::string Cookie::GetDomain() const {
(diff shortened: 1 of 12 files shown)


cpr/cookies.cpp
```diff
@@ -1,51 +1,97 @@
...
+
+const std::string Cookie::GetPath() const {
+    return path_;
+}
+
+bool Cookie::IsHttpsOnly() const {
+    return httpsOnly_;
+}
+
+time_t Cookie::GetExpires() const {
```
COM8, line 20: `time_t` is C. Please change this to something like a [`std::chrono::time_point`](https://en.cppreference.com/w/cpp/chrono/time_point). While doing this, please ensure it's compatible with our current cpp standard.

include/cpr/cookies.h
```diff
@@ -3,12 +3,38 @@
...
 
 namespace cpr {
 
+static const int EXPIRES_STRING_SIZE = 100;
+
+class Cookie {
+  public:
+    Cookie() = default;
+    Cookie(const std::string& name, const std::string& value, const std::string& domain = "", bool p_isIncludingSubdomains = false, const std::string& path = "/", bool p_isHttpsOnly = false, time_t expires = 0) : name_(name), value_(value), domain_(domain), includeSubdomains_(p_isIncludingSubdomains), path_(path), httpsOnly_(p_isHttpsOnly), expires_(expires){};
+    // Cookie(const std::string& name, const std::string& value, const std::string& domain = "", bool p_isIncludingSubdomains = false, const std::string& path = "", bool p_isHttpsOnly = false, uint32_t expires = 0) : name_(name), value_(value), domain_(domain), includeSubdomains_(p_isIncludingSubdomains), path_(path), httpsOnly_(p_isHttpsOnly), expires_(expires){};
```
COM8, line 18: Can this piece of code be removed?

include/cpr/cookies.h
```diff
@@ -25,26 +51,29 @@ class Cookies {
     bool encode{true};
 
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
-    Cookies(bool p_encode = true) : encode(p_encode) {}
-    Cookies(const std::initializer_list<std::pair<const std::string, std::string>>& pairs, bool p_encode = true) : encode(p_encode), map_{pairs} {}
+    Cookies(bool p_encode = true) : encode(p_encode){};
```
COM8, line 54:
```suggestion
    Cookies(bool p_encode = true) : encode(p_encode) {};
```

test/util_tests.cpp
```diff
@@ -7,6 +7,29 @@
 
 using namespace cpr;
 
+TEST(UtilParseCookiesTests, BasicParseTest) {
```
COM8, line 10: Hahaha nice test case :D

include/cpr/util.h
```diff
@@ -30,6 +30,8 @@ std::vector<std::string> split(const std::string& to_split, char delimiter);
 std::string urlEncode(const std::string& s);
 std::string urlDecode(const std::string& s);
 void secureStringClear(std::string& s);
+bool isTrue(const std::string& s);
```
COM8, line 33: Based on the implementation it only matches on `TRUE`. But HTTP is "usually" not case sensitive. Please don't compare it case sensitive and add a short comment here, when it function matches.

COM8 (changes requested):
Great work! This looks really nice and I also really appreciate the time you invest here!
Only a few things to sort out.

One general question:
When serializing cookies, what will be included? Don't we also should include all other properties here?

leviliangtw:
> Great work! This looks really nice and I also really appreciate the time you invest here! Only a few things to sort out.
> 
> One general question: When serializing cookies, what will be included? Don't we also should include all other properties here?

Nope, we should not include other properties while the client is going to set the "Cookie" header in a request. 
Here is the reference: [RFC 6265 - HTTP State Management Mechanism - 4.2. Cookie](https://datatracker.ietf.org/doc/html/rfc6265#section-4.2).

include/cpr/cookies.h
leviliangtw: I would like to modify the initialization like that: 
```suggestion
    Cookies(bool p_encode = true) : encode{p_encode} {};
```

include/cpr/cookies.h
```diff
@@ -25,26 +51,29 @@ class Cookies {
...
     const_iterator end() const;
     const_iterator cbegin() const;
     const_iterator cend() const;
+    void push_back(const Cookie& __x);
+    void emplace_back(const Cookie& __x);
+    void pop_back();
 
   private:
-    std::map<std::string, std::string> map_;
+    std::vector<cpr::Cookie> vectors_;
```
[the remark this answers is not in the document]
leviliangtw: Sure

test/httpServer.cpp
```diff
@@ -102,89 +102,88 @@ void HttpServer::OnRequestLowSpeedBytes(mg_connection* conn, http_message* /*msg
...
 
 void HttpServer::OnRequestBasicCookies(mg_connection* conn, http_message* /*msg*/) {
-    time_t t = time(nullptr) + 5; // Valid for 1 hour
-    char expire[100], expire_epoch[100];
-    snprintf(expire_epoch, sizeof(expire_epoch), "%lu", static_cast<unsigned long>(t));
-    strftime(expire, sizeof(expire), "%a, %d %b %Y %H:%M:%S GMT", gmtime(&t));
-    std::string cookie{"cookie=chocolate; expires=\"" + std::string{expire} + "\"; http-only;"};
-    std::string cookie2{"icecream=vanilla; expires=\"" + std::string{expire} + "\"; http-only;"};
+    time_t expires_time = 3905119080; // Expires=Wed, 30 Sep 2093 03:18:00 GMT
+    std::array<char, EXPIRES_STRING_SIZE> expires_string;
```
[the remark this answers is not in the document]
leviliangtw: The `EXPIRES_STRING_SIZE` is also used in the function `Cookie::GetExpiresString` inside the `cookie.cpp`.

include/cpr/cookies.h
```diff
@@ -3,12 +3,38 @@
...
 #include "cpr/curlholder.h"
 #include <initializer_list>
-#include <map>
 #include <sstream>
 #include <string>
+#include <vector>
 
 namespace cpr {
 
+static const int EXPIRES_STRING_SIZE = 100;
```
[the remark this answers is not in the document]
leviliangtw: This is because `std::array` needs `const` value for it's declaration, and I will add the following comments:
```cpp
/**
 * EXPIRES_STRING_SIZE is an explicitly static and const variable that could be only accessed within the same namespace and is immutable.
 * To be used for "std::array", the expression must have a constant value, so EXPIRES_STRING_SIZE must be a const value.
 **/
static const std::size_t EXPIRES_STRING_SIZE = 100;
```

cpr/cookies.cpp
leviliangtw: Done!

include/cpr/cookies.h
```diff
@@ -25,26 +55,29 @@ class Cookies {
...
+    using iterator = std::vector<cpr::Cookie>::iterator;
+    using const_iterator = std::vector<cpr::Cookie>::const_iterator;
 
     iterator begin();
     iterator end();
     const_iterator begin() const;
     const_iterator end() const;
     const_iterator cbegin() const;
     const_iterator cend() const;
+    void emplace_back(const Cookie& __x);
```
COM8, line 75: Please use a more descriptive name here like `str`.

COM8 (changes requested):
Only one small thing and we are ready to go!
Please also create a docs PR updating the cookie usage: https://docs.libcpr.org/advanced-usage.html#response-objects

COM8:
Also please rebase on master since the CI is running again 😎 .

include/cpr/cookies.h
leviliangtw: Fixed it!

leviliangtw:
> Also please rebase on master since the CI is running again sunglasses .

leviliangtw:
> Also please rebase on master since the CI is running again sunglasses .

Got it!

leviliangtw:
> Only one small thing and we are ready to go! Please also create a docs PR updating the cookie usage: https://docs.libcpr.org/advanced-usage.html#response-objects

The docs PR [#26](https://github.com/libcpr/docs/pull/26) is ready.

PR #783, opened 2022-07-11
MacOS and Windows OpenSSL CI fixes

COM8:
This PR aims to fix the failing CI runs for MacOS and Windows.


COM8:
The issues related to MacOS failing will be fixed in: https://github.com/libcpr/cpr/pull/784

Issue #293, opened 2018-10-15
Disable the 100-continue feature


WLBF:
Recently I meet the [`Session Object`](https://whoshuu.github.io/cpr/advanced-usage.html#session-objects) 100-continue problem. Once a request failed, the next time the cpr client will send `Expect: 100-continue` in next request with same url. Any idea about how to disable this feature. 

Related: 
[https://curl.haxx.se/mail/lib-2017-07/0013.html](https://curl.haxx.se/mail/lib-2017-07/0013.html)

fuzzbuster:
any solution?

xloem:
Hey, I followed your link and looked up how to disable it, and it sounds like the way is simply to specify it as empty:https://gms.tf/when-curl-sends-100-continue.html#disabling-expect-logic .  Have you tried this?

COM8:
@xloem Would you like to create a PR for this?

xloem:
I'm so sorry, I don't recall at all what I was working with that I visited this issue.  I don't understand the issue at this time.

leviliangtw:
Here I test this phenomenon with the following codes:

```cpp
Url url{"http://www.example.com"};
std::string filename{"test_file"};
// prepare a file larger than 1 MB
std::string content{std::string(1024 * 1024, 'a')};
std::ofstream test_file;
test_file.open(filename);
test_file << content;
test_file.close();
Session session{};
curl_easy_setopt(session.GetCurlHolder()->handle, CURLOPT_VERBOSE, 1L);
session.SetUrl(url);
session.SetMultipart({{"file", File{"test_file"}}});
Response response = session.Post();
```

Then I got the following output: 

```shell
*   Trying 93.184.216.34:80...
* Connected to www.example.com (93.184.216.34) port 80 (#0)
> POST / HTTP/1.1
Host: www.example.com
User-Agent: curl/7.80.0
Accept: */*
Accept-Encoding: deflate, gzip
Content-Length: 1048777
Content-Type: multipart/form-data; boundary=------------------------be95cef9532534a7
Expect: 100-continue

* We are completely uploaded and fine
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Encoding: gzip
< Accept-Ranges: bytes
< Cache-Control: max-age=604800
< Content-Type: text/html; charset=UTF-8
< Date: Tue, 26 Jul 2022 14:13:13 GMT
< Etag: "3147526947+gzip"
[...]

PR #798, opened 2022-07-26
Disable the header "Expect: 100-continue" by default (#293)

leviliangtw:
## Issue

- #293 

## Description

When you send a POST using `curl`, there will be a header "Expect: 100-continue" added if the file size is larger than 1 MB. 
The mechanism is aiming to reduce the cost if the server wants to deny the transmission of the data before it has been transmitted. 

Here are some nice references if you want to check them out: 

- [Curl: Time to disable "Expect: 100-continue" by default?](https://curl.se/mail/lib-2017-07/0013.html)
- [Expect 100-continue - Everything curl](https://everything.curl.dev/http/post/expect100)
- [When curl sends 100-continue | Georg's Log](https://gms.tf/when-curl-sends-100-continue.html)

## Changes

- [X] Set the default header to "Expect:" in the method Session::SetHeaderInternal()
- [X] A new mock HTTP server method and unit test

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -61,6 +61,13 @@ void Session::SetHeaderInternal() {
         }
     }
 
+    // libcurl would prepare the header "Expect: 100-continue" by default when uploading files larger than 1 MB.
+    // Here we would like to disable this feature:
+    curl_slist* temp = curl_slist_append(chunk, "Expect:");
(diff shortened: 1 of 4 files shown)


COM8 (approved):
Looks good. Thanks for your help with this one!

PR #799, opened 2022-08-01
Support to build only cpr (alone) as a static library

matthiasklein:
Instead of the global BUILD_SHARED_LIBS option, a private CPR_SHARED_LIB option is used.

Signed-off-by: Matthias Klein <matthias@extraklein.de>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,6 @@ macro(cpr_option OPTION_NAME OPTION_TEXT OPTION_DEFAULT)
     message(STATUS "  ${OPTION_NAME}: ${${OPTION_NAME}}")
 endmacro()
 
-option(BUILD_SHARED_LIBS "Build libraries as shared libraries" ON)
 message(STATUS "C++ Requests CMake Options")
 message(STATUS "=======================================================")
 cpr_option(CPR_GENERATE_COVERAGE "Set to ON to generate coverage reports." OFF)


CMakeLists.txt
```diff
@@ -48,9 +48,9 @@ macro(cpr_option OPTION_NAME OPTION_TEXT OPTION_DEFAULT)
     message(STATUS "  ${OPTION_NAME}: ${${OPTION_NAME}}")
 endmacro()
 
-option(BUILD_SHARED_LIBS "Build libraries as shared libraries" ON)
```
COM8, line 51: This line never should have existed in the first place. Thanks for noticing this.

CMakeLists.txt
```diff
@@ -48,9 +48,9 @@ macro(cpr_option OPTION_NAME OPTION_TEXT OPTION_DEFAULT)
...
 message(STATUS "C++ Requests CMake Options")
 message(STATUS "=======================================================")
+cpr_option(CPR_SHARED_LIB "Build cpr as shared library." ON)
```
COM8, line 53: I see your point here, but wouldn't something like:
```
include(FetchContent)
set(BUILD_SHARED_LIBS_BACKUP BUILD_SHARED_LIBS) # Backup old variable
set(BUILD_SHARED_LIBS OFF)
                         GIT_TAG 871ed52d350214a034f6ef8a3b8f51c5ce1bd400)
FetchContent_MakeAvailable(cpr)
set(BUILD_SHARED_LIBS BUILD_SHARED_LIBS_BACKUP) # Restore old variable
```

be a better solution, which is more "general"?

COM8:
Thanks for taking the time, having a look at this. We really appreciate it!

CMakeLists.txt
KingKili: I, agree. Especially if we consider that we are dependent on libcurl (which is also controlled via 'BUILD_SHARED_LIBS').

matthiasklein: I hadn't thought of that at all.
I have removed the area.

CMakeLists.txt
COM8: After a short discussion with @KingKili it turns out, I was wrong with this line. It is an `option`, which will set `BUILD_SHARED_LIBS` only and if only in case `BUILD_SHARED_LIBS` is not defined already. In this case, if nothing is set at all for `BUILD_SHARED_LIBS`, we will set it to `ON`. So, it is fine, how it is right now. It is no `cpr_option` to reduce complexity.

matthiasklein: Sorry for wasting your time ...

matthiasklein:
Sorry for wasting your time ...

Issue #715, opened 2022-03-20
Instruction how build for ios and android


sergeymild:
Hello, maybe i am stupid), but i don't understand how build this library for ios and android, it would be nice to have some kind of instructions. Thanks.

COM8:
I've never done this for those platforms either. Sooo I need a bit more context to help you with that e.g.:
* What is your build system?
* What OS?
* Any errors?
* What did you try?

sergeymild:
My build system: Cmake
What OS: Mac
Any errors: i tried build for ios and receive this error: `ld: library not found for -lsocket`
What did you try:
git clone https://github.com/libcpr/cpr
after download ios.toolchain.cmake `[os.toolchain.cmake](https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake)`
and put to a folder where i dowloaded cpr
after i did everything by doc

cd cpr
mkdir build
cd build
cmake .. -G Xcode -DCMAKE_TOOLCHAIN_FILE=../ios.toolchain.cmake -DPLATFORM=SIMULATOR64
cmake --build . --config Release

and receive this error:


```
Ld /Users/sergeigolishnikov/develpment/cpr/build/lib/Release/libcurl.dylib normal (in target 'libcurl' from project 'cpr')
    cd /Users/sergeigolishnikov/develpment/cpr
[...]

PR #801, opened 2022-08-11
iOS support

ospfranco:
One small correction on the `body.h` header file where clang fails with an implicit cast and added another flag to prevent searching for CA files which on sandboxed iOS apps is not possible and also not necessary since using Secure Transport already uses the OS layer for TLS verification.

Partially takes care of #715 I'm still playing around but I will provide complete instructions to build it later.


include/cpr/body.h
```diff
@@ -31,7 +31,7 @@ class Body : public StringHolder<Body> {
         std::ifstream is(file.filepath, std::ifstream::binary);
         if (is) {
             is.seekg(0, is.end);
-            int length = is.tellg();
+            int length = (int)is.tellg();
```
KingKili, line 34: Please don't use C-style casts. Use a `static_cast` instead

KingKili (changes requested):
Thanks for your work. What is the reason for the [.DS_Store](.DS_Store) file? If it is not required please also remove it.

ospfranco:
Don't you want to add .DS_store to gitignore? I guess people using macs also contribute every once in a while :)

KingKili:
> Don't you want to add .DS_store to gitignore? I guess people using macs also contribute every once in a while :)

Great Idea!

KingKili (approved):

KingKili (approved):

KingKili:
Thanks for your good work. I will merge it once the CI is run.

COM8 (approved):

PR #803, opened 2022-08-11
Fix iOS check for skipping CA bundle search

ospfranco:
This check is not working if the immediate flag above is not set, just check for it first before checking for the other value.


ospfranco:
nvm, it is necessary

KingKili (approved):
Looks good to me. Thanks!

PR #805, opened 2022-08-12
Enable ENABLE_DEBUG option when CMAKE_BUILD_TYPE is Debug

matthiasklein:
This enables more verbose logging like:
STATE: DID => PERFORMING handle 0x56082ff9dd38; line 2218 (connection #0)

Signed-off-by: Matthias Klein <matthias@extraklein.de>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,7 +222,7 @@ else()
         include(cmake/zlib_external.cmake)
     endif()
 
-    string( TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER )
+    string(TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER)
     if (BUILD_TYPE_LOWER STREQUAL "debug")
         message(STATUS "Enabled curl debug features")
         set(ENABLE_DEBUG ON CACHE INTERNAL "" FORCE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,12 @@ else()
         include(cmake/zlib_external.cmake)
     endif()
 
+    string( TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER )
+    if (BUILD_TYPE_LOWER STREQUAL "debug")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -222,6 +222,12 @@ else()
         include(cmake/zlib_external.cmake)
     endif()
 
+    string( TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER )
```
COM8, line 225:
```suggestion
    string(TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_LOWER)
```

COM8 (changes requested):
Thanks for looking into this. Is this an additional level above [`CURLOPT_VERBOSE`](https://curl.se/libcurl/c/CURLOPT_VERBOSE.html) when it comes to debug output in libcurl?
Or is this flag used during build only?

@KingKili Should we promote this to a global cpr option since there are additional debug targets, like building with sanitizers and so on?

matthiasklein:
> Or is this flag used during build only?
> 

Yes, the flag is used during Build only

COM8:
Then in my eyes we should propagate this to a cpr option. What do you think about this @KingKili ?

KingKili:
> Then in my eyes we should propagate this to a cpr option. What do you think about this @KingKili ?

I agree. Something like `curl_verbose_logging` makes more sense for me.

COM8:
OK, then @matthiasklein could you please add a cpr option called `` wrapping ``
Example:
```cmake
```

COM8:
OK, @matthiasklein could you please add a cpr option called something like `CURL_VERBOSE_LOGGING`, which wraps the curl `ENABLE_DEBUG` option?
```cmake
cpr_option(CURL_VERBOSE_LOGGING "Curl verbose logging during building curl ..." OFF)
```

matthiasklein:
> OK, @matthiasklein could you please add a cpr option called something like `CURL_VERBOSE_LOGGING`, which wraps the curl `ENABLE_DEBUG` option?

Yes, I do

matthiasklein:
Continued in the following PR: https://github.com/libcpr/cpr/pull/810

PR #806, opened 2022-08-12
📝 add documentation to CMake find-package support

guylevy84:
This PR adds documentation in README.md to CMake find-package support (#645)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,8 @@ If you prefer not to use `fetch_content`, you can download, build, and install t
 $ git clone https://github.com/libcpr/cpr.git
 $ cd cpr && mkdir build && cd build
 $ cmake .. -DCPR_USE_SYSTEM_CURL=ON
-$ make -j
-$ sudo make install
+$ cmake --build .
+$ sudo cmake --build . --target install
 ```
 In your `CMakeLists.txt`:
 ```cmake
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,6 +96,23 @@ That should do it!
...
+#### find_package():
+If you prefer not to use `fetch_content`, you can download, build, and install the library and then use CMake `find_package()` function to integrate it into a project.
+
+**Note:** this feature is feasible only if CPR_USE_SYSTEM_CURL is set. (see [#645](https://github.com/libcpr/cpr/pull/645))
+```Bash
+$ git clone https://github.com/libcpr/cpr.git
+$ cd cpr && mkdir build && cd build
+$ cmake .. -DCPR_USE_SYSTEM_CURL=ON
+$ make -j
+$ sudo make install
```
COM8, line 108: I prefer the CMake way of doing this, which is platform independent:
```suggestion
$ cmake --build .
$ sudo cmake --build . --target install
```

COM8 (changes requested):
Thanks for your PR! Only one minor thing and we are ready to do.

guylevy84:
@COM8, since CMake version >= 3.15 we can do `sudo cmake --install .` instead of `sudo cmake --build . --target install`

COM8 (approved):
Good point. Thanks!

PR #812, opened 2022-08-24
Compatiblity for OSX without filesystem.h

kpeo:
There is a long story of pain with filesystem.h on OSX using CLang, so for those who still wants to compile cpr - this patch could be useful. Surely, boost is not the optimal decision, specially if project doesn't use it. Also there are some failed tests, that requires additional research and fixes.

Tested on OSX 10.13 Clang LLVM 10.0.1: 3 tests failed out of 23:
- cpr_post_tests (UrlEncodedPostTests.UrlPostAsyncSingleTest)
- cpr_head_tests (HeadTests.BasicHeadAsyncTest)
- cpr_options_tests (OptionsTests.AsyncBaseUrlTest, OptionsTests.AsyncSpecificUrlTest)


kpeo:
Yes, I see:

  Checking /home/runner/work/cpr/cpr/cpr/file.cpp ...
  /home/runner/work/cpr/cpr/include/cpr/filesystem.h:11:0: error: failed to evaluate #elif condition, division/modulo by zero [preprocessorErrorDirective]
  #elif __has_include(<boost/filesystem.hpp>)
  ^
  make[2]: *** [cpr/CMakeFiles/cpr.dir/build.make:216: cpr/CMakeFiles/cpr.dir/file.cpp.o] Error 1

Looks strange. Checking. Thank you!

kpeo:
With `__has_include("boost/filesystem.hpp")` cppcheck passed, but compilation failed.
As workaround for the issue with cppcheck, this line could be suppressed:
```
//cppcheck-suppress preprocessorErrorDirective
#elif __has_include(<boost/filesystem.hpp>)
```
and cppcheck should be started with `--inline-suppr` argument.

kpeo:
cppcheck should be fine now, since `--inline-suppr` already added in `cmake/cppcheck.cmake`

KingKili:
Would you mind creating a option for this. E.g. `CPR_USE_BOOST_FILESYSTEM`

kpeo:
yes, working on it. 
do I need to switch to another branch for PR while WIP (draft)?

KingKili:
No, there is no need for a new branch

include/CMakeLists.txt
```diff
@@ -51,8 +51,13 @@ target_sources(cpr PRIVATE
 )
 
 # Filesystem
-if ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.1) OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+include(CheckIncludeFileCXX)
```
KingKili, line 54: Maybe we can replace this with a `find_package` call for [boost](https://cmake.org/cmake/help/latest/module/FindBoost.html).

KingKili (changes requested):
First, thanks for your work. Two minor things.
1. Could you please have a look at the before mentioned find_package for boost
2. Could you add a CI run which uses boost filesystem

include/CMakeLists.txt
kpeo: Sorry for delay, I’ve been very sick these days and just got back to work.

So now CMake uses FindBoost first to find Boost.Filesystem in case we set CPR_USE_BOOST_FILESYSTEM option to ON (OFF by default) and if this option is disabled - we try to add Standard.Library instead.

Added CI run with Boost for "autossl" build configuration. Not sure we will need the full test matrix with other configurations/platforms for this Boost option yet.

KingKili (approved):
No worries. I hope you are better now.
If the CI succeeds, it is ready for merging. Thanks for your work.

include/CMakeLists.txt
```diff
@@ -51,7 +51,15 @@ target_sources(cpr PRIVATE
 )
 
 # Filesystem
-if ((CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.1) OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if(CPR_USE_BOOST_FILESYSTEM)
+  find_package(Boost 1.44 REQUIRED COMPONENTS filesystem)
+  if(Boost_FOUND)
+    include_directories(${Boost_INCLUDE_DIRS})
+    target_link_libraries(cpr PUBLIC ${Boost_LIBRARIES})
```
KingKili, line 58: One minor thing I just found. Please use targets. And replace the `include_directories` with `target_include_directories`

KingKili (changes requested):

include/CMakeLists.txt
kpeo: Oh, yes. Changed it to `target_include_directories`. Thanks!

KingKili (changes requested):
Can you please use targets like in [here](https://github.com/libcpr/cpr/blob/c642a767e7788d8ccd3de8ddfae00cd92d12d715/cpr/CMakeLists.txt#L34l) instead of the old variables `${...}`

KingKili (approved):

KingKili:
All looks good. Thanks for your work (and sorry for the many iterations). I will merge after the CI passes.

kpeo:
Great! Thank you! There are possible bugs on using fs::file still (e.g. suspecting multiperform.h for Download) - not found it yet, tracing/debugging in spare time.

PR #814, opened 2022-08-26
Add std::map based constructors to Proxies and ProxyAuthentication

kp-cat:
This PR extends Proxies and ProxyAuthentication classes with an extra constructor to be able to init with `std::map`, not just `std::initializer_list`.

In this way, the proxy can be initialized dynamically. It's not needed to write the proxy list in the code in advance.

### DIFF
--- a/cpr/proxies.cpp
+++ b/cpr/proxies.cpp
@@ -8,6 +8,7 @@
 namespace cpr {
 
 Proxies::Proxies(const std::initializer_list<std::pair<const std::string, std::string>>& hosts) : hosts_{hosts} {}
+Proxies::Proxies(const std::map<std::string, std::string>& hosts) : hosts_{hosts} {}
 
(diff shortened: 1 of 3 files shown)


COM8 (approved):
Looks good, thanks for contributing!

kp-cat:
Hi @COM8,

I see that this change is labeled CPR 1.10.0. Is it possible to release it in an earlier version? I would need this change and I don't like to create a local cpr patch for this in my repository.

How can I help with this release? Should I cherrypick this change and create a PR on [1.9.x](https://github.com/libcpr/cpr/tree/1.9.x) branch?

COM8:
Yes, sure I can quickly backport this to 1.9.x and prepare a release.

COM8:
New release is available: https://github.com/libcpr/cpr/releases/tag/1.9.2

kp-cat:
Thanks @COM8 🙏

PR #816, opened 2022-08-27
Remove redundant const qualification in auth.h

ciberst:


COM8 (approved):
Thanks for contributing and fixing this!

PR #820, opened 2022-09-07
Support manual hostname resolution

hexadec:
session.cpp:276 might be unsafe?
session.h:49-50 does not look that good


cpr/session.cpp
```diff
@@ -272,6 +272,14 @@ void Session::SetUrl(const Url& url) {
     url_ = url;
 }
 
+void Session::SetResolve(const Resolve& resolve) {
```
COM8, line 275: In my eyes `Session::SetResolve(...)` should override all previous existing resolves. In this case it should clear the current list first and then it should append the new values.

Also it might be useful to pass multiple `cpr::Resolve` objects to this function. Therefore we should also provide something like this: `Session::SetResolve(const std::vector<Resolve>& resolves)`. The same goes then obviously for the `SetOption(...)` call.

include/cpr/session.h
```diff
@@ -45,11 +46,15 @@ class Interceptor;
 class MultiPerform;
 
 class Session : public std::enable_shared_from_this<Session> {
+  private:
+    curl_slist * curlSlist_ = nullptr;
```
COM8, line 50: At the end of this class, there is already a list of private variables. Please move it down there.

COM8: And please use the following notation:
`curl_slist* resolveCurlList_{nullptr}`;

include/cpr/session.h
```diff
@@ -45,11 +46,15 @@ class Interceptor;
...
   public:
     Session();
     Session(const Session& other) = delete;
 
-    ~Session() = default;
+    ~Session() {
+      curl_slist_free_all(curlSlist_);
```
COM8, line 56: This is probably something we won't get around.
What one could do is create a wrapper type around `curl_slist` (e.g. `SList`), following the RAII pattern. In my eyes this would make more sense here.

COM8 (changes requested):
Hey,
thank you very much for taking the time to look into this!
I left a couple of comments and suggestions.
Please also create a PR against the [docs repo](https://github.com/libcpr/docs) with a bit of documentation for this feature.

COM8:
Also please add all your new includes to `cpr.h`.

hexadec:
I hopefully managed to fix the problems you have raised. In the meantime I made some additional changes, which includes switching to std::set when storing port numbers. I realised that the CurlHolder class is exactly what I need to store the resolveCurlList_ so I moved it there. I have also opened a PR in the docs repo.

cpr/session.cpp
```diff
@@ -272,6 +272,23 @@ void Session::SetUrl(const Url& url) {
...
+    SetResolves({resolve});
+}
+
+void Session::SetResolves(const std::vector<Resolve>& resolves) {
+    curl_slist_free_all(curl_->resolveCurlList_);
+    curl_->resolveCurlList_ = nullptr;
+    for (const Resolve& resolve : resolves) {
+        auto resolve_ports_iterator = resolve.ports.cbegin();
```
COM8, line 283: Please avoid using `auto` when ever possible. In my eyes it makes code harder to understand.

COM8: Any reason why this can't be a foreach loop?

include/cpr/curlholder.h
```diff
@@ -27,6 +27,7 @@ struct CurlHolder {
   public:
     CURL* handle{nullptr};
     struct curl_slist* chunk{nullptr};
+    struct curl_slist* resolveCurlList_{nullptr};
```
COM8, line 30:
```suggestion
    struct curl_slist* resolveCurlList{nullptr};
```

cpr/curlholder.cpp
```diff
@@ -20,6 +20,7 @@ CurlHolder::CurlHolder() {
 
 CurlHolder::~CurlHolder() {
     curl_slist_free_all(chunk);
+    curl_slist_free_all(resolveCurlList_);
```
COM8, line 23:
```suggestion
    curl_slist_free_all(resolveCurlList);
```

COM8 (changes requested):
Thank you, we are getting there. Also I noticed some of your changes are breaking our proxy tests.

COM8:
Also could you please add some test cases, where you resolve a couple of http and https domains to `127.0.0.1` or something like this.

hexadec:
I think proxy servers you use might be down. I cannot ping them, and all of my attempts to connect to the port and try and reproduce a proxy request failed. I can see in wireshark that even the TCP SYN package isn't going through

COM8 (approved):
Looks good. Thanks!

COM8:
> I think the proxy servers you use might be down. I cannot ping them, and all of my attempts to connect to the port specified in the code and try to reproduce a proxy request failed. I can see in wireshark that even the TCP SYN package isn't going through

I pushed a fix to `master`. Please rebase your changes once more and then we are ready to go, so I can initiate a final CI run.

COM8:
I replaced the "Not Responding" timeout test case with one that checks for multiple redirects.
Since I'm not allowed to push to your fork, You can find the patch here: https://github.com/COM8/cpr/commit/df415b6e9b598b3bb0acfd226c82ee14e411e379

Please apply it. It should deal with the rather flaky Windows test fails.

hexadec:
Thanks for your fix! I have merged in my fork

COM8 (approved):
Now we are ready to go. Thank you! I will merge this as soon as the CI run is done.

PR #842, opened 2022-10-19
Fixed compiler warnings

proxict:
gcc 12.2.0 in c++20 yielded 2 warnings (for `-Wshadow`, `-Wold-style-cast`)
 - use of old-style cast
 - shadowing parameters
 
This MR attempts to eliminate those warnings.

PS: Thanks a lot for this project, I've used it in a good amount of my personal projects and it has never let me down.  
This is my first MR to this repository, I hope I didn't miss anything important.


COM8:
Thanks for looking into this and fixing it!

PR #843, opened 2022-10-24
deb: split out dev package

dbungert:
The header files and similar are typically located in a libfoo-dev package, so move them there.
Plus a few other tweaks.

I suggest not running debmake in this fashion, it's more common to include the needed files and discard the rest.  A fine starting point though.

Not terribly well tested but should be closer to upload ready I think.


package-build/build-package.sh
```diff
@@ -17,15 +17,10 @@ tar -xzf $ARCHIVE_NAME
 
 cd $LIB_DIR
 
-echo -e "\n\nCalling debmake\n"
```
COM8, line 20: Any reason for removing this `echo` call?

package-build/debian-libcpr/copyright
```diff
@@ -14,6 +14,7 @@ Files:     .clang-format
            cpr-config.cmake
            cpr/*
            include/*
+           nuget/*
```
COM8, line 17: Why should we include the nuget stuff inside the deb package?

COM8 (changes requested):
Thanks for looking into this!
Perhaps this is the reason we failed to find someone to test our submission here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1017437

As far as I'm aware there are libXXX and libXXX-dev packages. The first only include binaries, while the last one contains headers and other stuff required for building in combination with the libXXX package.

package-build/debian-libcpr/copyright
dbungert: A Debian source package has at the heart the original upstream tarball.  If these files would be present in released tarballs, it's best that the copyright file reflect that.

As to the binary packages produced by the build process, the nuget directory is entirely irrelevant :)

If this entry isn't added, the [lintian](https://wiki.debian.org/Lintian) lint tool will complain.

In summary this is just about tidiness and has no outcome on the binary packages.

package-build/build-package.sh
dbungert: To be clear I'm proposing removing the debmake call entirely, and instead including the files that are needed that debmake might provide stubs for (and excluding the unneeded files).    So removing the echo is just a part of that.

dbungert:
Ah, I think I see the problem.

The Debian [ITP](https://wiki.debian.org/ITP) process is as the acronym says - an Intent to package.   Just a way to indicate to others that one is working on it.  Filing the ITP isn't quite enough to indicate that you want or need help uploading it.

To get it into the Debian archive, you need upload permissions or to find someone who does, and this is referred to as [Sponsorship](https://wiki.debian.org/DebianMentorsFaq#Sponsored_Packages)

I will point you to https://wiki.debian.org/DebianMentorsFaq#How_do_I_get_a_sponsor_for_my_package.3F for guidance on acquiring a sponsor - I lack Debian upload permissions so I can't help with that part.

I believe this PR will get you closer to having the package sponsor-ready.

COM8 (approved):
Aha, thanks for letting me know!
I run the workflow locally and pushed a few additional fixes. As long as there is nothing else, I will merge the PR tomorrow.

dbungert:
The additional fixes LGTM.

COM8 (approved):

PR #846, opened 2022-11-01
Add instructions for building with Bazel

cpsauer:
Hey, @COM8!

I finally got those Bazel rules done and released. They make it easy to build libcpr, curl, and dependencies across macOS, Android, iOS, and the other Apple platforms, with Windows and Linux likely coming soon. 
(It's actually just libcurl that's the tricky one--the libcpr rules themselves should already work across platforms, build static and dynamic libs across a variety of compilers, etc.)

As in our discussion over in https://github.com/libcpr/cpr/issues/715#issuecomment-1231161216, the advantage of making CPR easy to use from Bazel is that there's a sizable C++ community there, including most of the highly resourced companies. I'm hoping we might be able to bring a lot more folks the magic of CPR, since there's otherwise no good HTTPS request library easily available from Bazel.

Anyway, I'm hoping you'd be down to take this PR, pointing Bazel users over to the BUILD files and instructions, just released. (From our earlier discussion, it sounded like you'd prefer to keep them external to avoid maintaining without in house Bazel expertise. Totally get it.) I'll manage keep this and the curl files up to date; we're actively using them. And I'll see how many Bazelers I can send your way!

Thanks for your consideration!
Chris

P.S. Btw, someday, Bazel might be useful to you guys for pumping out prebuilt libraries across platforms or running tests. Bazel's good at bundling and fast, incremental builds and tests. But that's for some other time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,11 @@ find_package(cpr REQUIRED)
 add_executable(your_target_name your_target_name.cpp)
 target_link_libraries(your_target_name PRIVATE cpr::cpr)
 ```
+
+### Bazel
+
+Please refer to [hedronvision/bazel-make-cc-https-easy](https://github.com/hedronvision/bazel-make-cc-https-easy).
+
 ### Packages for Linux Distributions
 
 Alternatively, you may install a package specific to your Linux distribution. Since so few distributions currently have a package for cpr, most users will not be able to run your program with this approach.


COM8:
Thank you very much @cpsauer for looking into this.
Yes, in the future this might be a really nice way to automate testing.

Do you plan to continue maintain the Bazel build system?
In case yes, we could also transfer the ownership or fork it, so it lives inside the libcpr group. I could then give you permissions to maintain the code.

cpsauer:
Yes, definitely. I'll keep it up to date. We're actively relying on it in our main codebase, so you can be sure I will :) 

I think almost all of the continued maintenance will be tracking libcurl and the chances to their defines by platform. I've got some automation set up to help. Tracking boringssl and libcpr is fully automated.

Not least because of that automation (which I'm not so keen to redo, and which would require some org setup)--okay if we keep the repo in place, at least for now? Happy to give you write permissions. Just say the word.

COM8:
OK, then lets keep it as it is for now. Yes, you are welcome to give me some write permissions, so I can help updating it directly if needed.

COM8 (approved):

PR #855, opened 2022-12-01
Fixed GetDownloadFileLength error

zydxhs:
Hi,

GetDownloadFileLength returns an incorrect length value when the server does not support the HEAD method for obtaining the file length. The reason is that it does not check the HTTP status code.

### DIFF
--- a/test/download_tests.cpp
+++ b/test/download_tests.cpp
@@ -111,6 +111,26 @@ TEST(DownloadTests, RangeTestMultipleRangesOption) {
     EXPECT_EQ(download_size, response.downloaded_bytes);
 }
 
+bool real_write_data(std::string data, intptr_t userdata) {
+    std::string* dst = (std::string*) userdata;
+    *dst += data;
+    return true;
+}
+TEST(DownloadTests, GetDownloadFileLength) {
+    cpr::Url url{server->GetBaseUrl() + "/get_download_file_length.html"};
+    cpr::Session session;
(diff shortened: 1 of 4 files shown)


COM8 (changes requested):
Thanks for looking into this! Could you please add a unit test for this case?

zydxhs:
Let me try it.

zydxhs:
unit test is added.

test/download_tests.cpp
```diff
@@ -111,6 +111,26 @@ TEST(DownloadTests, RangeTestMultipleRangesOption) {
     EXPECT_EQ(download_size, response.downloaded_bytes);
 }
 
+bool real_write_data(std::string data, intptr_t userdata) {
+    std::string* dst = (std::string*) userdata;
```
COM8, line 115:
```suggestion
    std::string* dst = static_cast<std::string*>(userdata);
```

test/download_tests.cpp
```diff
@@ -111,6 +111,26 @@ TEST(DownloadTests, RangeTestMultipleRangesOption) {
...
+    *dst += data;
+    return true;
+}
```
COM8, line 118:
```suggestion
}

```

COM8 (changes requested):
Thanks, only 3 suggestions to commit and we are ready to go!

zydxhs:
It was my mistake and has been corrected.

PR #857, opened 2022-12-07
API type tweaks

tlammi:
- Use `std::string_view` instead of `const std::string_view&` as the type is trivially copyable
- Remove duplicates by replacing `const std::string&`and `std::string&&` with plain `std::string`

### DIFF
--- a/CppCheckSuppressions.txt
+++ b/CppCheckSuppressions.txt
@@ -1,2 +1,3 @@
 noExplicitConstructor
 ConfigurationNotChecked
+passedByValue
--- a/include/cpr/auth.h
+++ b/include/cpr/auth.h
@@ -11,8 +11,7 @@ enum class AuthMode { BASIC, DIGEST, NTLM };
 
 class Authentication {
   public:
-    Authentication(const std::string& username, const std::string& password, const AuthMode& auth_mode) : auth_string_{username + ":" + password}, auth_mode_{auth_mode} {}
-    Authentication(std::string&& username, std::string&& password, AuthMode&& auth_mode) : auth_string_{std::move(username) + ":" + std::move(password)}, auth_mode_{std::move(auth_mode)} {}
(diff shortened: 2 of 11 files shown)


COM8:
Thanks for looking into this and taking the time to create a PR for this!

I agree with using `std::string_view` instead of `const std::string_view&`.
But, removing the `std::string&&` leads to an additional copy in case someone moves a string into a `cpr::StringHolder` class. In my eyes it is clearer to have both options for `std::string`.

Or do I miss something?

tlammi:
Short story: no, there is no extra copy.
There is a check for this in newer [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html)

In theory there is an extra move but I think compilers are pretty good at optimizing that away.

[This question](https://stackoverflow.com/questions/37935393/pass-by-value-vs-pass-by-rvalue-reference) has a pretty good explanation for this.

These two snippets are pretty much identical in behavior:
```c++
int foo(std::string);
```
and
```c++
int foo(const std::string&);
int foo(std::string&&);
```

Calling the function in the former snippet with an rvalue ref will invoke `std::string(std::string&&)` constructor, but it can also
be called with an lvalue ref which causes copy ctor to be invoked.

tlammi:
So what are your thoughts?

COM8:
Ah, sorry. Forgot about it. I will look into it tomorrow.

COM8 (changes requested):
OK, I agree with all you proposed changes.
I also quickly looked at the assembly generated by gcc and clang.
In gcc it does not make a difference and in clang your version requires two `mov` less.

https://godbolt.org/z/jTE77jKfo

Now, only one thing left, please make sure to disable the check in `cppcheck` for it.

COM8 (approved):
Thanks very much for looking into this!

Issue #257, opened 2018-02-27
Feature request: cancelling/aborting pending request in async requests


abmantis:
If a request is not complete but its timeout value has not been reach (or is not set), there is no way to cancel it. While this is ok for synchronous requests (using cpr::Get), async requests should provide some way of canceling/terminating the request.

RuleOfThrees:
Hello! I'd like to take a look at this, if it's ok!

COM8:
Sure, you are more then welcome to take a look at this. Let me know in case you need any help.

leviliangtw:
I have also looked into this topic and studied a little about it. 
In my view, it's time to refactor and combine `async` and `multiperform` together. 

Here are some facts that I conclude: 

- Currently, there is no simple approach to directly interrupt and cancel an `async` thread. 
- It's feasible to use `std::atomic_bool` as an interrupt signal for an `async` thread; however, it seems not easy to cancel `curl_easy_perform`. 
- There are two approaches to canceling a curl request: 
  - Customized callback functions like `CURLOPT_XFERINFOFUNCTION`
  - Utilizing `curl_multi_remove_handle` by wrapping `curl_easy_perform` into `curl_multi_perform`

References: 

- [How to stop a std::async task | Studio Freya](https://studiofreya.com/2015/07/05/how-to-stop-a-std-async-task/)
- [c++ - Is there a way to cancel/detach a future in C++11? - Stack Overflow](https://stackoverflow.com/questions/12086622/is-there-a-way-to-cancel-detach-a-future-in-c11)
- [c++ - Is it Possible to terminate the std::async thread? - Stack Overflow](https://stackoverflow.com/questions/67907568/is-it-possible-to-terminate-the-stdasync-thread)
- [c - Example how to cancel a libCurl request - Stack Overflow](https://stackoverflow.com/questions/8415976/example-how-to-cancel-a-libcurl-request)
- [CURLOPT_XFERINFOFUNCTION](https://curl.se/libcurl/c/CURLOPT_XFERINFOFUNCTION.html)
- [curl_multi_remove_handle](https://curl.se/libcurl/c/curl_multi_remove_handle.html)

RuleOfThrees:
Sorry to take so long to give feedback, unfortunately real life intervened.
I have done some research on the matter, and experimented a bit. The conclusions so far are as follows:
The prospective cancellation function for an async request will have to be able to handle the following situations:

[...]

PR #864, opened 2022-12-16
MultiAsync

RuleOfThrees:
Here we aim to create an extension of the `cpr` API that introduces multiple, asynchronous, cancellable requests, an asynchronous analogue to the existing `MultiGet` etc. part of the API.
Internally, it is implemented using `cpr`'s `ThreadPool` infrastructure. The creation of the appropriate tests is also in progress.
Aims to close issue #257 , and the development follows the oultine discussed in the issue's comments.

### DIFF
--- a/include/cpr/async.h
+++ b/include/cpr/async.h
@@ -2,7 +2,6 @@
 #define CPR_ASYNC_H
 
 #include "async_wrapper.h"
-#include "session.h"
 #include "singleton.h"
 #include "threadpool.h"
 
@@ -18,7 +17,7 @@ class GlobalThreadPool : public ThreadPool {
 };
 
 /**
- * Return a wraper for a future, calling future.get() will wait until the task is done and return RetType.
+ * Return a wrapper for a future, calling future.get() will wait until the task is done and return RetType.
  * async(fn, args...)
  * async(std::bind(&Class::mem_fn, &obj))
  * async(std::mem_fn(&Class::mem_fn, &obj))
--- a/include/cpr/async.h
+++ b/include/cpr/async.h
@@ -1,10 +1,10 @@
 #ifndef CPR_ASYNC_H
 #define CPR_ASYNC_H
 
-#include "singleton.h"
-#include "threadpool.h"
 #include "async_wrapper.h"
 #include "session.h"
+#include "singleton.h"
+#include "threadpool.h"
 
 namespace cpr {
 
--- a/include/cpr/async.h
+++ b/include/cpr/async.h
@@ -3,6 +3,8 @@
 
 #include "singleton.h"
 #include "threadpool.h"
+#include "async_wrapper.h"
+#include "session.h"
 
 namespace cpr {
 
@@ -16,14 +18,14 @@ class GlobalThreadPool : public ThreadPool {
 };
 
 /**
- * Return a future, calling future.get() will wait task done and return RetType.
+ * Return a wraper for a future, calling future.get() will wait until the task is done and return RetType.
  * async(fn, args...)
  * async(std::bind(&Class::mem_fn, &obj))
  * async(std::mem_fn(&Class::mem_fn, &obj))
  **/
 template <class Fn, class... Args>
 auto async(Fn&& fn, Args&&... args) {
-    return GlobalThreadPool::GetInstance()->Submit(std::forward<Fn>(fn), std::forward<Args>(args)...);
+    return AsyncWrapper{GlobalThreadPool::GetInstance()->Submit(std::forward<Fn>(fn), std::forward<Args>(args)...)};
 }
 
 class async {
--- a/include/cpr/async_wrapper.h
+++ b/include/cpr/async_wrapper.h
@@ -41,7 +41,7 @@ class AsyncWrapper {
     ~AsyncWrapper() {
         if constexpr (isCancellable) {
             if(is_cancelled) {
-                is_cancelled->store(false);
+                is_cancelled->store(true);
             }
         }
     }
--- a/include/cpr/async_wrapper.h
+++ b/include/cpr/async_wrapper.h
@@ -10,11 +10,7 @@
 namespace cpr {
 
 
-enum class [[nodiscard]] CancellationResult {
-    failure,
-    success,
-    invalid_operation
-};
+enum class [[nodiscard]] CancellationResult { failure, success, invalid_operation };
 
 /**
  * A class template intended to wrap results of async operations (instances of std::future<T>)
@@ -22,17 +18,16 @@ enum class [[nodiscard]] CancellationResult {
  *
  * The RAII semantics are the same as std::future<T> - moveable, not copyable.
  */
-template<typename T, bool isCancellable = false>
+template <typename T, bool isCancellable = false>
 class AsyncWrapper {
-    private:
+  private:
     std::future<T> future;
     std::shared_ptr<std::atomic_bool> is_cancelled;
 
-    public:
-
+  public:
     // Constructors
-    explicit AsyncWrapper(std::future<T>&& f): future{std::move(f)} {}
-    AsyncWrapper(std::future<T>&& f, std::shared_ptr<std::atomic_bool>&& cancelledState): future{std::move(f)}, is_cancelled{std::move(cancelledState)} {}
+    explicit AsyncWrapper(std::future<T>&& f) : future{std::move(f)} {}
+    AsyncWrapper(std::future<T>&& f, std::shared_ptr<std::atomic_bool>&& cancelledState) : future{std::move(f)}, is_cancelled{std::move(cancelledState)} {}
 
     // Copy Semantics
     AsyncWrapper(const AsyncWrapper&) = delete;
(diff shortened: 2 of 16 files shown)


COM8:
Thanks for working on this PR!

Roughly during the next week I will have some time to look into it a bit closer and give you detailed feedback on the general way of solving this issue.

RuleOfThrees:
> 
>     * All publicly exposed functions start with a upper case letter. Private ones with a lower case one. I know this is not consequently done throughout the lib right now, but I'm planing a PR to fix this.
Thanks for the heads up, I was a bit perprexed regarding the naming conventions used! There are some `AsyncWrapper` member functions that do not follow this, and this will be corrected.
> 
>     * Method calls have to be camel case, not separated with an `_`.
Concerning the functions [here](https://github.com/libcpr/cpr/blob/84e9fe7c6f606ec1293d9495d79cbf20f3085606/include/cpr/async_wrapper.h#L49) , these are the names and signatures used for member functions of `std::future`, as seen [here](https://en.cppreference.com/w/cpp/thread/future). Changing their names would make `AsyncWrapper` not compatible with `std::future` and break compatibility with previous versions of `cpr`. There can be aliases to these methods, like `AsyncWrapper::WaitFor( ... )` in case someone would want to follow a consistent naming pattern, I suppose? Removing these methods would cause, for example, [this](https://github.com/libcpr/cpr/blob/1.9.x/test/callback_tests.cpp#L50) and many other callback tests to fail.
> 
>     * You define `cpr_pf_arg_t` as `double` in case libcurl < `0x072000` is used. Is this intended, since `CURL_OFF_T` is of type `int64_t `?
So, this requires some explanation, which I didn't provide in comments or commit messages, and that's my bad. Basically, this aims to work as a replacement for this preprocessor branching [here](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/util.h#L23), which is repeated [here, for example](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/util.h#L23). As I understand it, the progress function gets `double`s as arguments for versions prior to the one indicated, and `curl_off_t` aka `int64_t` afterwards. Using a type alias `cpr_pf_arg_t` we can avoid having two declarations for the function signature and improve readability, I think. Besides, I suspect that there should also be the same conditional signature [here](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/callback.h#L57), and the error hasn't occurred because the tests don't compile with the old version of `libcurl`. With the type alias, this can be corrected with minimal effort, if it's indeed a problem.
I'm planning to add tests for the functionality of `AsyncWrapper` now, and maybe something more regarding cancellation.

RuleOfThrees:
> From looking at your tests it seem that `MultiAsyncGet` supports tuples and not vectors which is very limiting, unless I'm mistaken. Let's say 50 URLs were created during runtime and inserted into a `std::vector<cpr::Url>`, how would I pass the vector to the function?

The extensive usage of tuples is contingent on three factors:

* The tuples that are given as arguments to `cpr::MultiGet`, or `cpr::MultiGetAsync`, each describe a set of parameters that characterize a specific request. So a `std::tuple<cpr::Url>` is one that only contains a url, a `std::tuple<cpr::Url, cpr::Payload, cpr::ProgressFunction>` one that has a specific payload and method for process monitoring, etc. An `std::vector` has multiple contents _of the same type_. 

You could however, say, "why can't I give a runtime-generated `std::vector<std::tuple<cpr::Url>>` containing my urls as argument to the `MultiOp`s?", which comes to the second point:

* `cpr` heavily depends on compile-time code generation. Therefore, each `MultiOp` is instantiated at compile-time for each specific combination of arguments. Which leads to the consequence, that `MultiGet<std::tuple<cpr::Url> /* repeats 50 times */>` is a different instantiation to the same function for a different number of arguments. These operations intend (the way I understand things, at least) to accomodate multiple operations, with complex parameters, that should function "as a group". Here, it's worth remarking that while `std::tuple` has a `constexpr` constructor, `std::vector` does not in C++17, which `cpr` targets.

So, to `Get` N `cpr::Url`s _generated at runtime_, I'd personally go this way:
```c++
int main() {
  std::vector<cpr::Url> v{getUrlsFromDatabase()};
  std::vector<cpr::AsyncResponse>resps{};
  std::transform(v.cbegin(), v.cend(), std::back_inserter(resps), cpr::GetAsync<cpr::Url>);
}
```
The `MultiOp`s, async or not, require compile-time awareness of the argument types and counts, the way it is now.

COM8:
> Concerning the functions [here](https://github.com/libcpr/cpr/blob/84e9fe7c6f606ec1293d9495d79cbf20f3085606/include/cpr/async_wrapper.h#L49) , these are the names and signatures used for member functions of `std::future`, as seen [here](https://en.cppreference.com/w/cpp/thread/future). Changing their names would make `AsyncWrapper` not compatible with `std::future` and break compatibility with previous versions of `cpr`. There can be aliases to these methods, like `AsyncWrapper::WaitFor( ... )` in case someone would want to follow a consistent naming pattern, I suppose? Removing these methods would cause, for example, [this](https://github.com/libcpr/cpr/blob/1.9.x/test/callback_tests.cpp#L50) and many other callback tests to fail.

In this case we keep the `_` naming convention for obvious reasons for functions related to `std`. Thanks for pointing it out!

> So, this requires some explanation, which I didn't provide in comments or commit messages, and that's my bad. Basically, this aims to work as a replacement for this preprocessor branching [here](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/util.h#L23), which is repeated [here, for example](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/util.h#L23). As I understand it, the progress function gets `double`s as arguments for versions prior to the one indicated, and `curl_off_t` aka `int64_t` afterwards. Using a type alias `cpr_pf_arg_t` we can avoid having two declarations for the function signature and improve readability, I think. Besides, I suspect that there should also be the same conditional signature [here](https://github.com/libcpr/cpr/blob/1.9.x/include/cpr/callback.h#L57), and the error hasn't occurred because the tests don't compile with the old version of `libcurl`. With the type alias, this can be corrected with minimal effort, if it's indeed a problem. I'm planning to add tests for the functionality of `AsyncWrapper` now, and maybe something more regarding cancellation.

Aha, sounds reasonable!

RuleOfThrees:
The bug causing test failure has been reproduced and, I believe, fixed: The reason was that in some tests from the `MultiAsyncCancelTests` set, the worker thread would continue executing the request shortly after the test was finished, and thus try to access mutexes or condition variables within the the `observer_fn` after their destructor was called (dangling references). I am altering the tests to ensure this scenario does not occur, and will commit after ensuring everything works as intended.

include/cpr/async.h
```diff
@@ -1,6 +1,8 @@
 #ifndef CPR_ASYNC_H
 #define CPR_ASYNC_H
 
+#include "async_wrapper.h"
+#include "session.h"
```
COM8, line 5:
```suggestion
#include "async_wrapper.h"
```

include/cpr/async.h
```diff
@@ -16,14 +18,14 @@ class GlobalThreadPool : public ThreadPool {
 };
 
 /**
- * Return a future, calling future.get() will wait task done and return RetType.
+ * Return a wraper for a future, calling future.get() will wait until the task is done and return RetType.
```
COM8, line 21:
```suggestion
 * Return a wrapper for a future, calling future.get() will wait until the task is done and return RetType.
```

include/cpr/async_wrapper.h
```diff
@@ -0,0 +1,144 @@
...
+
+#include <atomic>
+#include <future>
+#include <memory>
+
+#include "cpr/response.h"
+
+namespace cpr {
+
+
```
COM8, line 12:
```suggestion

```

include/cpr/async_wrapper.h
```diff
@@ -0,0 +1,144 @@
...
+    // Destructor
+    ~AsyncWrapper() {
+        if constexpr (isCancellable) {
+            if(is_cancelled) {
+                is_cancelled->store(true);
+            }
+        }
+    }
+
+
```
COM8, line 49:
```suggestion

```

include/cpr/util.h
```diff
@@ -20,11 +19,15 @@ size_t headerUserFunction(char* ptr, size_t size, size_t nmemb, const HeaderCall
...
-#endif
+
+template <typename T = ProgressCallback>
+int progressUserFunction(const T* progress, cpr_pf_arg_t dltotal, cpr_pf_arg_t dlnow, cpr_pf_arg_t ultotal, cpr_pf_arg_t ulnow) {
+    const int cancel_retval{1};
+    static_assert(cancel_retval != CURL_PROGRESSFUNC_CONTINUE);
+    return (*progress)(dltotal, dlnow, ultotal, ulnow) ? 0 : cancel_retval;
+}
+
+
```
COM8, line 30:
```suggestion

```

COM8 (changes requested):
I'm really happy with the outcome. Thanks. Only a few minor formatting things to fix and one unnecessary include.

With that done, we now only need a bit of user documentation over here (https://github.com/libcpr/docs) and then we are ready to go 🎉.

COM8 (approved):

PR #868, opened 2023-01-03
Fixed thread pool not creating enough new threads after awhile

xfangfang:
1. `InitialRun` will never be false, so `idle_thread_num` will only be decremented once when deleting threads. As the tasks progress, `idle_thread_num` will increase, which will result in the inability to create new threads;
2. Because `idle_thread_num` is size_t type, so it is unnecessary to judge whether it is less than 0;

Thank you very much for this project. I hope my fix is correct.

### DIFF
--- a/cpr/threadpool.cpp
+++ b/cpr/threadpool.cpp
@@ -100,9 +100,9 @@ bool ThreadPool::CreateThread() {
             if (task) {
                 task();
                 ++idle_thread_num;
-            } else if (initialRun) {
-                ++idle_thread_num;
-                initialRun = false;
+                if (initialRun) {
+                    initialRun = false;
+                }
             }
         }
     });


include/cpr/threadpool.h
```diff
@@ -67,7 +67,7 @@ class ThreadPool {
         if (status == STOP) {
             Start();
         }
-        if (idle_thread_num <= 0 && cur_thread_num < max_thread_num) {
+        if (idle_thread_num == 0 && cur_thread_num < max_thread_num) {
```
COM8, line 70: I prefer range based checks rather than equality.
This has the simple reason that in the case we get a bit flip or something like it, we are still "safe" and in my eyes it's clearer to read.

xfangfang: `idle_thread_num` is a size_t type variable, when it is 0, executing "--idle_thread_num;", it will becomes a super large value instead of a negative number.
If you want to be more safe, is it better to define `idle_thread_num` as a signed number?

xfangfang:
@COM8 Delete changes in threadpool.h

include/cpr/threadpool.h
COM8: Jup, I'm aware of that. It's just a general thing for me :)

COM8 (approved):

include/cpr/threadpool.h
xfangfang: > Jup, I'm aware of that. It's just a general thing for me :)

Sorry for that ~ I think you are right

PR #874, opened 2023-01-24
Add check for CMP0135

yeSpud:
On newer versions of cmake a warning is given about DOWNLOAD_EXTRACT_TIMESTAMP not being set.
```
CMake Warning (dev) at [...]/cmake-3.24/Modules/FetchContent.cmake:1267 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
```

This just adds a new check for those newer versions and sets [CMP0135](https://cmake.org/cmake/help/latest/policy/CMP0135.html) to the new policy.

Feel free to change it to the old policy if that would be a better fit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,9 @@ else()
     # Disable linting for curl
     clear_variable(DESTINATION CMAKE_CXX_CLANG_TIDY BACKUP CMAKE_CXX_CLANG_TIDY_BKP)
 
+    if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
+		cmake_policy(SET CMP0135 NEW)
+	endif()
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Thanks for your contribution @yeSpud !

Issue #892, opened 2023-03-30
[v1.10.2] g++ compile error: curlholder.h:31:12: error: using typedef-name ‘curl_mime’ after ‘struct’


stayknight:
### Description

When I integrate cpr into my project, the compiler gives this error:
```
/somerootpath/cpr/include/cpr/curlholder.h:31:12: error: using typedef-name ‘curl_mime’ after ‘struct’
   31 |     struct curl_mime* multipart{nullptr};
      |            ^~~~~~~~~
In file included from /somerootpath/cpr/include/cpr/curlholder.h:8,
                 from /somerootpath/cpr/include/cpr/cookies.h:4,
                 from /somerootpath/cpr/include/cpr/util.h:9,
                 from /somerootpath/cpr/cpr/auth.cpp:2:
/usr/include/arm-linux-gnueabihf/curl/curl.h:2099:33: note: ‘curl_mime’ has a previous declaration here
 2099 | typedef struct curl_mime_s      curl_mime;      /* Mime context. */
      |                                 ^~~~~~~~~
```

### Example/How to Reproduce

```cmake
set(ENV{CPR_USE_SYSTEM_CURL} "ON")
FetchContent_Declare(cpr
  URL https://github.com/libcpr/cpr/archive/refs/tags/1.10.2.tar.gz
  DOWNLOAD_NAME cpr-v1.10.2.tar.xz
)
FetchContent_MakeAvailable(cpr)
```

### Possible Fix

I don't know.

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS: ubuntu 20.04 LTS
- Compiler: g++ 9.4.0
- Version: 1.10.2

PR #893, opened 2023-03-30
fix error: using typedef-name ‘curl_mime’ after ‘struct’ (#892)

guylevy84:
we already have previous declaration of ‘curl_mime’ in curl.h: typedef struct curl_mime_s curl_mime;


COM8 (approved):
Awesome! Thank you!

Issue #894, opened 2023-03-31
CPack support


Aposhian:
### Is your feature request related to a problem?

I want to package this as a debian package so that other packages can pull it in easily.

### Possible Solution

Use CPack to generate deb packages. Basically just need to include cpack in the cmake so that post-build someone can call `cpack -G DEB` to actually generate the package.

### Alternatives

You could do native packaging for deb, but that's more work.

COM8:
@Aposhian thanks for opening this issue!
We actually already had a few tries of people creating a .deb package ([here](https://github.com/libcpr/cpr/blob/master/.github/workflows/build-deb.yml)). All of them failed at some point.

Using CPack would definitely be an alternative. Do you want to look into this? Else I will since I currently look into the process of packaging an RPM with CPack for a different project anyway.

Aposhian:
Yeah I can give it a go. It shouldn't be too difficult to setup Cpack.

RogueScholar:
I don't consider myself an elite packager, but I've had cpr building daily over on Launchpad for over six months on Launchpad, populating working deb packages in a [PPA for another project](https://code.launchpad.net/~appimagelauncher-team/+archive/ubuntu/build-deps). You can see the recipe [here](https://code.launchpad.net/~appimagelauncher-team/+recipe/cpr-daily-git), along with the Git repo that houses the [packaging files](https://git.launchpad.net/~appimagelauncher-team/appimageupdate/+git/cpr-packaging/tree/), in case you'd care to upstream them here. It did require [patching your CMake configuration](https://git.launchpad.net/~appimagelauncher-team/appimageupdate/+git/cpr-packaging/tree/debian/patches/0000-remove-Werror-from-CXXFLAGS.patch), though it wasn't anything too drastic.

It goes without saying that anyone is free to use the packages, the PPA or the packaging files for whatever they see fit, they're all out in the open for precisely that reason. Nor is this to say that building in support for the CPack system isn't a worthwhile time investment, I think it is since CPack is generally more forgiving than the Debian packaging toolchain. I'd love to see cpr land in the official Debian/Ubuntu package archives at some point, and perhaps this can be a starting point.

COM8:
@RogueScholar thanks for sharing this!
You are more than welcome to upstream your cpack changes here. I think this would be a great addition.

PR #896, opened 2023-04-03
initial cpack implementation

Aposhian:
Addresses #894.

This adds a very simple cpack implementation, supporting debian and source (non-distro specific) generators so far.

Adding RPM support is probably trivial, but I just need to test what additional vars need to be set.

@COM8 I figure the package contact should probably be you, but I don't know what contact email to put. I've also seen packages where the debian maintainer is listed here, not necessarily the upstream maintainer. So not exactly sure on that.


COM8 (approved):
Thank you very much! Do you also have time to update the CI run for building the .deb package?

COM8:
Closing since this one is stale.

PR #899, opened 2023-04-15
Add cpr package into Fedora Linux official repository

topazus:
Hi, I recently packaged cpr into Fedora official repository, and the cpr package will be available in a few days later.

Ref: https://src.fedoraproject.org/rpms/cpr
https://bodhi.fedoraproject.org/updates/?packages=cpr

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Alternatively, you may install a package specific to your Linux distribution. Si
 Currently, we are aware of packages for the following distributions:
 
 * [Arch Linux (AUR)](https://aur.archlinux.org/packages/cpr)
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Awesome. Thank you @topazus !
Is there a way, we can automate this in our CI?

topazus:
> Awesome. Thank you @topazus ! Is there a way, we can automate this in our CI?

Do you build and publish rpm package on Github release? Sadly, I am not familiar with building a CI/CD workflow with Github actions.

COM8:
The plan is to publish the result as an [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts). If you want, you also can just post a bash script and then I will transform it into a CI run.

PR #923, opened 2023-06-22
Add vector<Part> param to Multipart initialization

sivachandran:
Adds `std::vector<Part>` parameter support to `Multipart` initialization. The change allows us to construct the Multipart dynamically by adding the `Part` selectively and pass the vector to `Multipart`.


COM8:
@sivachandran thanks for contributing!
Could you please also add a move constructor here?

COM8 (approved):

PR #926, opened 2023-07-02
Allow disabling the 'Accept-Encoding' header

### LINKED ISSUE #925: acceptEncoding must provide a way to disable content-encoding

The original curl option CURLOPT_ACCEPT_ENCODING let disabling the content-encoding functionality when you pass NULL as the parameter, instead of using the empty string (which means enable the functionality and accept all known encodings).  The actual implementation in cpr let us choose the encoding but not disabling the functionality.

It is necessary due to some old servers incorrectly returning "Content-Enconding: UTF-8" making interoperability impossible.

### Possible Solution

Add a new encoding method "none", that makes

curl_easy_setopt(curl_->handle, CURLOPT_ACCEPT_ENCODING, null);

in cpr/session.cpp

### Alternatives

_No response_

### Additional Context

_No response_

COM8:
Resolves #925

Adds a new value for `AcceptEncoding`: `disabled`. When a user passes this value to a `AcceptEncoding` object it will prevent curl from creating a `Accept-Encoding`.

It has the same behaviour as passing `NULL` to the curl option [`CURLOPT_ACCEPT_ENCODING`](https://curl.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html).

When passing multiple values to `AcceptEncoding` including `disabled`, an exception of type `std::invalid_argument` get thrown when performing a request.

### Examples

````c++
cpr::Session session;
session.SetUrl("https://example.com");
session.SetAcceptEncoding({AcceptEncodingMethods::disabled}); // Disable setting the `Accept-Encoding` headder
Response response = session.Get();
```` 

````c++
cpr::Session session;
session.SetUrl("https://example.com");
session.SetAcceptEncoding({"disabled"}); // Disable setting the `Accept-Encoding` headder
Response response = session.Get();
```` 

````c++
cpr::Session session;
session.SetUrl("https://example.com");
session.SetAcceptEncoding({AcceptEncodingMethods::disabled, AcceptEncodingMethods::deflate});
Response response = session.Get(); // An exception of type  `std::invalid_argument` will be thrown here since multiple values are passed to `AcceptEncoding` where one of them is `disabled`
````

### DIFF
--- a/cpr/accept_encoding.cpp
+++ b/cpr/accept_encoding.cpp
@@ -1,14 +1,17 @@
 #include "cpr/accept_encoding.h"
 
 #include <algorithm>
+#include <cassert>
+#include <initializer_list>
 #include <iterator>
 #include <numeric>
+#include <stdexcept>
 
 namespace cpr {
 
 AcceptEncoding::AcceptEncoding(const std::initializer_list<AcceptEncodingMethods>& methods) {
     methods_.clear();
-    std::transform(methods.begin(), methods.end(), std::back_inserter(methods_), [&](cpr::AcceptEncodingMethods method) { return cpr::AcceptEncodingMethodsStringMap.at(method); });
+    std::transform(methods.begin(), methods.end(), std::inserter(methods_, methods_.begin()), [&](cpr::AcceptEncodingMethods method) { return cpr::AcceptEncodingMethodsStringMap.at(method); });
 }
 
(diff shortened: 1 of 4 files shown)


obelixzgz:
It's perfect.

obelixzgz:
Thank you very much

COM8:
Perfect. Will finish up this PR next weekend and add docs for it.

COM8:
Docs: https://github.com/libcpr/docs/pull/40

PR #929, opened 2023-07-15
boost/filesystem: use version 4 for compatibility with std::filesystem

cpsauer:
Hello again, @COM8,

I was messing around some more with similar backporting of std::filesystem with boost::filesystem in another project--and I was thinking: Might it make sense to move over to boost filesystem v4?

The advantage of using the newer api (rather than the older one, v3) is that it's more similar to std::filesystem, removing some potential future foot guns if we're just using it as a backport. ([Docs.](https://www.boost.org/doc/libs/1_82_0/libs/filesystem/doc/v4.html))

Thanks for being great!
Chris

### DIFF
--- a/include/cpr/filesystem.h
+++ b/include/cpr/filesystem.h
@@ -3,6 +3,7 @@
 
 // Include filesystem into the namespace "fs" from either "filesystem" or "experimental/filesystem" or "boost/filesystem"
 #ifdef CPR_USE_BOOST_FILESYSTEM
+#define BOOST_FILESYSTEM_VERSION 4 // Use the latest, with the closest behavior to std::filesystem.
 #include <boost/filesystem.hpp>
 namespace fs = boost::filesystem;
 // cppcheck-suppress preprocessorErrorDirective


COM8 (approved):
Agree! Thanks for updating it.

cpsauer:
Hey, Fabian, after some more research, I think I should update my recommendation here: 
**I really think we should backport std::filesystem with https://github.com/gulrak/filesystem instead of boost::filesystem.**

gulrak/filesystem is std::filesystem-compatible (and header-only), sidestepping a lot of the pain we were getting in my other project from the API differences between boost::filesystem and std::filesystem. I'm guessing it'll save you a lot of API-divergence pain, too, as code continues to build on filesystem.

What do you think?

Some notes:
- I started by tossing up a PR with them https://github.com/gulrak/filesystem/pull/167 to add support for non-desktop Apple platforms (which we needed) and to clean things up a bit. 
- Since it's header-only, the change could be as easy as grabbing some of their headers to replace the contents of libcpr's filesystem.h. 
  - I already hacked together something simple on the Bazel side--haven't pushed yet, but libcpr compiles cleanly if you just swap in gulrak/filesystem for boost, conditionals and all.
  - You guys almost certainly have comparative advantage over me with cmake, so I think I should ask if you'd do the actual integration there.
    [One tricky case to think about: If you want to support the shared-library case, you'll want to make sure the same filesystem importation gets picked even if the final binary has a higher min OS version (deployment target) than was used for libcpr compilation. Not sure how cmake usually handles this... 
- You might want to amend their conditionals further to use experimental/filesystem if you feel like it? Happy to add it to my PR, but I don't know the history or tradeoffs as well there as you might.

Cheers!
Chris

P.S. Wish I'd steered us to this straight away when I saw someone had added the broken boost backport...sorry I just learned about it.

PR #934, opened 2023-08-01
Avoid conflicts with external components

kone-tlammi:
Global namespace alias might cause conflicts with external components. I specifically hit this issue when writing tests for a daemon which uses this library.


COM8 (approved):
@kone-tlammi thanks for contributing!

PR #942, opened 2023-08-11
Add ZLIB::ZLIB alias and conditional renaming

sant0s12:
Hi :)

This pull request aims to fix the static zlib being renamed to zlib.dll when it should remain a .lib.
It also ads the ZLIB::ZLIB alias that curl tries to link to.


COM8 (approved):
Thanks! I will merge it as soon as the CI is done.

PR #962, opened 2023-09-14
Fix dangling reference to callback after move

kone-tlammi:
I initially tried a wrapper object for the callbacks but callbacks have dependencies between each others so a unique_ptr
was much simpler.


COM8 (approved):
Perfect. Thanks @kone-tlammi !

PR #966, opened 2023-09-22
Util Method to Parse Raw Cookie Strings

COM8:
As discussed via [gitter](https://app.gitter.im/#/room/#libcpr_community:gitter.im?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge) a option to parse raw cookie strings would be awesome in a context of [Set-Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) Headers.


COM8:
Discarding since it's no general approach. We are only able to parse specific curl provided cookies. So no real use case.

PR #991, opened 2023-12-11
Fix compiler warning C4244 on MSVC.

ssexa:
This fixes the only compiler warning I get when I compile cpr with MSVC (14.37.32822).


KingKili:
Thanks for your contribution! Overall looks good to me. If CI runs through we can merge.

PR #999, opened 2023-12-26
`cpr::Session` Is Now Unmovable/Copyable

### LINKED ISSUE #990: Incorrect implementation of cpr::Session move constructor and move assignment operator
### Description

When moving a cpr::Session object, the curl options CURLOPT_XFERINFODATA, CURLOPT_DEBUGDATA, CURLOPT_WRITEDATA, and others are not updated. 
As a result, when cpr tries to call, for example, the progress callback function, the progresscb_ of the old cpr::Session object that was moved is accessed.

### Example/How to Reproduce

The sample code below is guaranteed to crash in debug mode. (In release mode, it's a matter of luck)

```c++
#include <iostream>
#include <cpr/cpr.h>

cpr::Session CreateCprSession(const std::string& url)
{
    cpr::Session session;

    session.SetUrl(url);
    session.SetProgressCallback(cpr::ProgressCallback(
        [](cpr::cpr_off_t downloadTotal, cpr::cpr_off_t downloadNow, cpr::cpr_off_t uploadTotal, cpr::cpr_off_t uploadNow, intptr_t userdata) {
            std::cout << "Downloaded: " << downloadTotal << " / " << downloadNow << "\n";
            return true;
        }));

    return session;
}


int main()
{
    cpr::Session session = CreateCprSession("https://httpbin.org/get");

    std::cout << "Response: " << response.text << "\n";
    return 0;
}
```

### Possible Fix

I think it can be fixed by adding support to the constructor and assignment operator so that the necessary curl options are updated.

Other (specify in "Additional Context/Your Environment")

- OS: Windows 11 Pro 23H2
- cpr version: 1.10.5
The cpr was obtained from github and added to the project via add_subdirectory.

COM8:
### Why?
`cpr::Session` is a complex object and while making it movable/copyable one has to keep all properties in mind. An easier solution would be to wrap `cpr::Session` inside for example a `std::shared_ptr` to allow it to be movable.

Copyable does not make sense since a curl session is not copyable in a useful way.
What should happen if you copy a `cpr::Session` in mids of a request? Should there be a new request with the same destination and the same connection state? This is not possible in a sensible way.

Fixes #990.

### DIFF
--- a/include/cpr/session.h
+++ b/include/cpr/session.h
@@ -54,11 +54,11 @@ class Session : public std::enable_shared_from_this<Session> {
   public:
     Session();
     Session(const Session& other) = delete;
-    Session(Session&& old) = default;
+    Session(Session&& old) = delete;
 
     ~Session() = default;
 
-    Session& operator=(Session&& old) noexcept = default;
+    Session& operator=(Session&& old) noexcept = delete;
     Session& operator=(const Session& other) = delete;
 
     void SetUrl(const Url& url);
--- a/test/session_tests.cpp
+++ b/test/session_tests.cpp
@@ -1587,11 +1587,11 @@ TEST(CallbackTests, Move) {
     auto session = Session();
     session.SetDebugCallback(DebugCallback([](auto, auto, auto) {}));
 
-    auto use = +[](Session s) {
+    auto use = +[](Session& s) {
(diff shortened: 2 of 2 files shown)


COM8:
Will be keep open  for the next week or so, to ensure there are no other important arguments against this approach.

PR #1003, opened 2023-12-28
Defining NOMINMAX for windows to prevent std::{min,max} issues

### LINKED ISSUE #957: `minwindef.h` `min` and `max` macros conflicting with `std::numeric_limits`
### Description

I'm using `libcpr` together with another library ([`glaze`](https://github.com/stephenberry/glaze)). In files where both are used (libcpr for requests and Glaze for JSON parsing of the response text), compilation is failing with the following error:

```
[...]/include\glaze/util/stoui64.hpp:14:54: error: too few arguments provided to function-like macro invocation
      return a <= std::numeric_limits<uint64_t>::max() - b;
                                                     ^
[...]\Windows Kits\10\Include\10.0.22621.0\shared\minwindef.h:193:9: note: macro 'max' defined here
#define max(a,b)            (((a) > (b)) ? (a) : (b))
        ^
```

If the two libraries are included in the following way, the above error appears:
```cpp
#include <cpr/cpr.h>
#include <glaze/glaze.hpp>
```

By defining the NOMINMAX symbol before including the libcpr headers, the error disappears:
```cpp
#define NOMINMAX
#include <cpr/cpr.h>
#include <glaze/glaze.hpp>
```
Inverting the include order also works:
```cpp
#include <glaze/glaze.hpp>
#include <cpr/cpr.h>
```

### Example/How to Reproduce

On Windows, include the libcpr headers before any code that uses `std::numeric_limits::max()` or `std::numeric_limits::min()`.

### Possible Fix

Define the `NOMINMAX` preprocessor symbol before including Windows headers in `libcpr`. As far as I've seen the `min` and `max` macros from `minwindef.h` are unused by libcpr.

GitHub (branch e.g. master)

- OS: Windows 11
- Branch: 1.10.x

COM8:
Fixes #957.

Still in draft mode since it needs more testing.


maztheman:
IMHO the best way to deal with this is to change std::min() to (std::min)()

nlogozzo:
+1

PR #1020, opened 2024-01-27
WIP: SSL ctx callback

arekmula:
This PR implements feature requested in https://github.com/libcpr/cpr/issues/1019

Issue #1019, opened 2024-01-26
SSL certificate pinning


arekmula:
### Is your feature request related to a problem?

Hey,
For the past 2 years, my team has been using raw CURL implementation to perform networking. We basically had our own wrapper around CURL. Some time ago, we decided to use a well-established CURL wrapper, and we ended up with cpr :) 

We have a requirement to pin SSL certificates. I mean: the client application needs to pin leaf domain certificate first. If domain certificate is valid -> validation process is done. If domain certificate is invalid then as a fallback the client app should check intermediate CA certificate. If CA certificate is valid that means the domain's SSL certificate was changed or expired and new SSL fingerprint update is needed in the client application as soon as possible, but application still can communicate and transfer data. If both certificates are invalid then application shouldn't send or use any data to/from that source because man-in-the-middle attack is possible. 

We were given SHA-256 fingerprints of the CA and domain certificates and we used `CURLOPT_SSL_CTX_FUNCTION` and `CURLOPT_SSL_CTX_DATA` to pass the certificates and perform the validation. 

Is it possible to achieve the pinning with the current state of CPR? I couldn't find anything inside the docs regarding that. I see in the code this part:
```
#if SUPPORT_CURLOPT_SSL_CTX_FUNCTION
#ifdef OPENSSL_BACKEND_USED
    if (!options.ca_buffer.empty()) {
        curl_easy_setopt(curl_->handle, CURLOPT_SSL_CTX_FUNCTION, sslctx_function_load_ca_cert_from_buffer);
        curl_easy_setopt(curl_->handle, CURLOPT_SSL_CTX_DATA, options.ca_buffer.c_str());
    }
#endif
#endif
```

but I don't see an option to provide a custom callback.

### Possible Solution

Add the API to set a custom `CURLOPT_SSL_CTX_FUNCTION` callback with custom `CURLOPT_SSL_CTX_DATA` data.

COM8:
@arekmula thanks for reporting!
You are right, there is no real support for it yet. But I will try to add support for it over the weekend if everything goes as planed.
The Interface would than look something like:
```c++
    /**
     * This callback function gets called by libcurl just before the initialization of an SSL connection
     * after having processed all other SSL related options to give a last chance to an application
     * to modify the behavior of the SSL initialization.
     *
     * If an error is returned from the callback no attempt to establish a connection is made
     * and the perform operation returns the callback's error code.
     * For no error return CURLE_OK from inside 'curl/curl.h'
     *
[...]

PR #1021, opened 2024-01-28
SSL Context Callback

COM8:
Implements #1019.

As an extension to  #1020 this PR adds a function that gets called by libcurl just before the initialization of an SSL connection after having processed all other SSL related options to give a last chance to an application to modify the behavior of the SSL initialization.

Ref: https://curl.se/libcurl/c/CURLOPT_SSL_CTX_FUNCTION.html

How to use:
```c++
    size_t count{0};
    cpr::ssl::SslCtxCallback sslCtxCb{[](const std::shared_ptr<cpr::CurlHolder>& /*curl_holder*/, void* /*ssl_ctx*/, intptr_t userdata) {
                                          // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
                                          size_t* count = reinterpret_cast<size_t*>(userdata);
                                          (*count)++;
                                          return CURLE_OK;
                                      },
                                      // NOLINTNEXTLINE (cppcoreguidelines-pro-type-reinterpret-cast)
                                      reinterpret_cast<intptr_t>(&count)};

    cpr::SslOptions sslOpts = cpr::Ssl(sslCtxCb);

    cpr::Get(url, sslOpts);
    EXPECT_EQ(count, 1);
```

This option is incompatible with `cpr::ssl::CaBuffer`. In case both options are present an instance of `std::logic_error` gets thrown while performing the request.


arekmula (approved):

COM8:
Docs: https://github.com/libcpr/docs/pull/44

COM8:
Currently still WIP since something seams to be broken when using this functionality when cpr is used via fetch content e.g. in our example repo.

Unit tests as well as sanitizers do not find anything yet...

PR #1033, opened 2024-03-25
Optimization cookie time

0x703cd8:
In some environments (such as Win11 Visual Studio 2022), the long type is 32 bit, while the time_t type is 64 bit.

So in util.cpp>parseCookies()>`const std::time_t expires = static_cast<time_t>(std::stoul(tokens.at(static_cast<size_t>(CurlHTTPCookieField::Expires))));`, there may be an "stoul argument out of range" issue. Because `Cookie.ExpiresString` may be 64 bit, while the parameter for `std::stoul()` may be 32 bit.(From the source code of curl, it can be seen that the type of `Cookie.ExpiresString` is `time_t`)

So I added function `sTimestampToT()` to parse `Cookie.ExpiresString` based on different types of `time_t`.

What do everyone think? Welcome to correct.

Wishing everyone a happy life.

### DIFF
--- a/cpr/util.cpp
+++ b/cpr/util.cpp
@@ -12,6 +12,7 @@
 #include <ios>
 #include <sstream>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #if defined(_Win32)
@@ -53,7 +54,7 @@ Cookies parseCookies(curl_slist* raw_cookies) {
         while (tokens.size() < CURL_HTTP_COOKIE_SIZE) {
             tokens.emplace_back("");
         }
-        const std::time_t expires = static_cast<time_t>(std::stoul(tokens.at(static_cast<size_t>(CurlHTTPCookieField::Expires))));
+        const std::time_t expires = sTimestampToT(tokens.at(static_cast<size_t>(CurlHTTPCookieField::Expires)));
         cookies.emplace_back(Cookie{
                 tokens.at(static_cast<size_t>(CurlHTTPCookieField::Name)),
                 tokens.at(static_cast<size_t>(CurlHTTPCookieField::Value)),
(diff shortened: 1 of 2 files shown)


include/cpr/util.h
```diff
@@ -42,6 +42,11 @@ std::string urlDecode(const std::string& s);
 void secureStringClear(std::string& s);
 bool isTrue(const std::string& s);
 
+/**
+ * Parse string timestamp to time_t
```
COM8, line 46:
```suggestion
 * Parses the given std::string into time_t (unix ms).
 * This parsing happens time_t size agnostic since time_t does not use the same underlying type on all systems/compilers.
```

COM8 (changes requested):
Looks good thanks. Only one wording change and we are ready to go.

0x703cd8:
你好！      感谢你的来信，我会在查阅后尽快回复你。（自动回复）

cpr/util.cpp
```diff
@@ -237,4 +237,18 @@ bool isTrue(const std::string& s) {
     return temp_string == "true";
 }
 
+time_t sTimestampToT(const std::string& st){
+    if (std::is_same<time_t, unsigned long>::value){
```
COM8, line 241:
```suggestion
    if (std::is_integral_v<time_t, unsigned long>){
```

This is the more modern way of doing this. Please update all others as well.

0x703cd8: Hello COM8.
Thank you for your suggestion.
It seems that `std::is_integral_v` can only pass in one parameter.And it can only determine whether it is of type int,but cannot distinguish between different types of int,Is that so?
Welcome to correct.

include/cpr/util.h
0x703cd8: Very good annotation, I will modify it immediately.

cpr/util.cpp
COM8: Oh, sorry I meant `std::is_same_v<time_t, unsigned long>`.

0x703cd8: Okay, I understand. I'll make the changes right away.

COM8:
Perfect. Thanks. Now just a round of clang-tidy fixes and we are ready to go:
https://github.com/libcpr/cpr/actions/runs/8466060069/job/23195757541?pr=1033

0x703cd8:
> Perfect. Thanks. Now just a round of clang-tidy fixes and we are ready to go: https://github.com/libcpr/cpr/actions/runs/8466060069/job/23195757541?pr=1033
> 
> Are you able to access the build output behind this link?

I can see it, I'll go take a look

0x703cd8:
> Perfect. Thanks. Now just a round of clang-tidy fixes and we are ready to go: https://github.com/libcpr/cpr/actions/runs/8466060069/job/23195757541?pr=1033
> 
> Are you able to access the build output behind this link?

A troublesome problem has arisen: `error: consider replacing 'long' with 'int64' [google-runtime-int,-warnings-as-errors]`.
I need some time to figure out a solution.

COM8:
I quickly fixed all the open clang-tidy issues. Let me know if those changes are OK for you and then we are ready to merge your PR :)

0x703cd8:
> I quickly fixed all the open clang-tidy issues. Let me know if those changes are OK for you and then we are ready to merge your PR :)

No problem, then let's merge it.

0x703cd8:
> I quickly fixed all the open clang-tidy issues. Let me know if those changes are OK for you and then we are ready to merge your PR :)

No problem, then you can merge it.

PR #1042, opened 2024-04-10
Use std::call_once instead of double-checked locking in singletons

### LINKED ISSUE #1041: Possible race condition in singleton class GetInstance() function
### Description

I looked at the implementation of singleton classes and noticed that there is a [check for a null instance pointer](https://github.com/libcpr/cpr/blob/11e595428cb824bf24b9d70b286a69e9568b92da/include/cpr/singleton.h#L29) before the mutex is locked for initializing the singleton instance. I think this might produce a race condition as it differs from the patterns shown at https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_C++11 for lazy initialization of singleton classes. The [paper describing ThreadSanitizer](https://research.google/pubs/threadsanitizer-data-race-detection-in-practice/) mentions double-checked locking as a common source of race conditions because of possible instruction reordering or stale hardware cache memory.

I was troubleshooting a segmentation fault that seemed to be resolved when I used a local `cpr::ThreadPool` object instead of the singleton `cpr::GlobalThreadPool`, but I now think the memory management issues were unrelated to this library.

### Example/How to Reproduce

This is a hypothetical issue and I have not needed to write code that could trigger it.

### Possible Fix

_No response_

GitHub (branch e.g. master)

- OS: Ubuntu
- Version: Ubuntu 22.04.4 LTS

bllanos:
Fixes #1041

- I used `std::call_once` to initialize and destroy singleton classes.
- I added basic tests for singleton class initialization and shutdown functions.

## Notes

I added an assertion to the `ExitInstance()` function to help users detect if they call it multiple times in the same program by accident. Correct me if I am mistaken, but I assumed singleton classes do not need to be initialized and shutdown more than once. The previous version of the code allowed for multiple cycles of initialization and shutdown.

## Other

The CONTRIBUTING.md file mentions `format-check.sh` that I don't see in the repository, and it also links to a different location for the repository, `https://github.com/whoshuu/cpr`.

### DIFF
--- a/include/cpr/singleton.h
+++ b/include/cpr/singleton.h
@@ -1,6 +1,7 @@
 #ifndef CPR_SINGLETON_H
 #define CPR_SINGLETON_H
 
+#include <cassert>
 #include <mutex>
 
 #ifndef CPR_DISABLE_COPY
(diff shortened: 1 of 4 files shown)


include/cpr/singleton.h
```diff
@@ -10,38 +11,33 @@
...
+    Class* Class::s_pInstance = nullptr;                                     \
+    std::once_flag Class::s_getFlag{};                                       \
+    std::once_flag Class::s_exitFlag{};                                      \
+    Class* Class::GetInstance() {                                            \
+        std::call_once(Class::s_getFlag, []() { s_pInstance = new Class; }); \
+        return s_pInstance;                                                  \
+    }                                                                        \
+    void Class::ExitInstance() {                                             \
+        std::call_once(Class::s_exitFlag, []() {                             \
+            assert(s_pInstance != 0);                                        \
```
COM8, line 37:
```suggestion
            assert(s_pInstance0);                                            \
```

More cpr like null pointer check.

COM8 (changes requested):
@bllanos thank you very much for your PR! Looks good. Only a small assert code style change request ;)

COM8:
On your note: No, I'm not aware of any case where we use the singleton being initialized and destroyed multiple times. (Isn't this a bit against what a singleton is - in theory?).

Arg, yes the contribution docs. We've started rewriting them and extending them over [here](https://github.com/libcpr/docs/pull/37) but yeah. There is room for improvement.

COM8:
https://github.com/libcpr/cpr/issues/1047

PR #1046, opened 2024-04-28
add empty() for cpr::Cookies

xiaoloudongfeng:


cpr/cookies.cpp
```diff
@@ -100,6 +100,10 @@ void Cookies::emplace_back(const Cookie& str) {
     cookies_.emplace_back(str);
 }
 
+bool Cookies::empty() {
```
COM8, line 103:
```suggestion
bool Cookies::empty() const {
```

I would suggest making it const.

include/cpr/cookies.h
```diff
@@ -80,6 +80,7 @@ class Cookies {
     const_iterator cbegin() const;
     const_iterator cend() const;
     void emplace_back(const Cookie& str);
+    bool empty();
```
COM8, line 83:
```suggestion
    [[nodiscard]] bool empty() const;
```

COM8 (changes requested):
@xiaoloudongfeng thanks! Please make it const and we are ready to go.

COM8 (approved):
Thanks!

PR #1055, opened 2024-05-27
fix: fallback from curl_multi_poll to curl_multi_wait before curl 7.66

TontonJambon:
According to [this article](https://curl.se/mail/lib-2019-07/0069.html), curl_multi_poll is a "drop-in function replacement" to curl_multi_wait that "waits more", and was introduced in [7.66](https://curl.se/mail/archive-2019-09/0002.html).

Replace call to curl_multi_poll by curl_multi_wait when curl version is less that 7.66.

### DIFF
--- a/cpr/multiperform.cpp
+++ b/cpr/multiperform.cpp
@@ -10,6 +10,7 @@
 #include <cstddef>
 #include <curl/curl.h>
 #include <curl/multi.h>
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Thanks!

COM8 (approved):

COM8 (approved):

PR #1078, opened 2024-07-06
add support for CURLAUTH_NEGOTIATE

lubomirjurukov:
Added login with CURLAUTH_NEGOTIATE so Windows kerberos authentication will be possible.
Tested under Windows 10 with libcurl 8.8.0


COM8 (approved):
Thanks!

PR #1081, opened 2024-07-10
The header and write callback now uses a std::string_view as argument for data to avoid copying

COM8:
## Old

```c++
std::function<bool(std::string data, intptr_t userdata)> callback;
```

## New

```c++
std::function<bool(const std::string_view& data, intptr_t userdata)> callback;
```

With the new way we avoid copying the payload.

### DIFF
--- a/include/cpr/callback.h
+++ b/include/cpr/callback.h
@@ -29,26 +29,26 @@ class HeaderCallback {
   public:
     HeaderCallback() = default;
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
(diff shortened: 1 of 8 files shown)


COM8:
Every time you make a call to the following function `data` will be copied.
```c++
std::function<bool(std::string data, intptr_t userdata)> callback;
```

This means if e.g. `data` is multiple MB of size this copying is awfully inefficient.

wonbinbk:
Should we pass string_view by value here instead of cosnt&?

https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

COM8:
I guess this was a classical case where I extremely successfully applied _over optimization_. Would you like to create a PR to convert `const std::string_view&` to `std::string_view`?

PR #1084, opened 2024-07-18
Handle multiple headers with the same name

TontonJambon:
According to the [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html), multiple headers with the same name can be found in an HTTP Request :
```
Example-Field: Foo, Bar
Example-Field: Baz

```

As the current implementation of parseHeader (util.cpp) uses a map, only the last value is kept. However, section 5.3 of linked RFC tells that we can just append the fields using comma SP as separator and keep the semantic. This is what I propose in this PR.


COM8:
@TontonJambon thank you very much for contributing this!

Sadly I have a few issues with it.
1. Just appending them with a `, ` is not convenient for parsing later on. Instead I suggest switching the `std::string` to a `std::vector` or `std::variant<std::string, std::vector>`?
2. Please fix all existing tests.
3. Please add new test cases for this feature.

COM8:
Closing for no response. Feel free to reopen in case you want to continue working on it.

PR #1089, opened 2024-07-27
README: add info about a port in MacPorts

barracuda156:
Non-functional change: add info about install option for macOS via MacPorts.


COM8 (approved):
@barracuda156 awesome thanks!

PR #1093, opened 2024-07-31
Ignore certificate revocation checks in case of missing or offline distribution points on Windows

treyn008:
Set the `CURLSSLOPT_REVOKE_BEST_EFFORT` option, if curl is on the version 7.70.0 or later.

The SSL option is equivalent to the `--ssl-revoke-best-effort` curl flag. This option results in ignoring errors regarding the revocation check being unable to take place. This is only an issue on Windows, due to this being the default behavior from schannel.

([source](https://lists.gnu.org/archive/html/gnunet-svn/2020-04/msg01007.html))

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -206,6 +206,11 @@ void Session::prepareCommonShared() {
     curl_easy_setopt(curl_->handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NATIVE_CA);
 #endif
 
+#if LIBCURL_VERSION_NUM >= 0x074600 // 7.70.0
+    // Ignore certificate revocation checks in case of missing or offline distribution points, on Windows:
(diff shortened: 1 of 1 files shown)


treyn008:
Looking at libcurl’s documentation, this commit should be updated to set both of the SSL options in a single call, with a long bit mask conditionally ORing both the NATIVE_CA and REVOKE_BEST_EFFORT bits, depending on libcurl’s version.

COM8:
@treyn008  thanks for this PR! To me this should be an option and not be enabled by default. Else this could be a potential security issue.

Perhaps make it a part of `cpr::SslOptions`. Also please create a PR with a bit of docs here: https://github.com/libcpr/docs

COM8:
Closing for no response. Feel free to reopen in case you want to continue working on this PR.

Issue #1100, opened 2024-08-26
Response.status_code is 0 and response.text is empty for some of my users


typeshi12:
### Description

Hello,

I am a developer for a product that uses CPR to access various apis. However, a good sum of my users (not me, however) are encountering an issue accessing my website through CPR / cURL. 

I set up some debug message boxes, and this is what it shows on two users:
![image](https://github.com/user-attachments/assets/46caf892-02fa-44e1-a54c-ab29f6ed18d3)

The code behind that messagebox is this:
 ```
auto response = cpr::Get(cpr::Url{ url });
if (response.status_code != 200) {
     std::string errorMessage = oxorany("API returned a ") + std::to_string(response.status_code) + oxorany(" error, with content") + response.text + oxorany(" . Can't access. Contact [redacted]");

     Warden::get_singleton()->display(errorMessage);
 }
```

Addtionally, I had the users run a curl request to the same endpoint, which brings up this error for both users:
`curl: (35) schannel: next InitializeSecurityContext failed: SEC_E_INVALID_TOKEN (0x80090308) - The token supplied to the function is invalid`

Now this may not even be a CPR issue but rather a cURL issue, but seeing this is the library I'm using, I thought I'd try here first.
And before you say its an ISP block / country block issue, it's not. While a Vpn DOES fix this issue, it's still clearly not a block as the users can access the api endpoints just fine in their browser WITHOUT a vpn:

![image](https://github.com/user-attachments/assets/4011977d-16af-47e2-9129-c3636636efd7)
![image](https://github.com/user-attachments/assets/4f49d604-23ca-4e83-9e7b-e8d0c9c48dac)

### Example/How to Reproduce

1. Set up a simple C++ project with this code
```
int main(){
    auto response = cpr::Get(cpr::Url{ "https://getsolara.dev/api/endpoint.json" });
     if (response.status_code != 200) {
         std::string errorMessage = "API returned a " + std::to_string(response.status_code) + " error, with content" + response.text + " . Can't access.";

         MessageBoxA(NULL, errorMessage.c_str(), "Error", MB_OK);

         
     }
    return 0;
}
```
2. Observe the output
3. Do `curl 'https://getsolara.dev/api/endpoint.json'` in a CMD prompt
4. Observe output

### Possible Fix

A noticed that with using a VPN, the vpn uses a different IP to connect to the API (there are three) rather than what is seemingly the default IP. Not experienced much with C++ but maybe you can try getting a request to the different IPs.

[...]

PR #1111, opened 2024-09-20
Curl 8.10.1

COM8:
As described in #1100 on Windows curl changed the way certificate verification works. This lead to a bunch of certificates being rejected.

This has been fixed in curl 8.10.0 again and therefore we need to update.


rminderhoud:
Thank you @COM8

PR #1118, opened 2024-09-30
Add generic request function in api.h

### LINKED ISSUE #1117: Enhance "template metaprogramming magic" for HTTP methods

Library provides good experience by providing variadic template interface in HTTP methods, however, you can't _template-ly_  select HTTP method to call.

For example, you have `myGet` function:

```cpp
template <class... Ts>
cpr::Response myGet(Ts&&... ts) {
    return cpr::Get(cpr::Bearer{"ACCESS_TOKEN"}, cpr::Header{{"Content-Type", "application/json; charset=utf-8"}});
}
```

It's great, the headers will always be in the request. However, you also need the same wrapper for POST method.
I can duplicate code, but it's lame.

The other way is to provide the function to `myGet` (ignore the name):

```cpp
template <typename F, class... Ts>
cpr::Response myGet(F f, Ts&&... ts) {
    return f(std::forward<Ts>(ts)..., cpr::Bearer{"ACCESS_TOKEN"}, cpr::Header{{"Content-Type", "application/json; charset=utf-8"}});
}
```

The solution will not make you happy:

```cpp
    cpr::Response r = myGet(cpr::Post<cpr::Url, cpr::Body, cpr::Bearer, cpr::Header>, cpr::Url{"http://www.httpbin.org/post?a=b"}, cpr::Body{"{\"a\": true}"});
```

You have to specify template parameters for `cpr::Post` and it looks worse than making a duplicate.

### Possible Solution

Add `HTTPMethod` and a universal template to make a request.

```cpp
#include <cpr/cpr.h>

namespace cpr {

enum class HTTPMethod {
    kGet,
    kHead,
    kPost,
    kPut,
    kDelete,
    kOptions,
    kPatch,
};

template <HTTPMethod method, typename... Ts>
Response Request(Ts&&... ts) {
    cpr::Session session;
    cpr::priv::set_option(session, std::forward<Ts>(ts)...);

    using enum HTTPMethod;

    if constexpr (method == kGet) {
        return session.Get();
    } else if constexpr (method == kHead) {
        return session.Head();
    } else if constexpr (method == kPost) {
        return session.Post();
    } else if constexpr (method == kPut) {
        return session.Put();
    } else if constexpr (method == kDelete) {
        return session.Delete();
    } else if constexpr (method == kOptions) {
        return session.Options();
    } else if constexpr (method == kPatch) {
        return session.Patch();
    } else {
        static_assert(method == kGet, "Unknown method");
    }
}

}

template <cpr::HTTPMethod method, class... Ts>
cpr::Response MyResponse(Ts&&... ts) {
    return Request<method>(std::forward<Ts>(ts)..., cpr::Bearer{"ACCESS_TOKEN"}, cpr::Header{{"Content-Type", "application/json; charset=utf-8"}});
}

int main() {
    cpr::Response r = MyResponse<cpr::HTTPMethod::kPost>(cpr::Url{"http://www.httpbin.org/post?a=b"}, cpr::Body{"{\"a\": true}"});

    std::cout << r.text << std::endl;
}
```

### Alternatives

- compile-time if
- duplicated code for each method

### Additional Context

_No response_

GitSparTV:
Provides generic function to call any HTTP method.

This helps to avoid such code on your side:

```cpp
template<class ...Ts>
cpr::Response MakeRequest(std::string_view method, Ts&& ...ts) { // could be enum, string, or bool. Whatever what you use to select different HTTP method
    if (method == "GET") {
        return cpr::Get(std::forward<Ts>(ts)..., cpr::Header{{"Authorization", "token"}});
    else if (method == "POST") {
        return cpr::Post(std::forward<Ts>(ts)..., cpr::Header{{"Authorization", "token"}});
    } else {
        // and so on...
    }
}
```

Now you can do this:

```cpp
    template <cpr::HTTPMethod method, typename... Args>
    cpr::Response MakeRequest(Args&&... args) {
        return cpr::Request<method>(std::forward<Args>(args)..., cpr::Header{{"Authorization", "token"}});
    }
```

and call it like this:

```cpp
cpr::Response r = MakeRequest<cpr::HTTPMethod::kPost>(cpr::Url{"http://www.httpbin.org/post?a=b"}, cpr::Body{"{\"a\": true}"});
```

Ideas before the merge:
- Do we need tests for this function?
- This function can replace all existing functions like Post(), Get(), and other! Should I refactor it?
- HTTPMethod doesn't have Download since it's not a method. Do we need it?
- Do we need Async and/or Multi versions?

Resolves #1117

### DIFF
--- a/include/cpr/api.h
+++ b/include/cpr/api.h
@@ -23,6 +23,16 @@ namespace cpr {
 
 using AsyncResponse = AsyncWrapper<Response>;
 
+enum class HTTPMethod {
+    GET,
+    HEAD,
+    POST,
+    PUT,
+    DELETE,
+    OPTIONS,
+    PATCH,
+};
+
 namespace priv {
 
 template <bool processed_header, typename CurrentType>
@@ -384,6 +394,31 @@ std::vector<AsyncWrapper<Response, true>> MultiPutAsync(Ts&&... ts) {
     std::vector<AsyncWrapper<Response, true>> ret{};
     priv::setup_multiasync<&cpr::Session::Put>(ret, std::forward<Ts>(ts)...);
     return ret;
+}
+
+template <HTTPMethod method, typename... Ts>
+Response Request(Ts&&... ts) {
+    cpr::Session session;
+    cpr::priv::set_option(session, std::forward<Ts>(ts)...);
+
+    if constexpr (method == HTTPMethod::GET) {
+        return session.Get();
+    } else if constexpr (method == HTTPMethod::HEAD) {
+        return session.Head();
+    } else if constexpr (method == HTTPMethod::POST) {
+        return session.Post();
+    } else if constexpr (method == HTTPMethod::PUT) {
+        return session.Put();
+    } else if constexpr (method == HTTPMethod::DELETE) {
+        return session.Delete();
+    } else if constexpr (method == HTTPMethod::OPTIONS) {
+        return session.Options();
+    } else if constexpr (method == HTTPMethod::PATCH) {
+        return session.Patch();
+    } else {
+        // Must be template dependent until DR 2518 (C++23)
+        static_assert(method == HTTPMethod::GET, "Unknown method");
+    }
 }
 
 


COM8 (changes requested):
@GitSparTV Tanks for this PR! Looks good.
Could you please add a few test (examples) here: https://github.com/libcpr/cpr/tree/master/test

And a bit of documentation here: https://github.com/libcpr/docs

COM8:
@GitSparTV I'm closing this one for now since things are missing. Feel free to reopen in case you want to finish this PR.

PR #1125, opened 2024-10-14
Added handling no_proxy override through Proxies

gregory-shklover:


cpr/session.cpp
```diff
@@ -180,6 +180,15 @@ void Session::prepareCommonShared() {
             curl_easy_setopt(curl_->handle, CURLOPT_PROXYPASSWORD, proxyAuth_.GetPassword(protocol));
         }
     }
+    // handle NO_PROXY override passed through Proxies object
+    // Example: Proxies{"no_proxy": ""} will override environment variable definition with an empty list
+    const std::string no_proxy[] = {"no_proxy", "NO_PROXY"};
```
COM8, line 185: Please use a `std::array` not a C-Style Array.

cpr/session.cpp
```diff
@@ -180,6 +180,15 @@ void Session::prepareCommonShared() {
...
+    for (const auto& item : no_proxy) {
```
COM8, line 186: How about `std::any_of`?

test/proxy_tests.cpp
```diff
@@ -86,6 +86,25 @@ TEST(ProxyTests, ReferenceProxySessionTest) {
     EXPECT_EQ(ErrorCode::OK, response.error.code);
 }
 
+if 0
+TEST(ProxyTests, NoProxyTest) {
```
COM8, line 90: Please properly test your changes and let me know how I can verify them.

COM8 (changes requested):
Thanks for the PR!
A few comments and how about an empty string for no proxy?

Please also add a docs PR here: https://github.com/libcpr/docs

cpr/session.cpp
gregory-shklover: Does not seem that any_of would add readability.

test/proxy_tests.cpp
gregory-shklover: Added a test.
The test needs direct connection to the web + free proxy.

gregory-shklover:
Added documentation pull request: https://github.com/libcpr/docs/pull/51/files

COM8:
Thanks!

PR #1131, opened 2024-10-26
`Refactored` timeout.h: added template ctr and removed redundant ctrs (#1129)

### LINKED ISSUE #1129: Why does cpr::Timeout has two chrono bases ctors?
Timeout(const std::chrono::milliseconds& duration) 
Timeout(const std::chrono::seconds& duration)

Due to we have implicitly specified two chrono types, doing this will lead to inability to select proper ctor as mins\hours etc is not implicitly specified:

`const cpr::Timeout timeout{ 5min };`

Shouldn't we just create only ms ctor as any chrono duration type would be automatically casted to input parameter type?

laxerhd:
I have refactored the constructors of the `Timeout` class:
- Added template constructor that accepts any `std::chrono::duration` type and converts it to `std::chrono::milliseconds`
- Removed two redundant constructors for `milliseconds` and `seconds`
- Ctr for `int32_t` was untouched

I am not sure if we still want the `NOLINTNEXTLINE(...)` above the template ctr, but for consistency I have added it back.

- [x] I have made sure to run all test cases like described in the readme.

Fixes #1129.

### DIFF
--- a/include/cpr/timeout.h
+++ b/include/cpr/timeout.h
@@ -8,12 +8,14 @@ namespace cpr {
 
 class Timeout {
   public:
+    // Template constructor to accept any chrono duration type and convert it to milliseconds
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
-    Timeout(const std::chrono::milliseconds& duration) : ms{duration} {}
+    template <typename Rep, typename Period>
+    Timeout(const std::chrono::duration<Rep, Period>& duration)
+            : ms{std::chrono::duration_cast<std::chrono::milliseconds>(duration)} {}
+
     // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
     Timeout(const std::int32_t& milliseconds) : Timeout{std::chrono::milliseconds(milliseconds)} {}
-    // NOLINTNEXTLINE(google-explicit-constructor, hicpp-explicit-conversions)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,7 @@
 ## Announcements
 
 * This project is being maintained by [Fabian Sauter](https://github.com/com8) and [Kilian Traub](https://github.com/KingKili).
-* For quick help, and discussion libcpr also offer a [gitter](https://gitter.im/libcpr/community?utm_source=share-link&utm_medium=link&utm_campaign=share-link) chat.
+* For quick help, and discussion libcpr also offers a [gitter](https://gitter.im/libcpr/community?utm_source=share-link&utm_medium=link&utm_campaign=share-link) chat.
```
COM8, line 10: This stuff should not be included. As the other PR is now merged, try rebasing on the latest master or drop the commit if possible.

COM8 (changes requested):

README.md
laxerhd: Was a mistake on my side, I didn't want to include it here...

laxerhd:
The latest commit should now be dropped.

COM8 (approved):
Approved. Thanks.

To answer your question:
The NOLINT is still required since else clang-tidy will complain we should add `explicit` in front of the ctr to prevent an C++ offering an implicit cast from a Chrono duration to a timeout.

But in this case we want an implicit conversation to be possible.

PR #1135, opened 2024-11-05
Find the path to the certificate bundle in runtime

moratom:
The change makes the compiled library work on
different distributions, even if the path to the
certificate bundle in runtime is different compared to the path where it was during compilation.

This can be useful when the compilation is done on a manylinux Docker image which is based on CentOS, but the compiled library/binaries are used on
Ubuntu or Debian.

Related issue: https://github.com/curl/curl/issues/11411

Maintainers of CURL were not interested in mainlining this, so I understand if this is not suitable for the default behavior, but it would be nice to be able to at least optionally turn that on in runtime.

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -5,6 +5,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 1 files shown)


COM8:
@moratom thanks for submitting this PR!
Sadly I have to say, I'm on the same side as the curl developers here.

cpr provides a way for you to set the ca path manually during runtime e.g. via:
```c++
cpr::SslOptions opts;
opts.ca_info = "/some/path.ca";

cpr::Response r =   cpr::Get(opts, ...);
```

Using "random" paths on different distros is a security risk in case they are compromised since out of the box the directory structure has not the correct permissions set since this path is not intended to be used by this distro for that.

moratom:
Thanks for taking a look @COM8!

>Using "random" paths on different distros is a security risk in case they are compromised since out of the box the directory structure has not the correct permissions set since this path is not intended to be used by this distro for that.

That's a very good point.
Out of curiosity - is there a safer way from the top of your head to accomplish the same?

The specific use-case we have is that we use CPR&Curl in a project with Python bindings.

We provide manylinux wheels, which means we only compile once and it then generally works on *most* distributions. CPR/Curl is where this breaks as the wheels afterwards only work on Centos based distributions, but not on Ubuntu for example.
More accurately, the wheels still work, but the downloading feature doesn't.

COM8:
In my eyes this runtime check on startup is the only way. If you want to go the full way perhaps take `/etc/os-release` also into account for you educated ca path "guesses".

PR #1142, opened 2024-11-13
`Enhance`: Use unordered_map for CURL error mapping

### LINKED ISSUE #1140: `Improving Error Handling`: Missing PARTIAL_FILE and long `switch-case` refactoring
### Description

While taking a look at `error.h`, I noticed that the `ErrorCode` enum entry `PARTIAL_FILE` is not used in the `error.cpp` switch statement. Is this intended? 

I also wanted to ask if there was any specific reason for including `CURLE_OBSOLETE46` while not adding the other obsolete cURL error codes? Are the others related to unsupported protocols? 

`PARTIAL_FILE` and `CURLE_OBSOLETE46` were both added in this [commit](https://github.com/libcpr/cpr/pull/1110/commits/c698b9a2f5f457712973417b11b6ffb7ee6700f8) when adding better cpr error codes in september.

Additionally `error.cpp` looks kind of scarry with its 200-line `switch-case` statement. I was wondering if using an `unordered_map` instead of the `switch-case` statement might make the code a bit more readable and easier to maintain.

For example:
```cpp
static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = {
    {CURLE_OK, ErrorCode::OK},
    {CURLE_UNSUPPORTED_PROTOCOL, ErrorCode::UNSUPPORTED_PROTOCOL},
    // -- AND FOR ERROR CODES ONLY SUPPORTED IN SPECIFIC LIBCURL_VERSION --
#if LIBCURL_VERSION_NUM >= 0x074E00
    {CURLE_SETOPT_OPTION_SYNTAX, ErrorCode::SETOPT_OPTION_SYNTAX},
#endif
    ...
}
```

### Example/How to Reproduce

Files to take a look at: `error.cpp` and `error.h` as well as this [commit](https://github.com/libcpr/cpr/pull/1110/commits/c698b9a2f5f457712973417b11b6ffb7ee6700f8)

### Possible Fix

1. Adding `PARTIAL_FILE`-case to `error.cpp`
2. Include/Remove obsolete error codes for consistency
3. Refactor `errror.cpp` to use map instead of a `switch-case`

GitHub (branch e.g. master)

- OS: -
- Version: -

laxerhd:
Here is the corresponding PR which should close #1140, regarding some enhancements to the error handling in cpr.
I replaced the switch-case statement used for CURL error mapping with an `unordered_map`.

When using clang-tidy I am warned that the `initialization of 'curl_error_map' with static storage duration may throw an exception that cannot be caught`. 

I am unsure how we want to solve this:
- I could resolve this by introducing lazy initialization, wrapping the unordered_map inside a function that guarantees safe **runtime** initialization _(but would probably make it a little bit harder to read)_
- or, just ignore the message with `// NOLINT` _(I don't like that)_

The first solution would probably look something like this: 
```cpp
// Using this function instead of the curl_error_map variable
const std::unordered_map<std::int32_t, ErrorCode>& getCurlErrorMap() {
    static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = {
        {CURLE_OK, ErrorCode::OK},
        {CURLE_UNSUPPORTED_PROTOCOL, ErrorCode::UNSUPPORTED_PROTOCOL},
        {CURLE_FAILED_INIT, ErrorCode::FAILED_INIT},
        ...,
        };
    return curl_error_map;
}        
```
and then call this function in the `Error::getErrorCodeForCurlError`.

Are there any other options you would suggest?

Also, I wasn't sure for which error codes we include version checking (and how to find out when they were released), but I could find some mentions of `PARTIAL_FILE` in some changelogs in [early 2003](https://curl.se/changes.html), so I didn't add any checks for that.


laxerhd:
Like I already said, I am not sure how to proceed with the clang-tidy warning.

COM8:
> Like I already said, I am not quite sure how to proceed with the clang-tidy warning.

This is due to an updated version of clang-tidy. I fixed it in this PR: https://github.com/libcpr/cpr/pull/1144

Once merged, you can rebase onto the latest master.

COM8 (changes requested):
Awesome. Looks good! Please reabase onto: https://github.com/libcpr/cpr/pull/1144

laxerhd:
I hope I rebased it the right way... ;P

cpr/error.cpp
```diff
@@ -1,202 +1,165 @@
 #include "cpr/error.h"
-#include <cstdint>
-#include <curl/curl.h>
```
COM8, line 3: I guess you are missing those two includes for clang-tidy to work.

COM8 (changes requested):
Thanks for rebasing, but something went wrong there. You now also include the commits from my MR. This should not be the case.

How did you rebase?
The normal workflow for this would be:
* Add the original repo (this one) as e.g. `forkOrigin`.
* Rebase onto `forkOrigin/master`.
* `git push --force-with-lease`

PS: Top check which symbols are available in which curl version: https://github.com/curl/curl/blob/master/docs/libcurl/symbols-in-versions

cpr/error.cpp
laxerhd: Oh, I removed them because there was apparant reason for them. 
I will add them back, but why does clang-tidy need them?

laxerhd:
Did the cleanup.
Your commits should not be included anymore...

COM8 (approved):
Thanks! LGTM!

Issue #1150, opened 2024-12-04
Private key passed as a blob (ssl::KeyBlob) not working


b3rgschu3tz:
### Description

With version 1.11.1 passing a private key as blob as described in the docs (https://docs.libcpr.org/advanced-usage.html) does not work. The following error is returned:
`unable to set private key file: '(memory blob)' type PEM`

### Example/How to Reproduce

Like in the example:
```
cpr::SslOptions sslOpts = cpr::Ssl(ssl::CertFile{"cert.pem"}, ssl::KeyBlob{"-----BEGIN RSA PRIVATE KEY-----[...]"});
cpr::Response r = cpr::Get(cpr::Url{"https://www.httpbin.org/get"}, sslOpts);
```

### Possible Fix

In the session.cpp file https://github.com/libcpr/cpr/blob/master/cpr/session.cpp#L471 the local string `key_blob` is assigned to the `blob.data` value. 
```cpp
...
else if (!options.key_blob.empty()) {
        std::string key_blob(options.key_blob);
        curl_blob blob{};
        // NOLINTNEXTLINE (readability-container-data-pointer)
        blob.data = &key_blob[0];
        blob.len = key_blob.length();
        curl_easy_setopt(curl_->handle, CURLOPT_SSLKEY_BLOB, &blob);
        if (!options.key_type.empty()) {
            curl_easy_setopt(curl_->handle, CURLOPT_SSLKEYTYPE, options.key_type.c_str());
        }
        if (!options.key_pass.empty()) {
            curl_easy_setopt(curl_->handle, CURLOPT_KEYPASSWD, options.key_pass.c_str());
        }
...
```

In the [curl documentation](https://curl.se/libcurl/c/CURLOPT_SSLKEY_BLOB.html) it is stated:
_If the blob is initialized with the flags member of struct curl_blob set to CURL_BLOB_COPY, the application does not have to keep the buffer around after setting this._
=> If it CURL_BLOB_COPY is not set the buffer has to be kept alive until the request is performed.
As far is I can understand `key_blob` will go out of scope and be destroyed after the if-branch, thus `CURL_BLOB_COPY` needs to be set. 

When set, it works as expected.

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS: Debian
- Version: 1.11.1

COM8:
@b3rgschu3tz thanks for reporting! Confirmed. This is a bug. Would you like to create a PR for fix this?

[...]

PR #1151, opened 2024-12-04
[BUG] Fix cpr::ssl:KeyBlob: Copy blob to curl

b3rgschu3tz:
This fix addresses https://github.com/libcpr/cpr/issues/1150 
When passing a private key as blob as described in the docs (https://docs.libcpr.org/advanced-usage.html) the following error was returned for version 1.11.0:
_unable to set private key file: '(memory blob)' type PEM_


COM8 (approved):
Thanks @b3rgschu3tz!
LGTM

Issue #1165, opened 2025-01-25
Some tests fail with libcurl 8.11.1


Jerry-Terrasse:
### Description

When running tests after building cpr, I get:

```
81% tests passed, 5 tests failed out of 27

Total Test time (real) =  23.41 sec

The following tests FAILED:
	  1 - cpr_get_tests (Failed)
	  2 - cpr_post_tests (Failed)
	  6 - cpr_head_tests (Failed)
	 10 - cpr_raw_body_tests (Failed)
	 13 - cpr_error_tests (Failed)
```

[LastTest.log](https://github.com/user-attachments/files/18545123/LastTest.log)

### Example/How to Reproduce

```shell
git clone git@github.com:libcpr/cpr.git
cd cpr && mkdir build && cd build
cmake .. -DCPR_USE_SYSTEM_CURL=ON -DBUILD_SHARED_LIBS=OFF -DCPR_USE_SYSTEM_GTEST=ON -DCPR_BUILD_TESTS=ON -Wno-dev
cmake --build . --parallel
ctest
```

### Possible Fix

I've had a look into `test/get_tests.cpp`, the test failed at line:

```cpp
EXPECT_EQ(ErrorCode::COULDNT_RESOLVE_HOST, response.error.code);
```

In LastTest.log:

```
Expected equality of these values:
  ErrorCode::COULDNT_RESOLVE_HOST
    Which is: 4-byte object <06-00 00-00>
  response.error.code
    Which is: 4-byte object <07-00 00-00>
```

Which means `libcurl` on my machine returns error code `7` instead of `6`:

```shell
$ curl --version
curl 8.11.1 (x86_64-pc-linux-gnu) libcurl/8.11.1 OpenSSL/3.4.0 zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.0 nghttp2/1.64.0 nghttp3/1.6.0
Release-Date: 2024-12-11
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTP3 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
$ curl "http://bad_host/"
[...]

PR #1166, opened 2025-01-25
fix: let bad-host-tests pass when there is DNS error redirection

Jerry-Terrasse:
## Reason

In bad host tests, when the DNS server is configured with error redirection, it may return a fake IP such as `127.0.0.2` instead of an NXDOMAIN response, which causes the tests fail. #1165 

## Solution

Add a "OR" condition to those tests, both `CURLE_COULDNT_RESOLVE_HOST (6)` and `CURLE_COULDNT_CONNECT (7)` should pass.


COM8 (approved):
Awesome! Thanks for taking care of this!

Issue #1040, opened 2024-04-07
Thread Pool Deadlock While Task Wait


COM8:
### Description

There are potential cases where a `cpr::ThreadPool` can get stuck in a potential deadlock. They are both caused by a race condition involving `task_cond`.

It can happen that a thread gets stuck waiting for `task_cond` to get notified but during this there is no one who can notify him like if the caller invokes `cpr::ThreadPool::Wait()`. 

### Example/How to Reproduce

Run the currently disabled `ThreadPoolTests` from within https://github.com/libcpr/cpr/issues/1035 over and over again. From time to time they will get stuck.

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

Fedora 38, https://github.com/libcpr/cpr/issues/1035

baderouaich:
The ThreadPool being stuck I encountered, It was never stopping, it seems like the issue has to do with the scope of the status_locks, if you notice the Stop() method:
```cpp
int ThreadPool::Stop() {
    std::unique_lock status_lock(status_wait_mutex); // <- this will lock the entire scope of Stop(), after notifying, 
// the mutex is still held during the notification which will not give the threads a chance to acquire it and get the notification (in the thread loop)
    
    if (STOP == status) {
      return -1;
    }
    status = STOP;
    status_wait_cond.notify_all();
    task_cond.notify_all();
    
    for (auto& i : threads) {
        if (i.thread->joinable()) {
            i.thread->join();
        }
    }

    threads.clear();
    cur_thread_num = 0;
    idle_thread_num = 0;
    
    return 0;
}
```

 I think it should be:

```cpp
int ThreadPool::Stop() {
     { 
         // lock only this scope to ensure that the mutex is not held during the notification process, allowing the waiting threads to acquire it immediately after being awakened
          std::unique_lock status_lock(status_wait_mutex);
          if (STOP == status) {
            return -1;
          }
          status = STOP;
     }
    status_wait_cond.notify_all();
    task_cond.notify_all();

 ...
}
```

I tried that locally and it seems the thread pool is waiting and stopping normally so far, I will test with more use cases and get back to you if there is another issue.

Issue #988, opened 2023-12-04
Force the end of the thread pool


204065248:
### Is your feature request related to a problem?

Force the end of the thread pool to clean up unexecuted tasks when the program exits

### Possible Solution

cpr::clear(); ?

COM8:
The thread pool is not intended by design to be shut down from the outside. It will be cleaned up manually (except the currently running tasks) when exiting the application. This is actually a bug...

This needs to change so we implicitly also close those/cancel those jobs.

For now please make sure you cancel/await all unfinished tasks before shutting down your application.

PR #1168, opened 2025-01-26
WIP: Threadpool Refactor

COM8:
Refactors `cpr::ThreadPool` to get rid of all those deadlocks and overly complex code making it at least a bit simpler.

### Fixes:
* https://github.com/libcpr/cpr/issues/1040
* https://github.com/libcpr/cpr/issues/988


portsip:
Hi, I would like to suggest use a third thread pool implementation: https://github.com/bshoshany/thread-pool

COM8:
> Hi, I would like to suggest use a third thread pool implementation: https://github.com/bshoshany/thread-pool

Agree. If I'm unable to fix the macos socket issues, I will most likely switch and experiment with that one. Good point.

roemil:
> The ASIO thread pool is also a good choice.

Yes! With ASIO threadpool we could get an awaitable Get/Post and use that within a boost coroutine. That would be nice to have.

PR #1170, opened 2025-02-03
Replaced the secureStringClear mechanism with a SecureString class

mikael-s-persson:
Replaced the secureStringClear mechanism with a SecureString class

The existing mechanism isn't fully secure. The idea was to fill the old memory of sensitive strings (e.g., auth) with zeros before deallocating the memory. However, the implementation, via the secureStringClear, only really did this on the final destruction of certain objects containing those sensitive strings. That is not sufficient because assignment operations or temporary strings (e.g., urlEncode) would not get the same treatment (e.g. not clearing the old memory deallocated during assignment).

As far as I know, a more reliable (and convenient) way of doing this is through a custom allocator that zeroes out any deallocated memory, and then, define a SecureString as simply a std::string that uses that secure allocator. So, that's what I implemented in this PR.


COM8:
Oh wow @mikael-s-persson! Thank you very much for your great work! 

I agree this makes total sense to do this with a custom allocator. That way we can by far better control clearing the sensitive data. 

I will try to give you some feedback and make a review over the next weekend.

mikael-s-persson:
My pleasure, just a simple fix in passing.

Please note that this PR causes some breaking changes in API, like `string_view` instead of `const string&` and things like that, but I think those are unlikely to cause much breakage for users and are very simple to fix up in case of breaking when upgrading.

COM8 (approved):
Thanks LGTM!
Also the proposed API changes are in my eyes nothing major.

The CI failures are expected and I will fix them soon on master.

L-Super:
After I upgraded CPR, I found this breaking change. I checked the commit and don't think it's a good idea. If the concern is about sensitive data, just clean it up directly. There's no need to introduce an additional `SecureString` class.

COM8:
@L-Super thanks for raising your concerns about this. In my eyes there are only benefits to this approach besides the breaking nature of this change.
Best case the string gets cleared and since we are doing it via a custom allocator we do not have to care about it.
Worst case it's just a string-like type you can use like any other string like type.

L-Super:
> @L-Super thanks for raising your concerns about this. In my eyes there are only benefits to this approach besides the breaking nature of this change. Best case the string gets cleared and since we are doing it via a custom allocator we do not have to care about it. Worst case it's just a string-like type you can use like any other string like type.

After carefully analyzing this issue, I understood the intent. The root cause is that when std::string is destructed, the OS only marks the memory as free instead of erasing the data, which poses a security risk.

Issue #1172, opened 2025-02-03
Session: Delete one ore more select headers


simue:
### Is your feature request related to a problem?

I'd like there to be the possibility to delete one or more headers when using sessions.
Currently, we can overwrite all headers with `SetHeader`, or update select ones with `UpdateHeader`. What's missing is to delete one or more headers from `header_`.

### Possible Solution

I'd like to either implement a function `bool Session::DeleteHeader(const std::string& headerKey)`, or a getter for the headers: `Header& GetHeaders()`, with which the user can delete (and inspect) the existing headers that may have been set with previous requests.

What's your take on this - does it make sense, or is this the wrong approach?

PR #1173, opened 2025-02-05
Getter function for Session::header_ to enable the user to read back all headers set and delete select ones

simue:
This PR implements a possible solution to #1172

### DIFF
--- a/cpr/error.cpp
+++ b/cpr/error.cpp
@@ -5,8 +5,8 @@
 #include <unordered_map>
 
 namespace cpr {
+// NOLINTNEXTLINE - (needed because of static init)
 static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = {
-        // NOLINT - (needed because of static init)
         {CURLE_OK, ErrorCode::OK},
         {CURLE_UNSUPPORTED_PROTOCOL, ErrorCode::UNSUPPORTED_PROTOCOL},
         {CURLE_FAILED_INIT, ErrorCode::FAILED_INIT},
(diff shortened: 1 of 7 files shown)


COM8:
@simue thanks for contributing! I will try to review it next weekend.

cpr/error.cpp
```diff
@@ -1,11 +1,12 @@
...
-#include <unordered_map>
 #include <cstdint>
 #include <curl/curl.h>
+#include <curl/curlver.h>
+#include <unordered_map>
 
 namespace cpr {
-static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = { // NOLINT - (needed because of static init)
+static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = {
+        // NOLINT - (needed because of static init)
```
COM8, line 9:
```suggestion
// NOLINTNEXTLINE - (needed because of static init)
static const std::unordered_map<std::int32_t, ErrorCode> curl_error_map = {
```

That way we get around the formatting issue.

include/cpr/session.h
```diff
@@ -66,6 +66,7 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetParameters(Parameters&& parameters);
     void SetHeader(const Header& header);
     void UpdateHeader(const Header& header);
+    Header& GetHeader();
```
COM8, line 69:
```suggestion
    Header& GetHeader();
    [[nodiscard]] const Header& GetHeader() const;
```

Could you please also add a const version of this function?

COM8 (changes requested):
Only two minor things. Thank you very much for submitting this PR!

COM8 (approved):
LGTM! Thanks!

PR #1178, opened 2025-02-12
Status code int32_t -> long

liyishuai:
per https://github.com/libcpr/cpr/pull/325#discussion_r324968409:
> The current Response class uses int32_t for the status code, so this should probably match up.

350b0f81e6ab70b47e0eef7089b68da07ff491e9 changed the `status_code` from `int32_t` to `long`. This PR is to align the datatypes.


COM8 (approved):
Thanks! LGTM.

PR #1179, opened 2025-02-13
Add Session::RemoveContent()

### LINKED ISSUE #1171: Using sessions, the body/payload/multipart of the previous request is sent if not replaced.
### Description

I am using a session to send multiple sequential requests. Some are GET, some are POST.
On a GET that is sent after a POST, I see that the `content_` of the first request being used as custom method, see example below.

```
[03/Feb/2025 22:33:14] code 501, message Unsupported method ('{"thisIsMyJsonBody": "OfThePreviousPostRequest"}GET')
```

I'd expect the session to either internally reset `content_`(maybe in `makeRequest`?), or if not possible, I'd expect there to be the possibility of resetting `content_` explicitly.
Since `content_`'s default is std::monostate, I'd like to introduce a reset function, setting std::monostate.

Interestingly enough, there already is a function `hasBodyOrPayload` which generates a custom GET request and a comment stating GET with body is illegal. But no setter/reset for `content_`.

### Example/How to Reproduce

1. Create a `cpr::Session`
2. Send POST
3. Send GET (without changing content_ via its various setters)
4. Perform the request
5. Observe method shown in server which contains the body of the POST request.

### Possible Fix

See above, I'd like to introduce a reset function for `content_`.

conan

- OS: Linux
- Version: custom on kernel 5.15

simue:
Closes #1171

### DIFF
--- a/include/cpr/session.h
+++ b/include/cpr/session.h
@@ -112,7 +112,16 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetAcceptEncoding(const AcceptEncoding& accept_encoding);
     void SetAcceptEncoding(AcceptEncoding&& accept_encoding);
     void SetLimitRate(const LimitRate& limit_rate);
+
+    /**
+     * Returns a reference to the content sent in previous request
+     */
     [[nodiscard]] const Content& GetContent() const;
+
+    /**
+     * Removes the content sent in previous request, so it will not be sent with the next request
+     * Call this before doing a request that is specified not to send a body, e.g. GET
+     */
     void RemoveContent();
 
     // For cancellable requests
--- a/include/cpr/session.h
+++ b/include/cpr/session.h
@@ -46,6 +46,7 @@
 namespace cpr {
 
 using AsyncResponse = AsyncWrapper<Response>;
+using Content = std::variant<std::monostate, cpr::Payload, cpr::Body, cpr::Multipart>;
 
 class Interceptor;
 class MultiPerform;
@@ -111,6 +112,8 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetAcceptEncoding(const AcceptEncoding& accept_encoding);
     void SetAcceptEncoding(AcceptEncoding&& accept_encoding);
     void SetLimitRate(const LimitRate& limit_rate);
+    [[nodiscard]] const Content& GetContent() const;
+    void RemoveContent();
 
     // For cancellable requests
     void SetCancellationParam(std::shared_ptr<std::atomic_bool> param);
@@ -237,7 +240,7 @@ class Session : public std::enable_shared_from_this<Session> {
 
 
     bool chunkedTransferEncoding_{false};
-    std::variant<std::monostate, cpr::Payload, cpr::Body, cpr::Multipart> content_{std::monostate{}};
+    Content content_{std::monostate{}};
     std::shared_ptr<CurlHolder> curl_;
     Url url_;
     Parameters parameters_;
--- a/include/cpr/session.h
+++ b/include/cpr/session.h
@@ -114,8 +114,8 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetLimitRate(const LimitRate& limit_rate);
 
     /**
-     * Returns a reference to the content sent in previous request
-     */
+      * Returns a reference to the content sent in previous request.
+      **/
     [[nodiscard]] const Content& GetContent() const;
 
     /**
--- a/include/cpr/session.h
+++ b/include/cpr/session.h
@@ -119,7 +119,7 @@ class Session : public std::enable_shared_from_this<Session> {
     [[nodiscard]] const Content& GetContent() const;
 
     /**
-     * Removes the content sent in previous request, so it will not be sent with the next request
+     * Removes the content sent in previous request from internal state, so it will not be sent with the next request
      * Call this before doing a request that is specified not to send a body, e.g. GET
      */
     void RemoveContent();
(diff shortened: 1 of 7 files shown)


COM8 (changes requested):
@simue perfect. Thanks!
Could you please add a bit of in line documentation for your two new functions like I'm doing it here:
https://github.com/libcpr/cpr/pull/1168

Also please add a few sentences of documentation about your feature here:
https://github.com/libcpr/docs

simue:
Now, that I am trying to document this, I am hesitating with the resulting interface:
1.per default, Session keeps the previous payload around.
2.user has to explicitly clear the internal state (if he knows the library well enough)

Do you know if people are actually using session to send the same payload multiple times? Because if they are not, I´d rather call RemoveContent() after every request that carries a body.

include/cpr/session.h
```diff
@@ -111,6 +112,8 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetAcceptEncoding(const AcceptEncoding& accept_encoding);
     void SetAcceptEncoding(AcceptEncoding&& accept_encoding);
     void SetLimitRate(const LimitRate& limit_rate);
+    [[nodiscard]] const Content& GetContent() const;
```
COM8, line 115: Sorry, meant docs here in the header :)

COM8 (changes requested):

include/cpr/session.h
simue: must have looked in the wrong header and thought you meant the source files :D Added some docs to the header!

simue:
MultiperformPerformTests.MultiperformMixedMethodsPerformTest seems to be flaky - did you see it have issues before?

COM8:
> Now, that I am trying to document this, I am hesitating with the resulting interface: 1.per default, Session keeps the previous payload around. 2.user has to explicitly clear the internal state (if he knows the library well enough)
> 
> Do you know if people are actually using session to send the same payload multiple times? Because if they are not, I´d rather call RemoveContent() after every request that carries a body.

Ah, yes :D
Yes, people are actually using this behaviour (and I'm as well for some of my API requests). Because of this and since it would be a really drastic change in how `cpr::Session` works I'd keep it as it is.

This always has been a bit of a discussion point but we decided to go this way. Even if it's the more intuitive way that in reality might almost always require more code to clear the content before reusing.

COM8:
> MultiperformPerformTests.MultiperformMixedMethodsPerformTest seems to be flaky - did you see it have issues before?

Yes, especially on macOS. This is a limitation of the mock server we are using for our tests.

include/cpr/session.h
```diff
@@ -112,6 +113,17 @@ class Session : public std::enable_shared_from_this<Session> {
     void SetAcceptEncoding(AcceptEncoding&& accept_encoding);
     void SetLimitRate(const LimitRate& limit_rate);
 
+    /**
+     * Returns a reference to the content sent in previous request
+     */
```
COM8, line 118:
```suggestion
    /**
      * Returns a reference to the content sent in previous request.
      **/
```

That way its the same as for all other comments. I know I need a linter for it at some point ;)

include/cpr/session.h
```diff
@@ -112,6 +113,17 @@ class Session : public std::enable_shared_from_this<Session> {
...
+    [[nodiscard]] const Content& GetContent() const;
+
+    /**
+     * Removes the content sent in previous request from internal state, so it will not be sent with the next request
+     * Call this before doing a request that is specified not to send a body, e.g. GET
+     */
```
COM8, line 124:
```suggestion
    /**
      * Removes the content sent in previous request from internal state, so it will not be sent with the next request.
      * Call this before doing a request that is specified not to send a body, e.g. GET.
      **/
```

test/session_tests.cpp
```diff
@@ -1005,9 +1005,34 @@ TEST(DifferentMethodTests, MultipleDeleteHeadPutGetPostTest) {
...
+            std::string expected_text{"Header reflect GET"};
+            EXPECT_EQ(expected_text, response.text);
+            EXPECT_EQ(url, response.url);
+            EXPECT_EQ(200, response.status_code);
+            EXPECT_EQ(ErrorCode::OK, response.error.code);
+        }
+        {
+            session.RemoveContent();
+            session.SetUrl(urlMultipartPost);
+            std::string fileContentsBinary{"kn3d0c9j213riknmfd039dj10idn2309cmxkle2rsdj0239mdf"};
```
COM8, line 1024: Any reason for this string `kn3d0c9j213riknmfd039dj10idn2309cmxkle2rsdj0239mdf`. Looks like it encodes something but I'm not sure what :D

COM8 (changes requested):
Thanks! 2 Minor things regarding your comments and the we are ready to go!

test/session_tests.cpp
COM8: Perhaps I'm just a bit too burned by the XZ CVE (https://en.wikipedia.org/wiki/XZ_Utils_backdoor).

simue: oh, sorry - not very thoughtful of me!
No, it's either part of an image payload, or by the looks of it, it might just be me punching on my keyboard. To be honest I don´t remember.

But of course you're totally right - I made it more human readable

COM8 (approved):

test/session_tests.cpp
COM8: Thanks!

COM8:
Approved, merged and is now part of the 1.11.x branch. I will try to make a new release over the next weekend.

simue:
thank you very much for the good collaboration!

PR #1181, opened 2025-02-14
Fix Seg-fault when setting proxy username + password

### LINKED ISSUE #1180: Segfault in Session::prepareCommonShared with proxy authentication
### Description

I get a segfault with the following call stack:
```
libc.so.6![Unknown/Just-In-Time compiled code] (Unknown Source:0)
Curl_setstropt(char ** charp, const char * s) (\home\user\.conan2\p\b\libcu0c6ab0202663c\b\src\lib\setopt.c:68)
Curl_vsetopt(struct Curl_easy * data, CURLoption option, va_list param) (\home\user\.conan2\p\b\libcu0c6ab0202663c\b\src\lib\setopt.c:1604)
curl_easy_setopt(struct Curl_easy * data, CURLoption tag) (\home\user\.conan2\p\b\libcu0c6ab0202663c\b\src\lib\setopt.c:3171)
cpr::Session::prepareCommonShared(cpr::Session * const this) (\home\user\repos\tmp\cpr\cpr\session.cpp:184)
cpr::Session::prepareCommon(cpr::Session * const this) (\home\user\repos\tmp\cpr\cpr\session.cpp:226)
cpr::Session::PreparePost(cpr::Session * const this) (\home\user\repos\tmp\cpr\cpr\session.cpp:860)
cpr::Session::Post(cpr::Session * const this) (\home\user\repos\tmp\cpr\cpr\session.cpp:748)
[...]
```

This happens when a proxy is set, along with proxy authentication.
Setting the proxy authentication manually in session.cpp resolved the issue for me:
```
    // Proxy:
    const std::string protocol = url_.str().substr(0, url_.str().find(':'));
    if (proxies_.has(protocol)) {
        curl_easy_setopt(curl_->handle, CURLOPT_PROXY, proxies_[protocol].c_str());
        if (proxyAuth_.has(protocol)) {
            // this worked:
            //  std::string username = "user";
            //  std::string password = "password";
            //  curl_easy_setopt(curl_->handle, CURLOPT_PROXYUSERNAME, username.c_str());
            //  curl_easy_setopt(curl_->handle, CURLOPT_PROXYPASSWORD, password.c_str());
            // this does trigger the segfault (original code):
            curl_easy_setopt(curl_->handle, CURLOPT_PROXYUSERNAME, proxyAuth_.GetUsername(protocol));
            curl_easy_setopt(curl_->handle, CURLOPT_PROXYPASSWORD, proxyAuth_.GetPassword(protocol));
        }
    }
```

### Example/How to Reproduce

1. Create a `cpr::Session`
2. Set proxy with proxy authentication
3. Perform POST
4. Segfault

### Possible Fix

I didn't look into that, yet

GitHub (branch e.g. master)

- OS: Linux
- Version: custom kernel 5.15

simue:
Providing a char* to underlying data of string_view instead string_view itself to curl api.
Fixes #1180

### DIFF
--- a/cpr/curlholder.cpp
+++ b/cpr/curlholder.cpp
@@ -1,9 +1,9 @@
 #include "cpr/curlholder.h"
+#include "cpr/secure_string.h"
 #include <cassert>
 #include <curl/curl.h>
 #include <curl/easy.h>
 #include <string_view>
-#include "cpr/secure_string.h"
 
 namespace cpr {
 CurlHolder::CurlHolder() {
--- a/cpr/proxyauth.cpp
+++ b/cpr/proxyauth.cpp
@@ -1,6 +1,7 @@
 #include "cpr/proxyauth.h"
 #include <string>
 #include <string_view>
+#include "cpr/secure_string.h"
 
 namespace cpr {
 
--- a/cpr/proxyauth.cpp
+++ b/cpr/proxyauth.cpp
@@ -12,6 +12,14 @@ std::string_view EncodedAuthentication::GetPassword() const {
     return password;
 }
 
+const util::SecureString& EncodedAuthentication::GetUsernameUnderlying() const {
+    return username;
+}
+
+const util::SecureString& EncodedAuthentication::GetPasswordUnderlying() const {
+    return password;
+}
+
 bool ProxyAuthentication::has(const std::string& protocol) const {
     return proxyAuth_.count(protocol) > 0;
 }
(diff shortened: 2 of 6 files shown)


COM8 (changes requested):
@simue thank you very much for fixing this!
Please also apply the change for those lines here:
https://github.com/libcpr/cpr/blob/master/cpr/session.cpp#L665

I also noticed that this is copy/paste code here so it could be singled out into its own function.

COM8:
Also the clang tidy warning can be ignored in this case although a better way would be to replace returning a `std::string_view` with `const std::string&` in my eyes.

COM8 (approved):
Well done. Thanks for interoperating those changes.

Issue #1184, opened 2025-02-20
add support for curl anyauth


hyOzd:
### Is your feature request related to a problem?

I'm trying to send SOAP requests to an ONVIF camera which requires digest authenticaton. When `--digest` option is selected curl initially sends an empty request to initiate authorization. Unfortunately this particular device doesn't reply to the request if its content is empty. Solution seems to be sending the full request content also in the initial request. The way to do that in curl seems to be `--anyauth` option. With this option curl sends the initial request with full content and it figures out the auth mechanism from the first reply. Then it sends the request again this time with the authorization information. 

### Possible Solution

Add an authorization option like `cpr::AuthMode::ANY` that sets `CURLAUTH_ANY` flag.

### Alternatives

As a workaround I can set the flag by accessing curl handle directly, not sure if this is official part of API or just something added for tests.

```
curl_easy_setopt(session.GetCurlHolder()->handle, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
```

COM8:
@hyOzd thanks for reporting! This sounds like a nice feature to have. Would you like to create a PR for it?

PR #1187, opened 2025-02-23
add curl's ANY and ANSAFE authorization options

hyOzd:
Implements #1184

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -409,6 +409,14 @@ void Session::SetAuth(const Authentication& auth) {
             curl_easy_setopt(curl_->handle, CURLOPT_HTTPAUTH, CURLAUTH_NEGOTIATE);
             curl_easy_setopt(curl_->handle, CURLOPT_USERPWD, auth.GetAuthString());
             break;
+        case AuthMode::ANY:
+            curl_easy_setopt(curl_->handle, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
+            curl_easy_setopt(curl_->handle, CURLOPT_USERPWD, auth.GetAuthString());
+            break;
+        case AuthMode::ANYSAFE:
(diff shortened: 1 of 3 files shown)


hyOzd:
I actually wanted to refactor `Session::SetAuth` a bit to reduce code duplication. Not sure if that would be desired?

COM8:
> I actually wanted to refactor `Session::SetAuth` a bit to reduce code duplication. Not sure if that would be desired?

What exactly are your plans?
If you got a good idea, you are always welcome to refactor it.
Nevertheless, I would suggest doing this in a follow-up PR if possible so we follow the rule, one PR per change.

hyOzd:
```
void Session::SetAuth(const Authentication& auth) {
    long auth_flag;
    // Ignore here since this has been defined by libcurl.
    switch (auth.GetAuthMode()) {
        case AuthMode::BASIC:
            auth_flag = CURLAUTH_BASIC;
            break;
        case AuthMode::DIGEST:
            auth_flag = CURLAUTH_DIGEST;
            break;
        case AuthMode::NTLM:
            auth_flag = CURLAUTH_NTLM;
            break;
        case AuthMode::NEGOTIATE:
            auth_flag = CURLAUTH_NEGOTIATE;
            break;
        case AuthMode::ANY:
            auth_flag = CURLAUTH_ANY;
            break;
        case AuthMode::ANYSAFE:
            auth_flag = CURLAUTH_ANYSAFE;
            break;
    }

    curl_easy_setopt(curl_->handle, CURLOPT_HTTPAUTH, auth_flag);
    curl_easy_setopt(curl_->handle, CURLOPT_USERPWD, auth.GetAuthString());    
}
```

Something like that. But I don't think it is worth opening a separate PR.

COM8:
Ah, I see. If you want, you can make a minimal PR with it. It would be a good amount of cleanup.

Else I will try to take care of it eventually.

PR #1188, opened 2025-02-23
Fixed memory leak in threadpool

zManu3k:
Transitioning from raw pointers (`std::thread*`) to `std::shared_ptr<std::thread>` to fix memory leak (there seems to be nothing that `delete` the raw ptr after the thread is terminated).

- `std::thread` objects now are wrapped in `std::shared_ptr` (so the memory is now automatically managed)
- `CreateThread` now creates thread using `std::make_shared`
- `AddThread` has been modified to accept `std::shared_ptr<std::thread>` instead of raw pointers

### DIFF
--- a/cpr/threadpool.cpp
+++ b/cpr/threadpool.cpp
@@ -2,7 +2,6 @@
 #include <algorithm>
 #include <chrono>
 #include <cstddef>
-#include <ctime>
 #include <memory>
 #include <mutex>
 #include <thread>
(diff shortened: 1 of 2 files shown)


COM8 (approved):
Oh, wow, thanks :D

Nice catch. That's why I started refactoring the thread pool: https://github.com/libcpr/cpr/pull/1168

I will merge it for the next 1.11.x release. 1.12.0 then will get the fully refactored thread pool once I dealt with the macOS issues in the CI.

COM8:
Cherry-picked to the `1.11.x` branch and will be part of the `1.11.3` release.

PR #1197, opened 2025-03-10
Add std::to_string functionality for ErrorCode to ease human meaningful logging

psomers3:


include/cpr/error.h
```diff
@@ -89,6 +90,77 @@ enum class ErrorCode {
     UNKNOWN_ERROR = 1000,
 };
 
+constexpr std::array<std::pair<ErrorCode, const char*>, 63> error_code_to_string_mapping = {{{ErrorCode::OK, "OK"},
```
COM8, line 93: I personally prefer using a `std::unordered_map` here since it is a bit more tolerant and robust against error codes not being in order.

COM8 (changes requested):
Good idea. I would love to get this in.

include/cpr/error.h
psomers3: you can't make it a constexpr though. I also would prefer that approach.

COM8: What is the value in making a constexpr version of `to_string` compared to a runtime version?
I agree it makes sense to have a constexpr version, but in my eyes this `to_string` functionality is more helpful if you get an error during runtime e.g. via a response object and then you want to print the error in a useful manner.

psomers3: I have moved to using `inline` instead of constexpr to avoid "duplicate definition" errors on compiling. I will need a more experienced c++ person to tell me if that is an inefficient thing to do. But now we have the map ;)

include/cpr/error.h
```diff
@@ -89,6 +90,71 @@ enum class ErrorCode {
     UNKNOWN_ERROR = 1000,
 };
 
+inline std::unordered_map<ErrorCode, std::string> error_code_to_string_mapping = {{ErrorCode::OK, "OK"},
```
COM8, line 93:
```suggestion
inline const std::unordered_map<ErrorCode, std::string> error_code_to_string_mapping = {{ErrorCode::OK, "OK"},
```

That's my only suggestion to prevent others from modifying it.

Other than that well done. Thanks for pushing it past the last mile!

COM8 (changes requested):

include/cpr/error.h
COM8: Your approach is good. I'd expect the compiler to implicitly inline the map if needed, but better to make it explicit.

include/cpr/error.h
psomers3: Done. 👍

COM8 (approved):
Thanks!

Issue #1198, opened 2025-03-13
Unable to select MultiThreaded or MultiThreadedDLL in MSVC environment


BestArthur:
### Description

Issue: Unable to select MultiThreaded or MultiThreadedDLL in MSVC environment

Description
In the CMakeLists.txt file, when BUILD_SHARED_LIBS is set to FALSE (OFF), the runtime library is forcibly set to MultiThreaded.
Instead of this hardcoded behavior, it would be better to use the CMAKE_MSVC_RUNTIME_LIBRARY option to allow MultiThreaded and MultiThreadedDLL as configurable build options.

Let me know if you need any modifications! 🚀

### Example/How to Reproduce

1. Create a `cpr::Session`
2. Set option ...
3. Perform the request

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS:
- Version:

PR #1199, opened 2025-03-13
feat: Use CMAKE_MSVC_RUNTIME_LIBRARY for runtime selection in MSVC(#1…

BestArthur:
feat: Use CMAKE_MSVC_RUNTIME_LIBRARY for runtime selection in MSVC(#1198)

Previously, when BUILD_SHARED_LIBS was set to FALSE (OFF), the runtime library was
forcibly set to MultiThreaded. This commit replaces that behavior by using the
CMAKE_MSVC_RUNTIME_LIBRARY option, allowing users to select between MultiThreaded
and MultiThreadedDLL as a build option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,6 @@ if (MSVC)
         message(STATUS "Build windows dynamic libs.")
     else()
         # Add this to build windows pure static library.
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Agree @BestArthur. This should be set by the user. Thanks for creating the MR.

PR #1205, opened 2025-03-29
Make cpr::async and HTTP (Get, Post, Put, etc.) callbacks cancelable

### LINKED ISSUE #916: Cancellable GetCallback and friends

I'm trying to implement request cancellation for some requests done via GetCallback. Historically, we'd have a shared_ptr to a flag that's then checked in the callback, but I noticed that 1.10 had some cancellation semantics, so I was interested in using those instead. However, it appears that GetCallback and friends don't support cancellation, only MultiGetAsync and friends do.

### Possible Solution

I'd like to see GetCallback be cancellable as MultiGetAsync is. It seems like cpr::async could return `AsyncWrapper<T, true>`, but the trivial change didn't compile and I haven't looked further into it. 

Alternatively, I'd be happy to use MultiGetCallback with 1 request, if such existed.

### Alternatives

For now, we can solve this on our end with some shared flag checked in the callback.

### Additional Context

_No response_

Anohkka:
cpr::async can be made cancelable with a template parameter, false by default to keep old behavior.

Closes #916.

### DIFF
--- a/include/cpr/api.h
+++ b/include/cpr/api.h
@@ -131,7 +131,7 @@ AsyncResponse GetAsync(Ts... ts) {
 template <typename Then, typename... Ts>
 // NOLINTNEXTLINE(fuchsia-trailing-return)
 auto GetCallback(Then then, Ts... ts) {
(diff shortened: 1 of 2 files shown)


COM8 (approved):
@Anohkka  thank you very much for taking care of this.
LGTM!

PR #1206, opened 2025-03-30
Refactor AsyncWrapper to make it safer

Anohkka:
It's currently possible to construct an unsafe cancelable AsyncWrapper by explicitly specifying the template arguments in the constructor, like so:
```
#include "cpr/cpr.h"

int main(int, char**) {
    auto wrapper = cpr::AsyncWrapper<void, true>{std::future<void>{}};
    
    // Dereferences a null pointer
    auto _ = wrapper.IsCancelled();
}
```
To prevent that, we split AsyncWrapper it into two template specializations. AsyncWrapper<T, false> keeps most of the code, while cancelation-related code is made exclusive to AsyncWrapper<T, true>. Most importantly, we can prevent a cancelable AsyncWrapper from being constructed without a provided pointer. AsyncWrappers that aren't cancelable get to be a bit slimmer, no longer containing the pointer to keep track of cancelation state.

### DIFF
--- a/include/cpr/async_wrapper.h
+++ b/include/cpr/async_wrapper.h
@@ -91,7 +91,6 @@ class AsyncWrapper<T, true> : public AsyncWrapper<T, false> {
 
   public:
     // Constructors
-    AsyncWrapper() = default;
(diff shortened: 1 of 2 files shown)


COM8 (approved):
Thanks!

PR #1214, opened 2025-04-21
Implemented cpr::BodyView.

### LINKED ISSUE #1207: Non data owning cpr::Body for payloads

cURL is able to process POST requests with external body source (CURLOPT_POSTFIELDS + CURLOPT_POSTFIELDSIZE_LARGE). We are using this to speed up publishing of large binary data chunks (once constructed, many times sent, approximately 5mb per chunk). Unfortunately, `cpr::Body`/`cpr::Payload` does not provide view functionality over external storage (like `std::span` do).

### Possible Solution

I want to ask you to implement `cpr::BodyView` constructed from pointer + size (`std::span<const char>` will do) and inject it in `Content` variant (among with `cpr::Body`, `cpr::Payload`, third one and `std::monostate`) so target solution will be something like that:
```cpp
std::string post_json;
session.SetBody( cpr::BodyView{ post_json } );
session.Post();
```
```cpp
std::vector<char> binary_body;
session.SetBody( cpr::BodyView{ binary_body } );
session.Post();
```
Of course, users must guarantee data existence until request is done.

### Alternatives

For 1.10.5 our solution was:
```cpp
auto curl = s.GetCurlHolder();
s.SetBody( "" ); // didn't figured out why, legacy anyway.
curl_easy_setopt( curl->handle, CURLOPT_POSTFIELDSIZE_LARGE, static_cast<curl_off_t>( body_str.size() ) );
curl_easy_setopt( curl->handle, CURLOPT_POSTFIELDS, body_str.c_str() );
```
Pretty easy, straitforward, worked out well, we were happy bypassing internal cURL storage for our prepared data chunks.

Version 1.11.2 introduced different internal logic of handling bodies so we brought new solution with interceptor
```cpp
class MyInterceptor : public cpr::Interceptor {
public:
    explicit MyInterceptor( std::span<const char> new_body ) : m_new_body( new_body ) {}
    /*Skipped rule-of-5 ctors, operators, dtor*/

    [[nodiscard]] cpr::Response intercept( cpr::Session &session ) override
    {
        const std::shared_ptr<cpr::CurlHolder> curl = session.GetCurlHolder();
        if( !curl ) throw std::runtime_error{}; // Be gentle with any pointer, just in case.

        // Removing existing body will not reset CURLOPT_POSTFIELDS during process(...) call (again).
        // Session (or cURL handle) still initialized like it contains meaningful body. 
        // Feels like a really dirty hack.
        session.RemoveContent();

        // Finally we're get to point where no one can hold us back. Setting new body source just for cURL handle.
        curl_easy_setopt( curl->handle, CURLOPT_POSTFIELDSIZE_LARGE, static_cast<curl_off_t>( m_new_body.size() ) );
        curl_easy_setopt( curl->handle, CURLOPT_POSTFIELDS, m_new_body.data() );

        // At this point session 
        //    (a) initialized as POST request, 
        //    (b) has empty body so no changes to CURLOPT_POSTFIELDS will be made 
        //         (checked with debugger and we're not using read callbacks, anyway),
        //    (c) cURL will use our body source for request.
        return proceed( session );
    }

private:
    std::span<const char> m_new_body;
};

// Before session.Post() call.
void use_external_body( cpr::Session &session, std::span<const char> body_view )
{
    // Properly initialize session's body (and cURL handle with body parameters). 
    // Real body will be applied a bit later using interceptor.
    session.SetBody( {} );
    session.AddInterceptor( std::make_shared<MyInterceptor>( body_view ) );
}
```

T-Maxxx:
fixes #1207

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ compile_commands.json
 
 # macOS
 .DS_Store
+
+# CLion
+.idea/
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -289,6 +289,10 @@ void Session::RemoveContent() {
             curl_mime_free(curl_->multipart);
             curl_->multipart = nullptr;
         }
+    } else if (std::holds_alternative<cpr::BodyView>(content_)) {
+        // set default values, so curl does not send a body in subsequent requests
+        curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDSIZE_LARGE, -1);
+        curl_easy_setopt(curl_->handle, CURLOPT_POSTFIELDS, nullptr);
     }
     content_ = std::monostate{};
 }
@@ -486,6 +490,11 @@ void Session::SetBody(const Body& body) {
 
 void Session::SetBody(Body&& body) {
     content_ = std::move(body);
+}
+
+void Session::SetBodyView(BodyView body) {
+    static_assert(std::is_trivially_copyable_v<BodyView>, "BodyView expected to be trivially copyable otherwise will need some std::move across codebase");
+    content_ = body;
 }
 
 void Session::SetLowSpeed(const LowSpeed& low_speed) {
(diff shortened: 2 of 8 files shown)


COM8:
@T-Maxxx thanks for contributing!
I will try to review it as soon as possible but this week is quite busy for me.

Could you please update/extend the docs for this: https://github.com/libcpr/docs

COM8 (changes requested):
LGTM! Now just a bit of clang-tidy, cpcheck fixes and docs needed.

COM8:
Docs: https://github.com/libcpr/docs/pull/56

Issue #1174, opened 2025-02-08
`cpr::LowSpeed` Is Not Y2038 Safe


COM8:
### Description

`cpr::LowSpeed` has a 32 bit time value (`std::int32_t time;`). This is not year 2038 safe. Instead we should replace it with `std::chrono` for handling time.
Even though it's just the seconds and probably will never happen we should use `std::chrono`.

### Example/How to Reproduce

Not applicable.

### Possible Fix

Refactor `cpr::LowSpeed` to use `std::chrono`.

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS: Linux (Fedora 41)
- Version: 1.11.0

PR #1215, opened 2025-04-28
Changed LowSpeed to use std::chrono

pntzio:
Use std::chrono::seconds instead of std::int32_t for time parameter in cpr::LowSpeed


pntzio:
Fixes #1174

COM8 (approved):
Thanks!

Issue #1222, opened 2025-05-16
Add `CURLINFO_PRIMARY_IP` and `CURLINFO_PRIMARY_PORT` to `cpr::Response`


gitpaladin:
### Possible Solution

# Feature Request: Add `CURLINFO_PRIMARY_IP` and `CURLINFO_PRIMARY_PORT` to `cpr::Response`

## What

As a developer using CPR for HTTP requests, I need to access the resolved IP address and port of the target endpoint after making a request. This capability is crucial for monitoring, logging, and reporting purposes. Currently, CPR does not expose these details in the `cpr::Response` object.

## Why

Having access to the resolved IP and port is **highly valuable** for:
- Accurate monitoring of network traffic
- Detailed logging for debugging and auditing
- Proper reporting in distributed systems
- Security compliance and tracking

This feature is commonly required in production environments and would significantly enhance CPR's utility for enterprise and large-scale applications.

## Possible Solution

Extend the `cpr::Response` object to include the resolved IP and port by retrieving them during response creation:

```cpp
std::string ip;
long port;
curl_easy_getinfo(curl, CURLINFO_PRIMARY_IP, &ip);
curl_easy_getinfo(curl, CURLINFO_PRIMARY_PORT, &port);
```

These values should be stored in the `cpr::Response` object and made accessible via new member functions like `GetPrimaryIP()` and `GetPrimaryPort()`.

PR #1223, opened 2025-05-16
Add `primary_ip` `primary_port` to `Response`

gitpaladin:
see https://github.com/libcpr/cpr/issues/1222


COM8 (approved):
@gitpaladin thanks for contributing. LGTM!

PR #1232, opened 2025-07-03
Fix util namespaces to use the project with C++17 applications

sergeymarkow:
It is ok to use library which is built with newer C++ standart but it will be great to have including headers to keep C++17 standart. 

Unfortunatly I need to use Qt5.15 which still use C++17. Moreover I cannot rebuild Qt5.15 with other C++ params I can use already built project ((. I think in some other cases such change can be required.

### DIFF
--- a/include/cpr/util.h
+++ b/include/cpr/util.h
@@ -11,7 +11,8 @@
 #include "cpr/cprtypes.h"
 #include "cpr/secure_string.h"
 
-namespace cpr::util {
+namespace cpr{
+namespace util {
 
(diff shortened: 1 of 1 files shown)


COM8:
@sergeymarkow thanks for contributing!
But I always thought this was c++17 and also the standard says so: https://en.cppreference.com/w/cpp/language/namespace.html

Is this a special case where QT5 does not support this?

COM8:
Closing for no response. Feel free to reopen in case this is still a problem.

PR #1233, opened 2025-07-06
fix: Calling empty callbacks

nlogozzo:
When setting empty callbacks via `session.setProgressCallback({})` or other methods, a `bad_function_call` will be thrown when calling the callback.

This PR will check if callbacks are empty before calling them to prevent this.


COM8 (approved):
@nlogozzo thanks!

PR #1246, opened 2025-08-16
curl_container: allow calling GetContent without CurlHolder

muxelplexer:
This can be relevant when the parameters need to be passed to some different library (i.e. libmpv, gstreamer) and there is no cpr::Session to get a CurlHolder from.

---

I ran the clang-format script - it brought up issues but not with the file i touched. If there is a better way already implemented please let me know :)


include/cpr/curl_container.h
```diff
@@ -42,6 +42,12 @@ class CurlContainer {
 
     const std::string GetContent(const CurlHolder&) const;
```
COM8, line 43: Could you please add a docs comment to the old function as well?

include/cpr/curl_container.h
```diff
@@ -42,6 +42,12 @@ class CurlContainer {
...
 
+    /**
+     * Returns the URL while ignoring `encode`. This allows calling without
+     * active `CurlHolder`.
+     **/
+    const std::string GetContent() const;
```
COM8, line 49: Please add tests inside `test/structures_tests.cpp`.

COM8 (changes requested):
Thanks for submitting this MR.
Are you aware that something like this would also wok:
```c++
parameters.GetContent(CurlHolder());
```

(constructing a `CurlHolder()` is compatibly expensive)

COM8:
For running clang-format there is: `scripts/run_clang_format.sh`.

muxelplexer:
Running the `clang-format` script i get following changed files:
```
On branch muxelplexer/getcontent
Your branch is up to date with 'origin-mux/muxelplexer/getcontent'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   cpr/connection_pool.cpp
        modified:   test/abstractServer.hpp
        modified:   test/connection_pool_tests.cpp
        modified:   test/proxy_tests.cpp
```

Can add the formatting of those into this MR but none of them _should_ belong to my changes.

PR #1248, opened 2025-08-25
fix: make is_same_v check constexpr in set_option_internal

bi0ha2ard:
Adds a missing `constexpr` to check the result of `std::is_same_v<CurrentType, Header>`, since its value is known at compiletime.


COM8 (approved):
Thanks!

Issue #1186, opened 2025-02-23
`cpr::Multiperform` Assignment Not Working


COM8:
### Description

`cpr::Multiperform` assignment is broken once the object has been used before.

Workaround:
Use a `std::uniqe_ptr` around `cpr::Multiperform`.

### Example/How to Reproduce

Take this test case:
```c++
TEST(MultiperformGetTests, MultiperformAssignAfterUseTest) {
    MultiPerform multiperform;
    {
        Url url{server->GetBaseUrl() + "/hello.html"};
        std::shared_ptr<Session> session = std::make_shared<Session>();
        session->SetUrl(url);
        multiperform.AddSession(session);
        std::vector<Response> responses = multiperform.Get();

        EXPECT_EQ(responses.size(), 1);
        std::string expected_text{"Hello world!"};
        EXPECT_EQ(expected_text, responses.at(0).text);
        EXPECT_EQ(url, responses.at(0).url);
        EXPECT_EQ(std::string{"text/html"}, responses.at(0).header["content-type"]);
        EXPECT_EQ(200, responses.at(0).status_code);
        EXPECT_EQ(ErrorCode::OK, responses.at(0).error.code);
    }

    {
        Url url{server->GetBaseUrl() + "/hello.html"};
        std::shared_ptr<Session> session = std::make_shared<Session>();
        session->SetUrl(url);
        multiperform = MultiPerform(); // This line does not work if the Muliperform object was used before
        multiperform.AddSession(session);
        std::vector<Response> responses = multiperform.Get();

        EXPECT_EQ(responses.size(), 1);
        std::string expected_text{"Hello world!"};
        EXPECT_EQ(expected_text, responses.at(0).text);
        EXPECT_EQ(url, responses.at(0).url);
        EXPECT_EQ(std::string{"text/html"}, responses.at(0).header["content-type"]);
        EXPECT_EQ(200, responses.at(0).status_code);
        EXPECT_EQ(ErrorCode::OK, responses.at(0).error.code);
    }
}

Running it leads to a segmentation fault in Line 361 of `multiperform.cpp`.

![Image](https://github.com/user-attachments/assets/4857c2e9-b889-44b4-849c-7915f3b71045)

```

### Possible Fix

Reset everything during assignment operator.

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS: Fedora 41
- Version: 1.11.2

Issue #1047, opened 2024-04-28
Multiple Invocations of cpr::MultiPerform Segfault


COM8:
### Description

Invoking any `cpr::MultiPerform` request e.g. `cpr::MultiPerform::Get();` multiple times on the same object leads to a crash (SegFault) during the second invocation.

This leads to `cpr::MultiPerform` to being reusable.

### Example/How to Reproduce

Here is an example test case:
```c++
TEST(MultiperformGetTests, MultiperformSingleSessionMultiGetTest) {
    Url url{server->GetBaseUrl() + "/hello.html"};
    std::shared_ptr<Session> session = std::make_shared<Session>();
    session->SetUrl(url);
    MultiPerform multiperform;
    multiperform.AddSession(session);
    std::vector<Response> responses = multiperform.Get();

    EXPECT_EQ(responses.size(), 1);
    std::string expected_text{"Hello world!"};
    EXPECT_EQ(expected_text, responses.at(0).text);
    EXPECT_EQ(url, responses.at(0).url);
    EXPECT_EQ(std::string{"text/html"}, responses.at(0).header["content-type"]);
    EXPECT_EQ(200, responses.at(0).status_code);
    EXPECT_EQ(ErrorCode::OK, responses.at(0).error.code);

    // Invoke for the second time
    responses = multiperform.Get(); // Fails here
    EXPECT_EQ(responses.size(), 1);
    EXPECT_EQ(expected_text, responses.at(0).text);
    EXPECT_EQ(url, responses.at(0).url);
    EXPECT_EQ(std::string{"text/html"}, responses.at(0).header["content-type"]);
    EXPECT_EQ(200, responses.at(0).status_code);
    EXPECT_EQ(ErrorCode::OK, responses.at(0).error.code);
}
```

### Where did you get it from?

GitHub (branch e.g. master)

### Additional Context/Your Environment

- OS: Fedora 39
- Version: Master as of 28.04.2024 - bed704ff000a28292aebcdf3868d937f853ef082

PR #1249, opened 2025-09-01
cpr::MultiPerform fixes - #1047 and #1186

Anohkka:
Fixes for #1047 and #1186. I took the liberty of changing cpr::Multiperform code to use structured bindings for better readability as well, but I can just revert that commit if so desired.

- #1047 was happening because easy handles added to a multi stack can't normally be reused, but we can work around that by removing and adding handles to the multi stack.

- #1186 was caused by the default move constructor and move assignment operator copying over the invalid `current_interceptor_` and `first_interceptor_` iterators from the moved-from MultiPerform object.


COM8 (approved):
Wonderful! Thank you very much for fixing this.

PR #1254, opened 2025-09-20
Add initial devcontainer configuration

s20021072:


COM8:
@s20021072 thanks for contributing. Will there be more than this?

PR #1262, opened 2025-10-25
Optimize Base64Decode function

Tina-1300:
This PR replaces the existing Base64Decode implementation with a more optimized version : 

- Pre-allocates the output string to avoid dynamic reallocations.
- Decodes 4 Base64 characters into 3 bytes per iteration.
- Branchless main loop for better CPU performance.
- Handles padding '=' efficiently after the main loop.
- Uses a static lookup table initialized once, reducing overhead on repeated calls.

Performance improvements are noticeable for large Base64 inputs.

### DIFF
--- a/test/abstractServer.cpp
+++ b/test/abstractServer.cpp
@@ -94,34 +94,48 @@ void AbstractServer::ResetConnectionCount() {
     unique_connections.clear();
 }
 
-static const std::string base64_chars =
-        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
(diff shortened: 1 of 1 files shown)


COM8:
> It's great to see you again. I thought the project was dead, so I'm glad @COM8 replied. How do you come up with such good library ideas? I've been racking my brain, but I can't find a problem to solve, and I'd love to be the creator of a well-known library like you, or (the creator of Laravel). I admire you a lot.

Thanks 😅. I have to be open with you. I did not invent this lib. I just took it over a few years ago when it basically was not maintained any more and needed some large changes.

PR #1270, opened 2025-11-10
Added support for CURLOPT_SSLCERT_BLOB

ssauermann:
This PR adds support for setting the client certificate as a blob using CURLOPT_SSLCERT_BLOB instead of passing a file path.

Implemented like the existing CertFile option: CertBlob with an alias PemBlob and a derived DerBlob class. A matching test case is also there.


COM8 (approved):
Thanks!

Issue #1014, opened 2024-01-14
Handling SSE


vymao:
### Is your feature request related to a problem?

Would be great if CPR could handle server sent events for streamed data. Currently I don't see mention of this in the documentation, and I'm trying to work around it by making WriteCallbacks, but this isn't optimal.

### Possible Solution

A callback that instead could take in [server sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) with error handling, instead of us having to parse it out manually and do a lot of string formatting with WriteCallbacks.

COM8:
As I understand it, a better "unbuffered" mode that invokes a callback for every event received is desired here?

vymao:
Essentially yes, since right now, WriteCallbacks receive data in the form of a string, which isn't ideal to format if we are expecting JSON payloads.

COM8:
Right now this feature is not planed, but you are more than welcome to create PR for it.

PR #1274, opened 2025-11-19
Add support for Server Sent Events (SSE)

bachp:
This extension allows to handle SSE events
by invoking a callback whenever an event is received.

Related documentation: https://github.com/libcpr/docs/pull/59

Closes https://github.com/libcpr/cpr/issues/1014

I tested this against an [axum](https://docs.rs/axum/latest/axum/) server with the following example:

```cpp
#include <cpr/cpr.h>
#include <iostream>

int main() {
    std::cout << "Connecting to SSE endpoint at http://localhost:9976/hello..." << std::endl;

    cpr::Session session;
    session.SetUrl(cpr::Url{"http://localhost:9976/hello"});

    // Set up SSE callback to print events to console
    session.SetServerSentEventCallback(
        cpr::ServerSentEventCallback{
            [](cpr::ServerSentEvent&& event, intptr_t userdata) {
                std::cout << "=== New Event ===" << std::endl;
                std::cout << "Event Type: " << event.event << std::endl;
                std::cout << "Data: " << event.data << std::endl;

                if (event.id.has_value()) {
                    std::cout << "ID: " << event.id.value() << std::endl;
                }

                if (event.retry.has_value()) {
                    std::cout << "Retry: " << event.retry.value() << " ms" << std::endl;
                }

                std::cout << std::endl;

                // Return true to continue receiving events
                return true;
            }
        }
    );

    std::cout << "Connection closed." << std::endl;
    std::cout << "Status Code: " << response.status_code << std::endl;

    return 0;
}
```


cpr/sse.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    if (colon_pos == std::string::npos) {
+        // No colon, entire line is the field name
+        field = line;
+        value = "";
+    } else {
+        field = line.substr(0, colon_pos);
+        // Skip the colon and optional leading space
+        size_t value_start = colon_pos + 1;
+        if (value_start < line.size() && line[value_start] == ' ') {
```
COM8: Does it make sense to trim the string left and right in case there are multiple white space characters?

COM8 (changes requested):
@bachp Thank you very much for contributing  this!
I only have a minor comment and clang-tidy is not yet completely happy. The `abseil-string-find-str-contains` clang-tidy error can be added to the ignore list inside `.clang-tidy`.
The other CI runs are currently expected to fail.

COM8:
@bachp I rebased and then created a new release: https://github.com/libcpr/cpr/releases/tag/1.14.0

PR #1285, opened 2026-01-08
default to NONE for PostRedirectFlags

transmissions11:
this unfortunately may be breaking for some users, so if strong backwards compatibility is important maybe wait to include this in a major release or disregard it entirely, but:

both `libcurl` and python's `requests` library do _not_ preserve POST on redirects for 301/302/303 statuses by default (see https://curl.se/libcurl/c/CURLOPT_POSTREDIR.html). this is crucial for various common OAuth/SAML flows which expect GET requests after redirects. 

given this behavior is the default in other widely-used environments (notably browsers as well), and particularly because cpr aims to be a "spiritual port" of `requests`, i believe `NONE` is a much more sensible default than `POST_ALL`.


transmissions11:
ah sorry looks like i'd need to update some tests, if you think you'd be willing to merge i can go ahead and fix them

PR #1291, opened 2026-01-27
Updated Bazel Instructions

J-B-Blankenship:
**CMake Approach Issues**

_First Issue_
Non-ASCII characters cannot be parsed by glob. 

```
ERROR: /root/.cache/bazel/_bazel_root/646bcbb063181c5dac5a581e1c4fefdd/external/_main~_repo_rules~cpr/BUILD.bazel:22:6: Foreign Cc - CMake: Building cpr failed: error reading file '@@_main~_repo_rules~cpr//:test/data/test_file_hello_äüöp_2585_你好.txt': /root/.cache/bazel/_bazel_root/646bcbb063181c5dac5a581e1c4fefdd/external/_main~_repo_rules~cpr/test/data/test_file_hello_äüöp_2585_你好.txt (No such file or directory)
```

_Second Issue_
Resolving this error still causes an issue with `<filesystem>`. I was unable to override C++17 for C++20 to make this work with the `cmake` toolchain.

**Repo Reference Issue**
In addition, the referenced Bazel repo is deprecated by several years. I was unable to get them to work with BAZEL modules.

**Solution**
Use Bazel to build the project with a patch to remove the generated version file `cpr/cprver.h` (only created with the CMake command). 

This was confirmed to work on Ubuntu 24.04. Modifications should be straightforward for Bazel users on Windows or Mac as it uses common toolchains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,8 +147,6 @@ ctest -VV # -VV is optional since it enables verbose output
 ```
 
 ### Bazel
-Please refer to [hedronvision/bazel-make-cc-https-easy](https://github.com/hedronvision/bazel-make-cc-https-easy) or
-
 `cpr` can be added as an extension by adding the following lines to your bazel MODULE file (tested with Bazel 8). Edit the versions as needed.
 ```starlark
 bazel_dep(name = "curl", version = "8.8.0.bcr.3")
--- a/README.md
+++ b/README.md
@@ -152,48 +152,43 @@ Please refer to [hedronvision/bazel-make-cc-https-easy](https://github.com/hedro
 `cpr` can be added as an extension by adding the following lines to your bazel MODULE file (tested with Bazel 8). Edit the versions as needed.
 ```starlark
 bazel_dep(name = "curl", version = "8.8.0.bcr.3")
-bazel_dep(name = "platforms", version = "0.0.11")
-bazel_dep(name = "zlib", version = "1.3.1.bcr.5")
-bazel_dep(name = "boringssl", version = "0.20250212.0")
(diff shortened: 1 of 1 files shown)


J-B-Blankenship:
> @J-B-Blankenship thanks for contributing! Could you please provide a minimal example for this? I've never really used bazel and this would make testing a bit easier.

I have this repo where it is being used: https://github.com/RogueDB/roguedb-open-source. It should be stable for the `master` branch.

If you want a quick test, you can do the following:
 - `docker compose run develop` to build and run the image with all the dependencies pre-installed

Inside the container:
 - `bazel build //...` -> There is C++ code in `getting_started/` that is using CPR. If it builds (as it should), then you are good to go.

The change primarily does the following:

 - Uses the compiler used by Bazel (clang in my case)
 - Uses the flags in `.bazelrc` or passed via command line (eg. using C++20 with -O3)
 - Bypasses CMake and excludes the generated version file with a patch (eg. in-place application of a `git diff`).

Please let me know if you have additional questions. If I can make the instructions clearer, I can also do that.

COM8 (approved):

spinningthelock:
@J-B-Blankenship Does the https://registry.bazel.build/modules/cpr module not work for you?

J-B-Blankenship:
> @J-B-Blankenship Does the https://registry.bazel.build/modules/cpr module not work for you?

I honestly just looked at the README.md. I did not try the Bazel registry (didn't know it existed).

spinningthelock:
No problem. I can put up a PR to update the instructions.

PR #1298, opened 2026-02-23
⚡️ Share SSL session data in ConnectionPool for TLS session resumption

guylevy84:
Without this, each new easy handle using the shared connection pool must perform a full TLS handshake even when reusing a pooled TCP connection. Sharing `CURL_LOCK_DATA_SSL_SESSION` allows libcurl to resume TLS sessions across handles.


COM8 (approved):
@guylevy84 thanks for contributing!

PR #1304, opened 2026-04-25
Update Readme instructions to use BCR

spinningthelock:
Previous instructions on how to use cpr with Bazel was outdated. The module is available via Bazel Central Registry, and can be added to the the project with a single line addition.


COM8 (approved):
Thanks @spinningthelock!

PR #1305, opened 2026-04-28
Add CPR_SKIP_INSTALL option

qr243vbi:
Add CPR_SKIP_INSTALL to use as bundled dependency


CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ endmacro()
 
 message(STATUS "C++ Requests CMake Options")
 message(STATUS "=======================================================")
+cpr_option(CPR_SKIP_INSTALL "Set to ON to use as bundled dependency" OFF)
```
COM8: Negative options are always harder to get your head around. How about:

```suggestion
cpr_option(CPR_ENABLE_INSTALL "Set to ON to enable all cpr install targets." ON)
```

COM8 (changes requested):
@qr243vbi thanks for contributing!
Only one comment.

CMakeLists.txt
qr243vbi: Good

COM8 (approved):
Thanks!

PR #1128, opened 2024-10-18
Resolve CURLOPT_SSL_OPTIONS issues

gentooise:
Closes #1127 

Note for reviewers: I tried to keep equivalent `#ifdef` conditions, but I'm not really sure on the history/rationale of those. Please double check.

PR #1316, opened 2026-08-02
restore native ca ssl option by default

### LINKED ISSUE #1313: Regeession: CURLSSLOPT_NATIVE_CA is no longer automatically set
### Description

!1128 introduced a regression where CURLSSLOPT_NATIVE_CA is no longer set on a
`Session`'s construction.

This change was motivated by #1127, which discovered that `curl_easy_setopt`
was being used incorrectly. It was being used with the expectation that it was
additive, i.e. each call added an option, but the issue points out that
`curl_easy_setopt` is supposed to be called once, and that it cancels
previously set options. Instead, options should be OR'ed together before
calling `curl_easy_setopt`.

!1128 fixes this, but it moves some options from `prepareCommonShared` (which
gets called before each request) to `SetSslOptions`, which needs to be
explicitly called. One of these options is CURLSSLOPT_NATIVE_CA.

### Example/How to Reproduce

Create any request that requires verifying ssl via the Windows cert store.

### Possible Fix

_No response_

conan

- OS: Windows
- cpr version: >=1.11.1
- with_ssl=openssl

aouxwoux:
## what changed

restore CURLSSLOPT_NATIVE_CA when a Session is constructed on curl 7.71.0 and newer.

## why

#1128 correctly changed CURLOPT_SSL_OPTIONS to use one combined bitmask, but it moved the native-ca setting into SetSslOptions. that method is only called when ssl options are explicitly configured, so normal sessions no longer enable the native ca behavior on windows with openssl.

this keeps the bitmask fix in place while restoring the default constructor behavior. it also leaves the request preparation path free of repeated ssl-option writes.

## checks

- git diff --check passes.
- the local checkout does not have cmake or a c++ compiler installed, so i could not run the project test suite locally.

fixes #1313

### DIFF
--- a/cpr/session.cpp
+++ b/cpr/session.cpp
@@ -154,6 +154,11 @@ Session::Session() : curl_(new CurlHolder()) {
     curl_easy_setopt(curl_->handle, CURLOPT_NOSIGNAL, 1L);
 #endif
 
+#if LIBCURL_VERSION_NUM >= 0x074700 // 7.71.0
+    // Fix loading certs from Windows cert store when using OpenSSL:
+    curl_easy_setopt(curl_->handle, CURLOPT_SSL_OPTIONS, CURLSSLOPT_NATIVE_CA);
+#endif
+
 #if LIBCURL_VERSION_NUM >= 0x071900 // 7.25.0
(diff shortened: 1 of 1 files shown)


COM8 (approved):
Looks good. Thanks @aouxwoux.

Once you are happy with it, feel free to mark the PR as ready for review.

PR #1318, opened 2026-08-28
Fix malformed sha tag

### LINKED ISSUE #1317: Unneeded characters in CMake chapter
### Description

When i started using cpr, i started my own project using CMake, after that i found mistake in this code:
include(FetchContent)
                         GIT_TAG f091b2c061b307ee89b164c39976fc9202a1c79d.12.0) # Replace with your desired git commit from: https://github.com/libcpr/cpr/releases
FetchContent_MakeAvailable(cpr)
after the git tag there is .12.0 for some reason, and while it is there, code won't work.

### Example/How to Reproduce

-

### Possible Fix

Just delete it

GitHub (branch e.g. master)

- OS: Windows
- Version: 11

mbKaleb:
Fixes #1317 

## Changes:
- README.md: drops the `.12.0` so the snippet pins the bare commit SHA.
- readme-updater.yml: change the pattern to `(GIT_TAG) [^ )]+` so it
  replaces the whole token whatever its shape (SHA or dotted version) and
  stops at the closing paren. This should prevent partial-match leftovers

Verified the new expression against both historical shapes — `GIT_TAG 1.12.0)`
and `GIT_TAG f091…79d.12.0)` each come out as `GIT_TAG <new-sha>)`.

### DIFF
--- a/.github/workflows/readme-updater.yml
+++ b/.github/workflows/readme-updater.yml
@@ -19,7 +19,7 @@ jobs:
       with:
         ref: ${{github.sha}}
     - name: "Replace GIT_TAG"
-      run: sed -i -re 's/(GIT_TAG) [0-9a-f]+/\1 ${{github.sha}}/' README.md
+      run: sed -i -re 's/(GIT_TAG) [^ )]+/\1 ${{github.sha}}/' README.md
     - name: "Commit changes"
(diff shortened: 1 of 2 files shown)


COM8 (approved):
Thanks!

PR #1321, opened 2026-09-17
CMake: Use `OpenSSL_FOUND` over `OPENSSL_FOUND`

crueter:
The old variable is deprecated in `FindOpenSSL`, and find_package itself will set this.


COM8 (approved):
Thanks @crueter !

